PR #57, opened 2023-10-17
Refactor API

sam-finch-tezos:
# Context
Draft PR refactoring the API's 

Currently only the `Contract` API has been migrated, this means that most API's use the old approach and there are two transactions. This means some things are pretty messy 🤢.
Currently this introduces a bug that `Contract.create` does not commit the new contract until the end of the call. 

This PR proposes the following
* Global apis should not have internal state. A single stateful object `Jstz` contains the key value store, the transaction and relevant addresses.
* A slightly janky macro `with_jstz` to call functions providing the `Jstz` and `Host` objects.
* API's are hierarchical, with a single `Api` object for each module. Initialising an api initialises all sub apis. 
* a single function is provided to initialise all functions for a realm
* `Api` Trait is renames `GlobalApi` because all API's are now called `Api`


johnyob:
Closing as discussed in weekly sprint review. 
Possible future PRs from this work:
- Use our fork of Boa to provide `HostDefined`
- Use for a single struct for all `jstz` native objects (`tx`, `self_address`, `origin`, `referrer`)

PR #59, opened 2023-10-18
Refactor dyn Value

ryutamago:
# Context

`dyn Value` currently implements all the methods from `dyn Any`. We can avoid having to implement these methods and simplify the api by adding an `as_any` method that casts from `Value` to  `dyn Any`. 

# Description

add the `as_any` and `as_any_mut` methods to the `Value` trait

# Manually testing the PR

ran the hello world example and confirmed that it works


johnyob:
LGTM

johnyob (approved):

PR #60, opened 2023-10-18
Cli Repl: Automatically resolve promises and add last result variable

sam-finch-tezos:
# Context
The CLI repl does not support the `await` keyword at the lop level.
This PR turns the bug into a feature by adding the following behaviour
* If the repl command returns a promise the. repl will block until the promise is resolved and returns the fulfilment value of the promise
* The result of the last run command, or the fulfilment value in the case of a promise, is stored in a variable `_` for easy reuse.

**Related Tasks**: [REPL: Resolve promises automatically](https://app.asana.com/0/1205170694555856/1205666947698889/f)

# Manually testing the PR
```javascript
cargo run -- repl
> Contract.create("export default () => { console.log(`Hello from ${Ledger.selfAddress} 👋`); return new Response() }");
> new Request(`tezos://${_}`);
> Contract.call(_);
```

### DIFF
--- a/jstz_cli/src/repl.rs
+++ b/jstz_cli/src/repl.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use boa_engine::{JsResult, JsValue, Source};
+use boa_engine::{js_string, JsResult, JsValue, Source};
 use jstz_api::{http::HttpApi, url::UrlApi, ConsoleApi, KvApi, TextEncoderApi};
 use jstz_core::host::HostRuntime;
 use jstz_core::{
@@ -104,10 +104,10 @@ pub fn exec(self_address: Option<String>) -> Result<()> {
 fn evaluate(input: &str, rt: &mut Runtime, hrt: &mut (impl HostRuntime + 'static)) {
     let rt_output = runtime::with_host_runtime(hrt, || -> JsResult<JsValue> {
         let value = rt.eval(Source::from_bytes(input))?;
-
-        jstz_core::future::block_on(rt.run_event_loop());
-
-        Ok(value)
+        jstz_core::future::block_on(async {
+            rt.run_event_loop().await;
+            rt.resolve_value(&value).await
+        })
     });
 
     match rt_output {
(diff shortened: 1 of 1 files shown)


alanmarkoTrilitech:
I tried running it, but the example doesn't seem to work properly:

`Using mock self-address tz4RepLRepLRepLRepLRepLRepLRepN7Cu8j.
>> Contract.create("export default () => { console.log(`Hello from ${Ledger.selfAddress} 👋`); return new Response()) }");
[📜] Contract created: tz4RVBAHakYgHx7XYK1nacqCGikzWBGGm6Ld
tz4RVBAHakYgHx7XYK1nacqCGikzWBGGm6Ld
>> new Request(`tezos://${_}`);
[object Object]
>> Contract.call(_);
Evaluating: "export default () => { console.log(`Hello from ${Ledger.selfAddress} 👋`); return new Response()) }"
Uncaught EvalError: JsError
>> `

sam-finch-tezos:
> I tried running it, but the example doesn't seem to work properly:

There was an extra bracket in the example contract. It should work now

alanmarkoTrilitech (approved):
LGTM!

jstz_cli/src/repl.rs
```diff
@@ -110,19 +110,32 @@ fn evaluate(input: &str, rt: &mut Runtime, hrt: &mut (impl HostRuntime + 'static
...
-                        .to_std_string_escaped()
-                );
+    jstz_core::future::block_on(async move {
+        match rt_output {
+            Ok(mut res) => {
+                if res.is_promise() {
+                    if let Ok(promise_result) = rt.resolve_value(&res).await {
+                        res = promise_result;
+                    }
+                }
```
johnyob, line 120: `resolve_value` will handle the `is_promise` logic. Can you move this to where we define `rt_output`

johnyob:
LGTM, please clean up your git history before merging (merge commits aren't allowed in `main`)

johnyob:
Changes look good. I'm not entirely happy that we automatically resolve promises (since I don't expect this from most REPLs). But for now it is a good workaround the issues we've had with promises in the REPL.

johnyob (approved):

PR #61, opened 2023-10-18
Cli: Allow data from pipes and files

sam-finch-tezos:
# Context
This pr makes it easier to deploy contracts and attach request bodies from files.

# Description
This PR changes the interface for a single argument from each of two commands
* `jstz deploy --contract-code` 
  - If the supplied argument is a file the code will be read from that file
  - if the supplied argument is not a file the code will be the argument itself
  - if the argument is not supplied the code will be read from a pipe
  - if there is no piped data an error is raised
  - This is now a positional argument
* `jstz run --data` 
  - If the supplied argument is a file the request body will be read from that file
  - if the supplied argument is not a file the request body will be the argument itself
  - if the argument is not supplied the body will be read from a pipe
  - if there is no piped data no body will be attached.

# Manually testing the PR
* direct input
    ```bash
    cargo run -- deploy --self-address $tz4 --function-code "`cat examples/print_body.js`"
    # copy contract address to clipboard
    cargo run -- run "tezos://`pbpaste`" $tz4 --data "`cat jstz_cli/sandbox.json`"
    ```
* file input
    ```bash
    cargo run -- deploy --self-address $tz4 --function-code examples/print_body.js
    # copy contract address to clipboard
    cargo run -- run "tezos://`pbpaste`" $tz4 --data jstz_cli/sandbox.json
    ```
* pipe input
    ```bash
    cat examples/print_body.js | cargo run -- deploy --self-address $tz4 
    # copy contract address to clipboard
    cat jstz_cli/sandbox.json | cargo run -- run "tezos://`pbpaste`"
    ```

### DIFF
--- a/jstz_cli/src/deploy.rs
+++ b/jstz_cli/src/deploy.rs
@@ -1,14 +1,23 @@
 use anyhow::Result;
 use serde_json::json;
 
-use crate::{config::Config, octez::OctezClient};
+use crate::{
+    config::Config,
+    octez::OctezClient,
+    utils::{from_file_or_id, piped_input},
+};
 
 pub fn exec(
     self_address: String,
-    contract_code: String,
+    contract_code: Option<String>,
     balance: u64,
     cfg: &Config,
 ) -> Result<()> {
+    // resolve contract code
+    let contract_code = contract_code
+        .map(from_file_or_id)
+        .or_else(piped_input)
+        .unwrap_or(String::default());
     // Create JSON message
     let jmsg = json!({
         "DeployContract": {
--- a/jstz_cli/src/deploy.rs
+++ b/jstz_cli/src/deploy.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use serde_json::json;
 
 use crate::{
@@ -17,7 +17,7 @@ pub fn exec(
     let contract_code = contract_code
         .map(from_file_or_id)
         .or_else(piped_input)
-        .unwrap_or(String::default());
+        .ok_or(anyhow!("No function code supplied"))?;
     // Create JSON message
     let jmsg = json!({
         "DeployContract": {
--- a/jstz_cli/src/main.rs
+++ b/jstz_cli/src/main.rs
@@ -26,12 +26,12 @@ enum Command {
         /// Address used when deploying the contract
         #[arg(short, long)]
         self_address: String,
-        /// Function code.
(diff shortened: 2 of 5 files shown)


alanmarkoTrilitech (approved):
What is the advantage of allowing deploying an empty smart function?

Otherwise, it seems reasonable.

jstz_cli/src/utils.rs
```diff
@@ -0,0 +1,21 @@
...
+    let lines: Result<Vec<_>, _> = io::stdin().lines().collect();
+    lines.expect("Can't read from stdin").join("\n")
+}
+pub fn piped_input() -> Option<String> {
+    let stdin = io::stdin();
+    if !stdin.is_terminal() {
+        Some(get_stdin())
+    } else {
+        None
+    }
```
johnyob, line 20: Naive question: What is the behaviour of `is_terminal` in shell scripts?

jstz_cli/src/main.rs
```diff
@@ -26,10 +27,10 @@ enum Command {
         #[arg(short, long)]
         self_address: String,
         /// Function code.
-        #[arg(short, long)]
-        function_code: String,
+        #[arg(short, long, default_value = None)]
+        function_code: Option<String>,
```
johnyob, line 31: It would be nice if this wasn't a flagged argument anymore, but the only unnamed positional argument

jstz_cli/src/deploy.rs
```diff
@@ -1,14 +1,23 @@
...
-    contract_code: String,
+    contract_code: Option<String>,
     balance: u64,
     cfg: &Config,
 ) -> Result<()> {
+    // resolve contract code
+    let contract_code = contract_code
+        .map(from_file_or_id)
+        .or_else(piped_input)
+        .unwrap_or(String::default());
```
johnyob, line 20: I think we should error here instead of deploying an empty smart function

jstz_cli/src/utils.rs
sam-finch-tezos: It's the same as in the terminal. It will return true if stdin is coming from the terminal (not from a pipe). The purpose of this line is to avoid the cli pausing and accepting input from the terminal. (Although we might want this behaviour)

johnyob:
LGTM. 

Before merging you might want to improve the commit name of 2d2da979da8eef2d4a9f13a69a53e8339594589d (looks like you used ` without escaping it properly)

PR #54, opened 2023-10-13
CLI: Deprecating SetTicketer

alanmarkoTrilitech:
# Context

https://app.asana.com/0/1205170694555918/1205581410505410
# Description
This PR deprecates SetTicketer external message. It creates an installer kernel using a setup-file with the KT1 address obtained after originating the rollup kernel.

# Manually testing the PR
cargo run --bin jstz -- sandbox start

PR #62, opened 2023-10-18
Bugfix: Incorrect `bridge_address` encoding

johnyob:
# Context

Following #54, the sandbox no-longer works due a panic in `jstz_kernel`.

# Description

This panic occurs due to a deserialization error. The root of this error was serializing the KT1 bridge address
as a tz4 address. 

# Manually testing the PR

```sh
cargo run -- sandbox start
```

### DIFF
--- a/jstz_cli/src/sandbox/daemon.rs
+++ b/jstz_cli/src/sandbox/daemon.rs
@@ -10,10 +10,10 @@ use std::{
 use anyhow::Result;
 use fs_extra::dir::CopyOptions;
 use jstz_core::kv::value::serialize;
-use jstz_crypto::public_key_hash::PublicKeyHash;
 use nix::libc::{SIGINT, SIGTERM};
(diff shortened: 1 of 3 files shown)


jstz_cli/src/sandbox/daemon.rs
```diff
@@ -409,7 +417,7 @@ pub fn main(cfg: &mut Config) -> Result<()> {
     cfg.save()?;
 
     // 4. Wait for the sandbox to shutdown (either by the user or by an error)
-    OctezThread::join(vec![rollup_node, baker, node])?;
+    OctezThread::join(vec![baker, rollup_node, node])?;
```
alanmarkoTrilitech, line 420: Why is this change needed?

alanmarkoTrilitech (approved):
Added a comment with a question regarding the  change in sandbox shut down. Otherwise it looks good to me!

PR #64, opened 2023-10-23
CLI Enhancements: Account creation + config file change

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205170694555856/1205670949626065/f

# Description
This PR enables account creation using `jstz account create command` command based on the passphrase provided. If none is provided a random one is generated. It saves the created account to a list in the config file. 

# Manually testing the PR
`cargo run -- account create bob`
`cargo run -- account create bob --passphrase greatpassphrase`


jstz_cli/src/accounts/account_list.rs
```diff
@@ -0,0 +1,39 @@
+use std::collections::HashMap;
```
johnyob, line 1: I don't really see a need for `AccountList`, please inline the `HashMap<String, Account>` type in `Config`

jstz_cli/src/accounts/account.rs
```diff
@@ -0,0 +1,32 @@
+use serde::{Deserialize, Serialize};
+
+// Represents an individual account
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Account {
+    alias: String,
+    address: String,
+}
```
johnyob, line 8: Missing PublicKey and SecretKey?

jstz_cli/src/accounts/account_management.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::{accounts::account::Account, config::Config};
+
+extern crate bs58;
+
+fn passphrase_to_ikm(passphrase: &str) -> [u8; 32] {
+    let hash = Sha256::digest(passphrase.as_bytes());
+    let mut ikm = [0u8; 32];
+    ikm.copy_from_slice(&hash);
+    ikm
+}
```
johnyob, line 16: Generate the input key material using our `Blake2b` 256-bit hasher from `jstz_crypto` instead of SHA-256

jstz_cli/src/accounts/account_management.rs
```diff
@@ -0,0 +1,48 @@
...
+pub fn generate_passphrase() -> String {
+    let mut rng = rand::thread_rng();
+    let mut passphrase = String::new();
+    for _ in 0..10 {
+        let random_number: u8 = rng.gen();
+        let random_char = (random_number % 26) + 97;
+        passphrase.push(random_char as char);
+    }
+    passphrase
+}
```
johnyob, line 31: I'd prefer to generate a BIP-39 mnemonic instead of some random string like this

jstz_cli/src/accounts/account_management.rs
```diff
@@ -0,0 +1,48 @@
...
+    let (sk, pk) = keypair_from_ikm(ikm).unwrap();
+    println!("Secret key: {}", sk);
+    println!("Public key: {}", pk);
+    let address = create_address(jstz_crypto::public_key::PublicKey::Bls(pk));
+
+    cfg.accounts().upsert(Account::new(alias, address.clone()));
+    cfg.save()?;
+
+    Ok(address)
+}
```
johnyob, line 48: Move to `account.rs`. This function could be `Account::from_passphrase(...)`

jstz_cli/src/accounts/account_management.rs
johnyob: Move to `jstz_crypto`

johnyob (changes requested):

jstz_cli/src/accounts/account_list.rs
alanmarkoTrilitech: The idea is that it should be an abstraction for account list hiding its actual implementation - maybe a better name would be AccountConfig. Right now it just wraps the hashmap, but later I wanted to extend it with modification of self-address, methods for printing all accounts, some methods of type "find me an account in the list that has this alias" etc.

johnyob: > but later I wanted to extend it with modification of self-address
You mean storing the 'current' account, which is set with `jstz login` and printed with `jstz whoami`?
> methods for printing all accounts
Likely not that best place to put that, observable behaviour (such as printing) should be attached with command code
> some methods of type "find me an account in the list that has this alias" etc.
Makes sense

Given this, I'd suggest moving `AccountList` to `config.rs` with the type `AccountConfig`

alanmarkoTrilitech: instead of printing I think there could be a method like to_string which would then be used by the command code and printed. I can move it to the config - it seems best to me to create a folder config with config.rs and accountconfing.rs and sandboxconfig.rs.

johnyob: I think given we have `SandboxConfig` and `Config` in `config`, we should probably add `AccountConfig` to here. 

Additionally while these structs have enough functionality to warrant a type, I don't think they'll have enough functionality to warrant a module.

jstz_cli/src/accounts/account.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    pub fn get_alias(&self) -> &String {
+        &self.alias
+    }
+
+    pub fn get_address(&self) -> &String {
+        &self.address
+    }
```
johnyob, line 40: Getters are an anti-pattern in Rust (unless there is non-trivial logic). Please remove and make `alias` and `address` public fields

jstz_cli/src/accounts/account.rs
```diff
@@ -0,0 +1,64 @@
...
+    PublicKeyHash::try_from(&pk).unwrap().to_base58()
+}
+
+// Represents an individual account
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Account {
+    alias: String,
+    address: String,
+    secret_key: String,
+    public_key: String,
```
johnyob, line 16: These could match their Rust type counterparts for both improved serialization + less conversions between strings and `PublicKey`/`SecretKey`

jstz_cli/src/accounts/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+            println!("Generated passphrase: {}", passphrase);
+            passphrase
+        }
+    };
+
+    let account = account::Account::from_passphrase(passphrase, alias)?;
+    cfg.accounts().upsert(account.clone());
+    cfg.save()?;
+
+    println!("Account created with address: {}", account.get_address());
```
johnyob, line 32: Unnecessary `.clone`. I recommend lifting the `println` then modifying the config

jstz_cli/src/config.rs
```diff
@@ -8,12 +9,48 @@ use std::{
...
+}
+
+impl AccountConfig {
+    /*
+    pub fn new() -> Self {
+        AccountList {
+            accounts: HashMap::new(),
+        }
+    }
+    */
```
johnyob, line 33: Please avoid committing possibly dead code in comments, these kinds of comments often bit rot quickly. 

If you have local changes that you'd like to perhaps use in the future just commit sections of the file and stash the changes you'd like to keep for later

jstz_cli/src/accounts/account.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        println!("Secret key: {}", sk);
+        println!("Public key: {}", pk);
+
+        let address = create_address(jstz_crypto::public_key::PublicKey::Bls(pk.clone()));
+        let new_account =
+            Account::new(alias, address.clone(), sk.to_string(), pk.to_string());
+
+        Ok(new_account)
+    }
```
johnyob, line 63: Some of this logic should be moved to `jstz_crypto` (especially in consideration of adding a `SecretKey` struct for signing operations). 

I recommend adding a `SecretKey` struct (in a `secret_key` module) and having a 
`keypair_with_hash_from_ikm` function exported by `jstz_crypto`.

jstz_cli/src/accounts/account.rs
```diff
@@ -0,0 +1,64 @@
...
+        secret_key: String,
+        public_key: String,
+    ) -> Self {
+        Account {
+            alias,
+            address,
+            secret_key,
+            public_key,
+        }
+    }
```
johnyob, line 32: `new` should only be implemented if you:
1. Want to keep the fields private or,
2. Non-trivial logic for initializing the struct value
Otherwise the idiomatic way is to use the Rust struct constructor e.g. `Account { ... }`. 

Given this, I think this `new` function should be removed

jstz_cli/src/main.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use clap::Parser;
 
+mod accounts;
```
johnyob, line 4: Nit: `account` not `accounts`

johnyob (changes requested):

johnyob (approved):
LGTM

PR #65, opened 2023-10-24
Proto: Coarse gas model

sam-finch-tezos:
# Context
Modifies Boa's instruction limit for use as a basic gas model

# Description
This is a draft PR as the instruction count may not be the right choice to implement gas.
* The instruction count is attached to the context, not to the realm, this means there can be a global gas limit for the entire transaction, but not individual limits for calls.
* As asynchronous calls may be interleaved it is unlikely that this issue can be solved by reducing the remaining instruction count for a subcontract call and then increasing it again.
* Modifying boa's profiler to allow custom profilers may be a better fit when ticks are introduced and possibly more acceptable for the boa team.

# Manually testing the PR
```bash
tz4=tz492MCfwp9V961DhNGmKzD642uhU8j6H5nB
cargo run -- deploy --self-address $tz4 --balance 0 --function-code "`cat examples/loop.js`"
tz4_loop=tz4XZbqPEFYCRwf4Vg18SY2wGJB1LUB5xTFK
cargo run --  cargo run -- run  "tezos://`tz4_loop`?iterations=10" $tz4 --gas-limit 500 # OK, we've got enough gas
cargo run --  cargo run -- run  "tezos://`tz4_loop`?iterations=100" $tz4 --gas-limit 500 # Error: GasLimitExceeded
```


johnyob (commented):
Partial review: Code changes look good (modulo the diff from not being rebased correctly). 

Need to test locally before I can approve

PR #67, opened 2023-10-24
Bugfix: Printing of native class prototypes

ryutamago:
# Context
There was a bug associated with converting the prototype of  `URL` and `URLSearchParam` to string. 
This is because the to_string methods we have only handles the conversion from`URL` and `URLSearchParam`, not the their prototypes. 

# Description
Made changes to the `URL` and `URLSearchParam` so that in case the js value is not of type Self, call the builtin Object's `to_string` method

# Manually testing the PR
In repl,
```
>> console.log(URL)
[🪵] function URL() { [native code] }
>> console.log(URLSearchParams)
[🪵] function URLSearchParams() { [native code] }
>> console.log(URL.prototype)
[🪵] [object Object]
>> console.log(URLSearchParams.prototype)
[🪵] [object Object]
```


johnyob (approved):
LGTM.

Please squash before merging (commit history is a little messy)

PR #69, opened 2023-10-24
Prettify Object Logs

sam-finch-tezos:
# Context

The console is not currently spec compliant. Javascript objects are logged as `[ object Object ]` rather than displaying their properties.  This PR updates that behaviour by printing the full object.

# Description

* Logging is updated to print objects longhand
* Strings within logs are escaped to avoid multiline strings
* The repl display is updated to print objects longhand
* In the repl we explicitly check if a js value is callable and revert to the previous behaviour.

# Manually testing the PR
In the repl
```javascript
> console.log({ the_answer : 42}) // { the_answer: 42 }
> console.log({multiline: "Hello\nWorld"}) // { multiline: "Hello\nWorld" }
// it's still possible to fool it
> console.log(['"', {foo: 42}, '"', {foo: 42}], "😭") // [ """, {\n foo: 42\n}, """, { foo: 42 } ] 😭
```
# Further work

* Boa's display functions are still a wip. particularly for functions and string escapes. We could contribute to updating them.
* Another idea for the future is to log the objects as tree-like value (e.g. XML/JSON etc), this could then be sent (via our logging infra) to the CLI and printed according the CLI's formatting of the tree.


johnyob:
I'm getting some rather strange logs that seem to arise from escaped characters e.g. `\n`, spaces, etc. I'm fairly certain it occurs to due to the use of `escape`

johnyob (changes requested):
I'm getting some rather strange logs that seem to arise from escaped characters e.g. \n, spaces, etc. I'm fairly certain it occurs to due to the use of `escape` -- I'd prefer to keep the change to REPL but not the `console` until this escaping is fixed

johnyob:
Closing in favour of the idea:

We want to stick with `.display`, we can enumerate lines before printing to the debug log, on each newline we insert the logging symbol + the associated indentation for the log group in the `console` API

PR #79, opened 2023-10-26
Cli: Trace

alanmarkoTrilitech:
# Context

https://app.asana.com/0/1205170694555918/1205507584925473

# Description
Adds trace command to the CLI - works similarly to ./view-console.sh.

# Manually testing the PR
`cargo run -- logs trace`

### DIFF
--- a/jstz_cli/src/main.rs
+++ b/jstz_cli/src/main.rs
@@ -7,6 +7,7 @@ mod bridge;
 mod config;
 mod deploy;
 mod jstz;
+mod logs;
 mod octez;
 mod repl;
 mod run;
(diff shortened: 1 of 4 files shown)


johnyob (changes requested):
Implementation of `trace` LGTM. 

Could you please rename this command to `jstz logs trace` to be more inline with our existing design document on logging.

jstz_cli/src/main.rs
```diff
@@ -5,10 +5,12 @@ mod account;
 mod bridge;
 mod config;
 mod deploy;
+mod logs_command;
```
johnyob, line 8: `logs_command` -> `logs`, fits better w/ existing naming

alanmarkoTrilitech: The main reason why I didn't like it was that logs is a special folder in VS Code making it a bit confusing.

alanmarkoTrilitech: And it is by default ignored in .gitignore

johnyob (approved):

PR #83, opened 2023-10-30
(DRAFT) Api: Implementation of TextEncoder

Lesenr1:
# Context

# Description

# Manually testing the PR


jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+    }
+
+    fn encoding() -> &'static str {
+        "utf-8"
+    }
+
+    fn encode(input: Option<String>) -> Uint8Array {
+        let s = input.unwrap_or_else(|| "".to_string());
+        s.into_bytes().into_boxed_slice()
+    }
```
johnyob, line 125: Definitely not spec compliant 😅 

I think in terms of sharing code between `encode` and `encode_into`, it would be best to implement their process: https://encoding.spec.whatwg.org/#concept-encoding-process.

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+                count += 1;
+                *dst_elt = *src_elt;
+            }
+        }
+        TextEncoderEncodeIntoResult {
+            read: count,
+            written: count,
+        }
+    }
+}
```
johnyob, line 144: Ditto, this is also not spec compliant

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+#[derive(Default, Clone, Trace, Finalize)]
 pub struct TextEncoderApi;
```
johnyob, line 146: `TextEncoderClass` not `TextEncoderApi`

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+        let arg = match args.get(0) {
+            None => None,
+            Some(arg) => {
+                let arg = arg
+                    .as_string()
+                    .ok_or_else(|| JsNativeError::typ().with_message("expected string"))?
+                    .to_std_string_escaped();
+                Some(arg)
+            }
+        };
```
johnyob, line 158: This should be supported by `args.get_or_undefined(0).try_js_into(context)`

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+        let source = args.get_or_undefined(0).try_js_into(context)?;
+        let destination = args.get_or_undefined(1);
+        // 2nd argument is a reference to a Uint8Array to be overwritten
+        let init: JsUint8Array = destination.try_js_into(context)?;
+        // initialize a boxed "destionation buffer" with the transpiled content of the Js object
+        let mut vec: Vec<u8> = Vec::with_capacity(init.length(context)?);
+        for index in 0..init.length(context)? {
+            vec.push(init.at(index as i64, context)?.to_uint8(context)?)
+        }
+        let destination_buffer: &mut Box<[u8]> = &mut Box::from(vec);
```
johnyob, line 181: Instead, you should convert the `JsUint8Array` to an `ArrayBuffer`:
https://docs.rs/boa_engine/latest/boa_engine/builtins/array_buffer/struct.ArrayBuffer.html.

Steps:
1. Convert to typed array data using `.as_typed_array()` (https://docs.rs/boa_engine/latest/boa_engine/object/struct.Object.html#method.as_typed_array)
2. Get array buffer view using https://docs.rs/boa_engine/latest/boa_engine/builtins/typed_array/integer_indexed_object/struct.IntegerIndexed.html#method.viewed_array_buffer.
3. Get `ArrayBuffer` from `JsObject` using https://docs.rs/boa_engine/latest/boa_engine/object/struct.Object.html#method.as_array_buffer_mut

You now have access to `array_buffer_data` :)

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
-            .function(
-                NativeFunction::from_fn_ptr(Self::atob),
-                js_string!("atob"),
-                0,
-            )
-            .function(
-                NativeFunction::from_fn_ptr(Self::btoa),
-                js_string!("btoa"),
-                0,
+impl NativeClass for TextEncoderApi {
```
johnyob, line 198: TextEncoderClass

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+        Ok(TextEncoder)
+    }
+
+    fn init(class: &mut ClassBuilder<'_, '_>) -> JsResult<()> {
+        let encoding = accessor!(
+            class.context(),
+            TextEncoder,
+            "encoding",
+            get:((_x, _context) => Ok(JsString::from(TextEncoder::encoding()).into()))
+        );
```
johnyob, line 217: Split accessors and `init`

packages/jstz-types/index.d.ts
```diff
@@ -85,7 +85,7 @@ declare interface Request extends Body {
 
 declare var Request: {
   readonly prototype: Request;
-  new (input: RequestInfo, init?: RequestInit): Request;
+  new(input: RequestInfo, init?: RequestInit): Request;
 };
```
johnyob, line 89: What is with the removal of the ` `

packages/jstz-types/index.d.ts
```diff
@@ -103,17 +103,23 @@ declare interface Response extends Body {
...
   readonly prototype: Response;
-  new (body?: BodyInit | null, init?: ResponseInit): Response;
+  new(body?: BodyInit | null, init?: ResponseInit): Response;
   json(data: unknown): Response;
   error(): Response;
 };
 
-declare interface TextEncoder {
-  atob(data: string): string;
-  btoa(data: string): string;
```
johnyob, line 113: We still want to retain `atob` and `btoa` -- please add these to the global scope.

johnyob (changes requested):

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+//   constructor();
+
+//   [NewObject] Uint8Array encode(optional USVString input = "");
+//   TextEncoderEncodeIntoResult encodeInto(USVString source, [AllowShared] Uint8Array destination);
+// };
+// TextEncoder includes TextEncoderCommon;
 
 #[derive(Trace, Finalize)]
-struct TextEncoder;
+pub struct TextEncoder;
```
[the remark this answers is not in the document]
Lesenr1: right, it doesn't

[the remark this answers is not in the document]
Lesenr1:
```
   --> jstz_api/src/text_encoder.rs:202:5
    |
37  | struct TextEncoder;
    | ------------------ `TextEncoder` declared as private
...
202 |     type Instance = TextEncoder;
    |     ^^^^^^^^^^^^^ can't leak private type

error[E0446]: private type `TextEncoder` in public interface
   --> jstz_api/src/text_encoder.rs:206:5
    |
37  |   struct TextEncoder;
    |   ------------------ `TextEncoder` declared as private
...
206 | /     fn constructor(
207 | |         _this: &JsNativeObject<TextEncoder>,
208 | |         _args: &[JsValue],
209 | |         _: &mut Context<'_>,
210 | |     ) -> JsResult<TextEncoder> {
    | |______________________________^ can't leak private type
```

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+        value
+            .as_object()
+            .and_then(|obj| obj.downcast_mut::<Self>())
+            .ok_or_else(|| {
+                JsNativeError::typ()
+                    .with_message("Failed to convert js value into rust type `TextEncoderEncodeIntoResult`")
+                    .into()
+            })
+    }
+}
```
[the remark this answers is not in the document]
Lesenr1: how can I test that ?

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+            TextEncoderEncodeIntoResult,
+            "read",
+            get:((x, _context) => Ok(JsBigInt::new(x.read).into()))
+        );
+        let written = accessor!(
+            class.context(),
+            TextEncoderEncodeIntoResult,
+            "written",
+            get:((x, _context) => Ok(JsBigInt::new(x.written).into()))
+        );
```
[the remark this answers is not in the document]
Lesenr1: Usually there is a struct `Something` and a struct `SomethingClass` where accessors are defined. could you explain that ?
Do I need `TextEncoderEncodeIntoResult` , `TextEncoderEncodeIntoResultClass` and `TextEncoderEncodeIntoResultApi` ?

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,249 @@
...
+        class
+            .accessor(js_string!("read"), read, Attribute::all())
+            .accessor(js_string!("written"), written, Attribute::all());
+
+        Ok(())
     }
 }
 
+// Note: Using Box as a way to get a fixed size array, probably wrong
+type Uint8Array = Box<[u8]>;
```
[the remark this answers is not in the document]
Lesenr1: it's fixed-length

packages/jstz-types/index.d.ts
Lesenr1: can they be expressed in term of TextEncoder.encode ?

packages/jstz-types/index.d.ts
Lesenr1: I get autoformat on files.. seems like this one got formatted, should I revert ? are you using a different formatter ?

packages/jstz-types/index.d.ts
johnyob: Maybe, but we already have existing implementations in Rust that are compliant with their Node & Deno equivalents

packages/jstz-types/index.d.ts
johnyob: Perhaps your prettier version differs from ours

jstz_api/src/text_encoder.rs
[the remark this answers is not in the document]
johnyob: Fairs 👍

jstz_api/src/text_encoder.rs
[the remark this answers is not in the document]
johnyob: On the JavaScript side, likely by structural equality on the objects.

jstz_api/src/text_encoder.rs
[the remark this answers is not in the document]
johnyob: > could you explain that ?

Provides namespacing between a Rust interface and a JavaScript interface (which either implements `NativeClass` or `Api`). 

Additionally, it separates the conversion boilerplate from the Rust implementation. Once `boa_bindgen` is implemented, we can likely delete the `_Class` or `_Api` structs. 

In your current implementation, I'd only have `TextEncoderEncodeIntoResult` and `TextEncoderEncodeIntoResultClass`.

jstz_api/src/text_encoder.rs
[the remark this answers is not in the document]
johnyob: Nonetheless, `Vec<u8>` is still likely the better choice (given my other comments below) -- given you allocate `Vec`s elsewhere + the underlying `ArrayBuffer` data is a `Vec<u8>`.  

But if we were truly in a scenario where the array's length was fixed, then `Box<[u8]>` would be the best choice.

jstz_api/src/text_encoder.rs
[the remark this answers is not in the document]
Lesenr1: done separating

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
Lesenr1: I struggle a lot to understand the spec..

> A [handler](https://encoding.spec.whatwg.org/#handler) algorithm takes an input [I/O queue](https://encoding.spec.whatwg.org/#concept-stream) and an [item](https://infra.spec.whatwg.org/#list-item), and returns finished, one or more [items](https://infra.spec.whatwg.org/#list-item), error optionally with a [code point](https://infra.spec.whatwg.org/#code-point), or continue.

what's the handler here ?

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,304 @@
...
+    fn init(class: &mut ClassBuilder<'_, '_>) -> JsResult<()> {
+        let read = Self::read(class.context());
+        let written = Self::written(class.context());
+        class
+            .accessor(js_string!("read"), read, Attribute::all())
+            .accessor(js_string!("written"), written, Attribute::all());
+
+        Ok(())
+    }
+}
```
johnyob, line 101: Let's keep this for now and add a TODO comment

Wrt to dictionaries, given their copying semantics when being passed / returned in the WHATWG spec, in future they should be Rust structs that implement `TryFromJs` and `TryIntoJs`.

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,304 @@
...
+        context: &mut Context,
+    ) -> JsResult<JsValue> {
+        let source = args
             .get_or_undefined(0)
             .as_string()
-            .ok_or_else(|| JsNativeError::typ().with_message("expected string"))?;
-        let result = TextEncoder::btoa(data)?;
-        Ok(result.into())
+            .map(|x| x.as_slice())
+            .unwrap();
```
johnyob, line 206: Avoid `.unwrap` unless the invariant is known. In this case you'd want to return a `JsError` (constructed using the `JsNativeError` API)

jstz_api/src/text_encoder.rs
```diff
@@ -1,71 +1,304 @@
...
+        let init: JsUint8Array = destination.try_js_into(context)?;
+        // initialize a boxed "destination buffer" with the transpiled content of the Js object
+        let destination: &mut Vec<u8> = &mut Vec::new();
+        // call the rust function
+        let result: TextEncoderEncodeIntoResult =
+            TextEncoder::encode_into(source, destination);
+        // write back the destination JsValue
+        for (i, value) in destination.iter().enumerate() {
+            init.fill(*value, Some(i), Some(i + 1), context)?;
+        }
```
johnyob, line 218: It would be good to get a reference to the `Vec` used by the typed array here (`destination`) instead of copying it in -- writing some utility functions for this would definitely help any work in streams (which very frequently uses the underlying buffer provided by typed arrays / array buffers / data views)

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
Lesenr1: done

jstz_api/src/text_encoder.rs
Lesenr1: tried to do that here : https://github.com/trilitech/jstz/blob/828b92bfcd1cd134324ab5f133809fef56a9dc3b/jstz_api/src/text_decoder.rs#L330-L380
But it seems that the underlying data struct. isn't always a `Vec`

Lesenr1:
closing for `remi@api/encoding`

PR #163, opened 2023-11-06
Proto: Coarse gas model

sam-finch-tezos:
# Context
This pr adds a global gas limit to jstz operations.

**Related Tasks**: [Add Gas To Transactions](https://app.asana.com/0/1205770721173531/1205771213629383/f) 

# Description

* Boa's existing instruction count used as a basic gas model
* Gas can be limited through the run contract transaction (default is `usize::MAX`)
* repl is always supplied with maximum amount of gas

# Manually testing the PR
``` bash
cargo run --bin jstz -- deploy < examples/loop.js
tz4_loop=`pbpaste` # copy new address to clipboard
cargo run --bin jstz -- run  "tezos://`tz4_loop`?iterations=10" $tz4 --gas-limit 500 # OK, we've got enough gas
cargo run --bin jstz -- run  "tezos://`tz4_loop`?iterations=100" $tz4 --gas-limit 500 # Error: GasLimitExceeded
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,10 +58,10 @@ git = "https://gitlab.com/tezos/tezos.git"
 git = "https://gitlab.com/tezos/tezos.git"
 
 [patch.crates-io]
-boa_ast = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_engine = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_gc = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_interner = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_macros = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_parser = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
-boa_profiler = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
+boa_ast = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_engine = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_gc = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_interner = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_macros = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_parser = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
+boa_profiler = { git = "https://github.com/trilitech/boa.git", branch = "sam.finch@/context/expose-instruction-count" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "ahash",
- "base64 0.21.4",
+ "base64 0.21.5",
  "bitflags 2.4.1",
  "brotli",
  "bytes",
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.38",
+ "syn 2.0.41",
 ]
 
 [[package]]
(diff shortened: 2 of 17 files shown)


jstz_cli/src/repl.rs
```diff
@@ -16,7 +16,7 @@ use crate::config::Config;
 pub fn exec(self_address: Option<String>, cfg: &Config) -> Result<()> {
     let address = &cfg.accounts.account_or_current(self_address)?.address;
 
-    let mut rt = Runtime::new().expect("Failed to create a new runtime.");
+    let mut rt = Runtime::new(usize::MAX).expect("Failed to create a new runtime.");
```
johnyob, line 19: It would be more ideal to split the runtime interface into creating a runtime w/out gas and one with. 
e.g. 
```rust
#[derive(Default)]
pub struct RuntimeOptions {
  gas_limit: Option<u32>
}

impl Runtime {
  pub fn new(options: &RuntimeOptions) -> JsResult<Self> {
     ...
  }  
}
```

jstz_cli/src/run.rs
```diff
@@ -47,6 +48,7 @@ pub async fn exec(
             method,
             headers: HeaderMap::default(),
             body,
+            gas_limit: gas_limit.try_into().unwrap_or(usize::MAX),
```
johnyob, line 51: Why not use `usize`?

johnyob (changes requested):
LGTM.

Minor suggestions on improving interfaces

jstz_cli/src/run.rs
sam-finch-tezos: On most architectures `usize` is a `u64`, but json doesn't handle `u64`s correctly so we have to pass a `u32`.
The only reason `|x : u32| -> usize x.try_into().unwrap()` would fail is on an architecture with `usize` represented as a `u16` (ie never) 
There could also be an issue as `usize` is most likely `u64` in the cli but a `u32` in wasm32. This will change with risc-v though.

jstz_cli/src/repl.rs
johnyob: I'd disagree. 
1. The struct provides a `Default` implementation for default options
2. I think it would be possible to max-out the repl by running extremely large computations on the REPL -- moreover, conceptually it makes sense for gas to be an optional part of our runtime

johnyob: > Gas is not an optional part of the rollup.

Yes, but our runtime needs to be designed to also work in a local environment (this is absolutely key)

> None is not a sensible default, it's the wrong type. The runtime requires a usize and there isn't a sensible default.

`None` is a sensible default, since the `instructions_remaining` limit can optionally be passed to the context builder. 

Additionally the `Option` provides a clearer interface, clear semantics (is metering enabled or not) & better engineering practices (avoiding max integer defaults).

sam-finch-tezos: > Yes, but our runtime needs to be designed to also work in a local environment (this is absolutely key)
This would be worth discussing, but that is a compile time decision not a runtime decision.  It would be possible to have this as a feature flag, but we can introduce this later.

> None is a sensible default, since the instructions_remaining limit can optionally be passed to the context builder.
No.  It's not an optional parameter, the boa developers are using the builder pattern which requires a default value, but not a sensible one.  If no value is supplied instructions remaining will be 0.

> Additionally the Option provides a clearer interface, clear semantics (is metering enabled or not) & better engineering practices (avoiding max integer defaults).
No. There is no default value, it is not an optional parameter.  There are special circumstances (the repl) under which we pass `usize::max` which is effectively unlimited gas.  The parameter is optional for now in the CLI to make it more convenient for us to test things.  Passing `usize::max` makes the intention far clearer than passing `None`.

johnyob: > This would be worth discussing, but that is a compile time decision not a runtime decision. It would be possible to have this as a feature flag, but we can introduce this later.

Can we introduce it now? Feature flags aren't complicated + will likely pay off in the future since `jstz_core` is likely to have feature flags as we try to improve crate separation & interfaces between them.

> No. It's not an optional parameter, the boa developers are using the builder pattern which requires a default value, but not a sensible one. If no value is supplied instructions remaining will be 0.

Ah I was unaware of this, I assumed they'd pick a better mechanism for this. 

> Passing usize::max makes the intention far clearer than passing None.

I'd disagree with this, passing `None` makes far more sense when you don't want to have a limit. Almost every metered runtime I know (including the Michelson interpreter in Tezos) provide an interface for optional gas.

sam-finch-tezos: I think we've now strayed far enough away from the original point that I can consider this matter to be closed.

Cargo.toml
```diff
@@ -58,10 +58,20 @@ git = "https://gitlab.com/tezos/tezos.git"
...
-boa_profiler = { git = "https://github.com/trilitech/boa.git", branch = "ajob410@create-realm-with-default-globals" }
+tezos-smart-rollup = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-host = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-core = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-mock = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-encoding = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-entrypoint = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-debug = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-panic-hook = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
+tezos-smart-rollup-storage = { git = "https://gitlab.com/tezos/tezos.git", branch = "ajob410@sdk-ord-paths" }
```
johnyob, line 69: I think these can be removed. We've pinned tezos deps above

johnyob (approved):
LGTM 🎉

Once build issues are fixed, this can be merged

PR #76, opened 2023-10-25
Kernel: Signed operations

johnyob:
# Context

**Related Tasks**: [Signing operations](https://app.asana.com/0/1205170694555856/1205647566435497/f)

# Description

**Dependencies**: Cli: Signed operations (@alanmarkoTrilitech )

This PR:
1. Hides proto-functions such as `run_contract` or `deploy_contract` (ensuring the only way to execute protocol operations is either a signed operation or a external operation). 
2. Removes external operation `ContractOrigination` (replaced by `DeployContract` operation)
3. Relocates deployment code to `script.rs` (consistent with `run` operation)
4. Removes `apply.rs` (obselete, since `executor` handles this logic)
5. Changes `ExternalMessage` to `SignedOperation` and `InternalMessage` to `ExternalOperation`

# Manually testing the PR

Pending CLI changes

PR #166, opened 2023-11-07
Bugfix: Fix missing `nonce.increment()`

johnyob:
# Context

Following from #76, when verifying the nonce, we do not increment it. 
This results in `InvalidNonce` errors when executing consecutive operations from a given account

# Description

This PR simply introduces a 1-liner fix that increments the nonce when running `verify_nonce`. 

# Manually testing the PR

```sh
cargo run --bin jstz -- account create alistair
cargo run --bin jstz -- login alistair
cargo run --bin jstz -- deploy examples/hello_world.js # Expected to be Ok
cargo run --bin jstz -- deploy examples/hello_world.js # Expected to be Ok
```

### DIFF
--- a/jstz_proto/src/operation.rs
+++ b/jstz_proto/src/operation.rs
@@ -39,6 +39,7 @@ impl Operation {
         let next_nonce = Account::nonce(rt, tx, &self.source)?;
 
         if self.nonce == *next_nonce {
+            next_nonce.increment();
             Ok(())
         } else {
             Err(Error::InvalidNonce)


jstz_proto/src/operation.rs
```diff
@@ -39,6 +39,7 @@ impl Operation {
         let next_nonce = Account::nonce(rt, tx, &self.source)?;
 
         if self.nonce == *next_nonce {
+            next_nonce.increment();
```
sam-finch-tezos, line 42: Just to confirm this is the correct behaviour. This will increment the nonce even if the operation later fails.

sam-finch-tezos (approved):
lgtm

PR #168, opened 2023-11-09
Node: Initial implementation of `/logs/<sf>/stream`

ryutamago:
# Description

This PR adds a log stream service that is responsible for reading the kernel.log file and broadcasting to the clients through SSE. 

**Related task**: [Name](https://app.asana.com/0/1205770721173533/1205770721347923)

First, we can start updating the kernel log file by running:
```
cargo run --bin jstz sandbox start
```
Run the jstz node with the path to the kernel file
```
cargo run --bin jstz-node --  --kernel-file-path logs/kernel.log
```
Now clients can read the logs by
```
curl http://localhost:8933/logs/foo/stream
```

# Manual testing

```sh
npm run docs:dev
```

# Checklist

- [x] Documentation has been locally tested
- [ ] **ALL** spec deviations of the implemented methods have been documented

### DIFF
--- a/jstz_node/src/main.rs
+++ b/jstz_node/src/main.rs
@@ -3,7 +3,7 @@ use std::io;
 use actix_web::{middleware::Logger, web::Data, App, HttpServer, Scope};
 use clap::Parser;
 use env_logger::Env;
-use services::{logs::stream_logs, LogService};
+use services::{logs::stream_logs, LogsService};
 use tokio_util::sync::CancellationToken;
 
 use crate::{
@@ -55,7 +55,7 @@ async fn main() -> io::Result<()> {
     let cancellation_token = CancellationToken::new();
 
     let (broadcaster, tail_file_handle) =
-        LogService::init(args.kernel_file_path, &cancellation_token);
+        LogsService::init(args.kernel_file_path, &cancellation_token);
 
     HttpServer::new(move || {
         App::new()
--- a/jstz_node/src/main.rs
+++ b/jstz_node/src/main.rs
@@ -1,8 +1,10 @@
 use std::io;
 
-use actix_web::{middleware::Logger, web::Data, App, HttpServer};
+use actix_web::{middleware::Logger, web::Data, App, HttpServer, Scope};
 use clap::Parser;
 use env_logger::Env;
+use services::{logs::stream_logs, LogService};
+use tokio_util::sync::CancellationToken;
 
 use crate::{
     rollup::RollupClient,
@@ -13,6 +15,7 @@ pub use error::{Error, Result};
 mod error;
 mod rollup;
 mod services;
+mod tailed_file;
 
 /// Endpoint details for the `octez-smart-rollup-node`
 const DEFAULT_ROLLUP_NODE_RPC_ADDR: &'static str = "127.0.0.1";
@@ -31,6 +34,9 @@ struct Args {
 
     #[arg(short, long)]
     rollup_endpoint: Option<String>,
+
+    #[arg(long)]
+    kernel_file_path: String,
 }
 
 #[actix_web::main]
@@ -46,14 +52,30 @@ async fn main() -> io::Result<()> {
 
     let rollup_client = Data::new(RollupClient::new(rollup_endpoint));
 
+    let cancellation_token = CancellationToken::new();
+
+    let (broadcaster, tail_file_handle) =
+        LogService::init(args.kernel_file_path, &cancellation_token);
+
     HttpServer::new(move || {
         App::new()
             .app_data(rollup_client.clone())
+            .service(
+                Scope::new("/logs")
+                    .app_data(Data::from(broadcaster.clone()))
+                    .service(stream_logs),
+            )
             .wrap(Logger::default())
             .configure(OperationsService::configure)
             .configure(AccountsService::configure)
     })
     .bind(ENDPOINT)?
     .run()
-    .await
+    .await?;
+
+    cancellation_token.cancel();
+
+    tail_file_handle.await.unwrap()?;
+
+    Ok(())
 }
--- a/jstz_node/src/services/logs/broadcaster.rs
+++ b/jstz_node/src/services/logs/broadcaster.rs
@@ -1,0 +1,91 @@
+use std::{sync::Arc, time::Duration};
+
+use actix_web::rt::time::interval;
+use actix_web_lab::{
+    sse::{self, Sse},
+    util::InfallibleStream,
+};
+use futures_util::future;
+use parking_lot::Mutex;
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+
+pub struct Broadcaster {
+    clients: Mutex<Vec<mpsc::Sender<sse::Event>>>, // TODO: Use a read-write lock instead?
+}
+
+// Pings clients every 10 seconds
+const PING_INTERVAL: u64 = 10;
+
+impl Broadcaster {
+    /// Constructs new broadcaster and spawns ping loop responsible for removing stale clients.
+    pub fn create() -> Arc<Self> {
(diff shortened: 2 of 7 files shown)


jstz_node/src/main.rs
```diff
@@ -46,14 +51,28 @@ async fn main() -> io::Result<()> {
 
     let rollup_client = Data::new(RollupClient::new(rollup_endpoint));
 
+    let (broadcaster, handler, cancel_read_file) =
+        LogService::init(args.kernel_file_path);
+
```
johnyob, line 56: Not necessarily in this PR, but it would be nice to come up w/ a trait for services

jstz_node/src/main.rs
```diff
@@ -46,14 +51,28 @@ async fn main() -> io::Result<()> {
...
     HttpServer::new(move || {
         App::new()
             .app_data(rollup_client.clone())
+            .service(
+                Scope::new("/logs")
+                    .app_data(Data::from(broadcaster.clone()))
```
johnyob, line 62: Data needs to be created outside the closure of `HttoServer::new` -- see https://docs.rs/actix-web/latest/actix_web/struct.App.html#shared-mutable-state

johnyob: Additionally, can this be added to the top-level `.app_data` (for consistency)?

jstz_node/src/services/logs/broadcaster.rs
```diff
@@ -0,0 +1,91 @@
...
+    util::InfallibleStream,
+};
+use futures_util::future;
+use parking_lot::Mutex;
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+
+pub struct Broadcaster {
+    clients: Mutex<Vec<mpsc::Sender<sse::Event>>>, // TODO: Use a read-write lock instead?
+}
```
johnyob, line 15: Is it possible to put the `Arc` inside this struct? Like:
```suggestion
pub struct Broadcaster {
    clients: Arc<Mutex<Vec<mpsc::Sender<sse::Event>>>>,
}
```

johnyob: Additionally, `Broadcaster` could then implement a cheap `Clone`

jstz_node/src/services/logs/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        let broadcaster = Broadcaster::create();
+
+        let read_file_cancel = CancellationToken::new();
+
+        let read_kernel_handle: JoinHandle<std::io::Result<()>> = actix_web::rt::spawn(
+            Self::read_file(path, Arc::clone(&broadcaster), read_file_cancel.clone()),
+        );
+
+        (broadcaster, read_kernel_handle, read_file_cancel)
+    }
```
johnyob, line 47: Parts of this feel slightly messy to me 🤔 

Here are some drive-by thoughts (that don't necessarily need addressing in this PR):
- `jstz_node` will likely have global 1 cancellation token for `n` of it's background tasks. 
-  Each background task will typically be a closure that takes `n` arguments (one of them being the cancellation token)
- `.init` should provide a way to add `app_data` instead of returning the data. `Data` can be local to a service or a collection of services and I think the trait we eventually define for services should take this into account.

jstz_node/src/services/logs/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    async fn read_file(
```
johnyob, line 49:
```suggestion
    async fn tail_kernel_logs_file(
```

jstz_node/src/tailed_file.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        return Ok(TailedFile(reader));
+    }
+}
+
+impl AsyncRead for TailedFile {
+    fn poll_read(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
```
johnyob, line 24: Use imports to avoid the `std::...` and `tokio::...` prefixes

jstz_node/src/tailed_file.rs
```diff
@@ -0,0 +1,44 @@
...
+        let this = Pin::new(&mut self.get_mut().0);
+        this.poll_read(cx, buf)
+    }
+}
+
+impl AsyncBufRead for TailedFile {
+    fn poll_fill_buf(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<std::io::Result<&[u8]>> {
```
johnyob, line 35: Ditto

jstz_node/src/tailed_file.rs
```diff
@@ -0,0 +1,44 @@
...
+        let this = Pin::new(&mut self.get_mut().0);
+        this.poll_fill_buf(cx)
+    }
+
+    fn consume(self: std::pin::Pin<&mut Self>, amt: usize) {
+        let this = Pin::new(&mut self.get_mut().0);
+        this.consume(amt)
```
johnyob, line 42: Ditto

jstz_node/src/tailed_file.rs
```diff
@@ -0,0 +1,44 @@
+use std::io::SeekFrom;
```
johnyob, line 1: Is there much point of re-implementing the traits?

Could we just have a newtype that is `BufReader<File>` that implements `Deref` and `DerefMut`.

https://doc.rust-lang.org/stable/core/ops/trait.Deref.html

johnyob (commented):
Some initial drive by thoughts, looking good so far

jstz_node/src/main.rs
ryutamago: > Data needs to be created outside the closure of `HttoServer::new` -- see https://docs.rs/actix-web/latest/actix_web/struct.App.html#shared-mutable-state

The broadcaster is created outside of Server, it is of `Arc` so can be shared across workers :)

ryutamago: > Additionally, can this be added to the top-level `.app_data` (for consistency)?

hmm in that case other services like operation and account can access the broadcaster. I thought it would be better to limit the scope? wdyt?

jstz_node/src/services/logs/broadcaster.rs
ryutamago: I've tried but in that case we will end up having two `Arc` since `Data` is esenttialy an `Arc`
```
pub struct Broadcaster {
    clients: Arc<Mutex<Vec<mpsc::Sender<sse::Event>>>>,
}

impl<T> Data<T> {
    /// Create new `Data` instance.
    pub fn new(state: T) -> Data<T> {
        Data(Arc::new(state))
    }
}
```
so
```
Data::new(Broadcaster)
```
is going to be 
```
Arc<Boradcaster {cllients:Arc::new(...)}>
```

jstz_node/src/services/logs/mod.rs
ryutamago: it could technically tail any file of the given path so i will keep it as `tail__file` ?

jstz_node/src/tailed_file.rs
ryutamago: good point, idt we even need the Deref/DefreMut here. we can just reimplement the lines method

jstz_node/src/services/mod.rs
```diff
@@ -1,5 +1,8 @@
 mod accounts;
+pub mod logs;
 mod operations;
 
 pub use accounts::AccountsService;
 pub use operations::OperationsService;
+
+pub use logs::LogService;
```
johnyob, line 8: Nit: `LogsService` (more consistent w/ others)

johnyob (approved):
LGTM.

Minor nit needs addressing, after that can be merged :)

PR #172, opened 2023-11-10
Api: Implement `.getSetCookie` for Headers

tj825:
# Description

**Related issue**: #139 

**Dependencies**:

Tests?

# Manual testing

```sh
nix develop
cargo run --bin jstz -- repl
>> h = new Headers()
{

}
>> h.append('foo', 'bar')
>> h.append('foo', 'baz')
>> h.append('Set-cookie', 'foo')
>> h.append('SET-COOKIE', 'bar')
>> h.get('foo')
"bar, baz"
>> h.get('bar')
null
>> h.get('Set-CoOkIe')
"foo, bar"
>> h.getSetCookie()
[ "foo", "bar" ]
>> h.set('set-cookie', 'blah')
>> h.getSetCookie()
[ "blah" ]
>> new Headers().getSetCookie()
[]
```

# Testing

# Checklist

- [ ] Changes follow the existing code style (use `make fmt-check` to check)
- [ ] Testing instructions have been added to PR

### DIFF
--- a/examples/fa2/package.json
+++ b/examples/fa2/package.json
@@ -8,7 +8,7 @@
     "@tezos/jstz": "^0.0.0"
   },
   "devDependencies": {
-    "esbuild": "^0.19.5"
+    "esbuild": "^0.19.6"
   },
   "scripts": {
     "check:types": "tsc --noEmit",
--- a/examples/hello-world/package.json
+++ b/examples/hello-world/package.json
@@ -8,7 +8,7 @@
     "@tezos/jstz": "^0.0.0"
   },
   "devDependencies": {
-    "esbuild": "^0.19.5"
+    "esbuild": "^0.19.6"
   },
   "scripts": {
     "build": "esbuild index.ts --bundle --format=esm --target=esnext --minify --outfile=dist/index.js"
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@tezos/jstz-types": "^0.0.0"
       },
       "devDependencies": {
-        "esbuild": "^0.19.5",
+        "esbuild": "^0.19.6",
         "prettier": "^3.1.0",
         "typescript": "^5.2.2",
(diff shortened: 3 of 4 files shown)


jstz_api/src/http/header.rs
```diff
@@ -103,27 +99,21 @@ impl Headers {
...
+        let headers = collect_headers(self.headers.get_all(name))?;
 
-                Ok(Some(Header::Multiple(values)))
-            }
-            _ => todo!(),
+        match headers.len() {
+            0 => Ok(None),
+            1 => Ok(Some(headers.get(0).expect("infallible").into())),
+            _ => Ok(Some(headers.join(", "))),
         }
```
johnyob, line 110: I think this is semantically the same:
```suggestion
        match headers.len() {
            0 => Ok(None),
            _ => Ok(Some(headers.join(", "))),
        }
```

johnyob: For collecting headers, I think it would still be useful to have a `Header` struct:
```rust
pub struct Header {
  pub headers: Vec<String> // non-empty
}

impl Header {
  pub fn try_from_iter<T>(iter: T) -> JsResult<Self> where T: IntoIterator<Item = A> {
    // Code from `collect_headers`
  }
}

// Default conversion by joining w/ ", "
impl IntoJs for Header {
  ...
}
```

jstz_api/src/http/header.rs
```diff
@@ -203,11 +193,25 @@ impl HeadersClass {
         let header = headers.get(&name)?;
 
         Ok(match header {
-            Some(header) => header.into_js(context),
+            Some(header) => JsValue::String(JsString::from(header)),
```
johnyob, line 196: Why doesn't `into_js` work?

johnyob (changes requested):

johnyob (approved):

PR #175, opened 2023-11-13
Node: KV getter

alanmarkoTrilitech:
# Context

https://app.asana.com/0/1205770721173533/1205771213624617

# Description
This PR adds KV getter to the node, supporting 'kv list' and 'kv get' commands for reading values from the KV storage. It supports --name option to specify either account alias or address.

# Manually testing the PR
```
jstz deploy --name cntr < counter.js
jstz run "tezos://cntr/"
jstz kv list --name cnt
jstz kv get counter --name cnt
```


jstz_node/src/rollup.rs
```diff
@@ -58,4 +60,43 @@ impl RollupClient {
...
+                    match list_of_strings {
+                        Ok(list) => Ok(Some(list)),
+                        Err(e) => Err(e),
+                    }
+                }
+                Ok(_) => Err(anyhow!(
+                    "Expected a JSON array but got a different structure"
+                )),
+                Err(e) => Err(anyhow!("Failed to parse content as JSON: {:?}", e)),
+            }
```
johnyob, line 97: You're manually deserializing here. It would be best to write a rust type (e.g. `SubkeysResponse`) that derives a serializer and deserializer for this type (like `ValueResponse`)

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +31,63 @@ async fn nonce(
     Ok(HttpResponse::Ok().json(nonce))
 }
 
+#[get("/{address}/kv/{key:.*}")]
```
johnyob, line 34: Naive question: What is the `.*` for?

jstz_node/src/services/accounts.rs
```diff
@@ -13,6 +14,7 @@ async fn nonce(
     rollup_client: Data<RollupClient>,
     path: Path<String>,
 ) -> Result<impl Responder> {
+    print!("nonce ");
```
johnyob, line 17:
```suggestion
```

johnyob (changes requested):

jstz_node/src/services/accounts.rs
alanmarkoTrilitech: Key can be empty - .* ensures that it is matched even then.

jstz_cli/src/kv.rs
```diff
@@ -0,0 +1,84 @@
...
+    cfg: &mut Config,
+) -> Result<()> {
+    let jstz_client = JstzClient::new(cfg);
+
+    let address = cfg.accounts.get_address_from(alias)?;
+
+    let key_string = match key {
+        Some(key) => key,
+        None => "".to_string(),
+    };
```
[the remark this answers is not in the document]
alanmarkoTrilitech: Yes, why not? it specifies the the subset of keys that we want to list.

jstz_node/src/services/accounts.rs
johnyob: Keys cannot be empty, I've just checked. 

For example, if you deploy:
```js
const KEY = "";

const handler = () => {
  let counter = Kv.get(KEY);
  console.log(`Counter: ${counter}`);
  if (counter === null) {
    counter = 0;
  } else {
    counter++;
  }
  Kv.set(KEY, counter);
  return new Response();
};

export default handler;
```
and then execute it, you obtain a `Err("EvalError: PathError: PathError::InvalidEmptyStep")` in the receipt.

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +30,63 @@ async fn nonce(
...
+    rollup_client: Data<RollupClient>,
+    path: Path<(String, String)>,
+) -> Result<impl Responder> {
+    let (address, key) = path.into_inner();
+
+    let storage_key = if key == "EMPTY" {
+        format!("/jstz_kv/{}", address)
+    } else {
+        format!("/jstz_kv/{}/{}", address, key)
+    };
```
johnyob, line 70: I'm not a massive fan of this. I think it is good for listing the root keys, but it feels messy 🤔

johnyob: Additionally keys like `TABLE/KEY` don't get matched by this endpoint

jstz_cli/src/kv.rs
```diff
@@ -0,0 +1,84 @@
...
+pub enum Command {
+    /// Get value for a key
+    Get {
+        /// Key
+        #[arg(value_name = "KEY")]
+        key: String,
+
+        /// User address or alias
+        #[arg(short, long, value_name = "ALIAS|ADDRESS")]
+        from: Option<String>,
```
johnyob, line 65: `from` is a confusing argument. Something like `--account` would be clearer

johnyob (changes requested):

jstz_node/src/services/accounts.rs
johnyob: Perhaps doing `/kv?key={}` would be best for us and `/kv/subkeys?key={}` 🤔

jstz_node/src/rollup.rs
```diff
@@ -17,11 +17,15 @@ pub struct ValueError {
 
 #[derive(Deserialize)]
 #[serde(untagged)]
+#[derive(Debug)]
```
johnyob, line 20: Nit: Multiple derives

```suggestion
#[derive(Debug, Deserialize)]
#[serde(untagged)]
```

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +32,67 @@ async fn nonce(
...
+    query: web::Query<HashMap<String, String>>,
+) -> Result<impl Responder> {
+    let address = path.into_inner();
+    let key = query.get("key").cloned().unwrap_or_else(|| "".to_string());
+
+    let storage_key = if key == "" {
+        format!("/jstz_kv/{}", address)
+    } else {
+        format!("/jstz_kv/{}/{}", address, key)
+    };
```
johnyob, line 48: Nit: instead of using a sentinel value (`""`) to indicate empty, we could use the `Some/None` returned by `query.get`

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +32,67 @@ async fn nonce(
...
```
johnyob, line 76: Ditto

johnyob (changes requested):

johnyob:
Overall almost mergable -- my main critic is the handling of the 'root' key vs normal keys is a bit messy. 
For root, I'd expect paths:
- `/accounts/<address>/kv`
- `/accounts/<address>/kv/subkeys`
instead of:
- `/accounts/<address>/kv?key=`
- `/accounts/<address>/kv/subkeys?key=`

In the client, I'd expect an enum for keys (either `Root` or `Key(String)` or simply reuse `Option<String>`). The API can then simply take `Into<Key>` (with `Option<String>` implementing `Into<Key>`)

alanmarkoTrilitech:
I've addressed the Nits. I don't think an enum fits there well. I've used Option<String> instead and refactored it so that there is less duplicated code.

johnyob:
> I've used Option instead and refactored it so that there is less duplicated code.

Sweet! I only suggested an enum bcs I thought the semantic differences between `Some/None` vs `Key/Root` wasn't clear, but I'm also happy just using an `Option`.

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +39,60 @@ async fn nonce(
...
 }
 
+#[get("/{address}/kv")]
+async fn kv(
+    rollup_client: Data<RollupClient>,
+    path: Path<String>,
+    query: web::Query<HashMap<String, String>>,
+) -> Result<impl Responder> {
+    let address = path.into_inner();
+    let key_option = query.get("key").cloned();
```
johnyob, line 49: Nit: `.as_ref()` provides a `Option<&str>` which would avoid some clones
NB: Don't commit suggestion, other changes are required for this

```suggestion
    let key_option = query.get("key").as_ref();
```

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +39,60 @@ async fn nonce(
     Ok(HttpResponse::Ok().json(nonce))
 }
 
+#[get("/{address}/kv")]
+async fn kv(
+    rollup_client: Data<RollupClient>,
+    path: Path<String>,
+    query: web::Query<HashMap<String, String>>,
```
johnyob, line 46: Nit: Here I think it is clearer (and more consistent) to use a top-level import (avoiding the qualified `web`)

```suggestion
    query: Query<HashMap<String, String>>,
```

jstz_node/src/services/accounts.rs
```diff
@@ -29,11 +39,60 @@ async fn nonce(
...
+) -> Result<impl Responder> {
+    let address = path.into_inner();
+    let key_option = query.get("key").cloned();
+
+    let storage_key = construct_storage_key(&address, &key_option);
+
+    let value = rollup_client.get_value(&storage_key).await?;
+
+    println!("value: {:?}", value);
```
johnyob, line 55:
```suggestion
```

johnyob (approved):
LGTM 🎉 

Just some v minor nits / suggestions

PR #176, opened 2023-11-13
Testing: WPT harness

johnyob:
# Context

**Related Tasks**: [Implement testing infra for `jstz` APIs](https://app.asana.com/0/1205770721173531/1205782959048392/f)

Currently, we have _no_ tests for `jstz_api`. This poses a problem wrt. specification compliance of our web API implementations. Our goal is to achieve confidence in the compliance of our implementations. Luckily we can do so with _web platform tests_  (https://web-platform-tests.org/). 

Web platform tests (WPT) is a test suite created by various browser implementations (Chrome, etc) for testing the compliance of their web API implementations. To run these tests, we run the _wpt server_ (using `wpt serve`). 
Once running, a browser implementation can visit `.html` files that run the tests. 

# Description

In order to adapt WPT tests to `jstz`'s environment, we need to be able to run and interact with the WPT server from Rust -- this is the primary purpose of the `jstz_wpt` crate. 

The rough (slightly abstracted) workflow of running WPT tests in `jstz` is as follows:
1. Read the WPT `manifest.json`, this provides a list of URLs for tests to run. They can be filtered using the `TestFilter`. 
2. For each test URL `r`:
    1. Fetch the `.html` page from `r` from the WPT server, scraping all `script` tags 
    2. Fetch the `.js` files for each script. One special _script_ that is included in every test is the _test harness_ (https://web-platform-tests.org/writing-tests/testharness-api.html). `jstz` implements this harness as a native Rust API, permitting us to read the test results from Rust. 
    3. Bundle scripts into a single `.js` file
    4. Run the `.js` file using `jstz`, reading the test results
    5. Write the results to a `WptReport`
3. Compare newly generated report against the report stored in `wptreport.json`

We utilise Rust's expect-test framework to permit automatic updating of the `wptreport.json` file using:
```sh
```

## Setting up WPT

Simply run:
```sh
cargo run --bin jstz-wpt -- init
```

If this successfully initializes WPT, then you should be able to run:
```sh
cargo run --bin jstz-wpt -- serve
```

# Manually testing the PR

```sh
cargo run --bin jstz-wpt -- init
cargo run --bin jstz-wpt -- serve
cargo test --test wpt
```

# Known limitations

There are still several issues with `jstz-wpt`:
1. Running tests is _slow_  (+ 60 seconds) -- this is likely because we are running tests sequentially. This will be triaged and fixed in a follow-up PR. 
2. Some tests require _fetch_ to retrieve `.json` files. This can be fixed in a follow-up PR by simply reading the file from the `wpt/` directory.


jstz_api/tests/wpt.rs
```diff
@@ -0,0 +1,326 @@
...
+            &[test_completion_callback.into()]
+        );
+    }
+}
+
+pub fn run_wpt_test_harness(bundle: &Bundle) -> JsResult<Box<TestHarnessReport>> {
+    let mut rt = Runtime::new()?;
+
+    // Initialize the host-defined object with the test harness report
+    {
```
Lesenr1, line 242: shouldn't we register the API here ?

This worked (i.e. has visible effects on the result)
```
rt.realm()
        .clone()
        .register_api(jstz_api::TextEncoderApi, rt.context());
```

Lesenr1 (approved):
Feels nice to use.
I just reviewed `jstz_api/tests/wpt.rs`

alanmarkoTrilitech:
I can't run serve:

```
Submodule 'jstz_wpt/wpt' (git@github.com:web-platform-tests/wpt.git) registered for path 'jstz_wpt/wpt'
Cloning into '/Users/alanmarko/projects/jstz_attempt2/jstz/jstz_wpt/wpt'...
Submodule path 'jstz_wpt/wpt': checked out '9c5a8826e241425c5f3f53cd9426c3cc4670cf5c'
Initialized WPT git submodule
192:jstz alanmarko$ cargo run --bin jstz-wpt -- serve
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/jstz-wpt serve`
Error: Failed to start wpt server
```

Lesenr1:
I noticed that if I am using the `todo!()` macro somewhere in my API implementation and run the test suite with `env UPDATE_EXPECT=1 cargo test --test wpt`, the report (`wptreport.json`) isn't updated (even if some of the tests passed)

It would be very nice if we could do something about that

jstz_api/tests/wpt.rs
```diff
@@ -0,0 +1,326 @@
...
+            BundleItem::Inline(script) | BundleItem::Resource(_, script) => {
+                rt.context().eval(Source::from_bytes(&format!(
+                    "(function() {{ {} }})()",
+                    script
+                )))?;
+
+                println!("done")
+            }
+        }
+    }
```
Lesenr1, line 274: I reported a problem on slack [here](https://tezos-dev.slack.com/archives/C061SSDBN69/p1700144220479869) . 
 
I am not sure at all... but in the `Resource` case, shouldn't it be ?
```
 rt.context().eval(Source::from_bytes(script))?;
 ``` 
  
 I did that on my branch and not only it managed to define `createBuffer`, it also added more tests to the list.

File: jstz_api/tests/wptreport.json [hunk 3]
```diff
@@ -0,0 +1,1176 @@
...
+                    "status": "Fail",
+                    "message": "TextDecoder is not defined"
+                  }
+                ],
+                "status": "Err",
+                "metrics": {
+                  "passed": 0,
+                  "failed": 6,
+                  "timed_out": 0
+                }
```
Lesenr1, line 46: question: What does it mean for the status to be "Err" here ?

johnyob: It means that the overall test failed (in this case bcs all the subtests failed)

johnyob:
> I noticed that if I am using the `todo!()` macro somewhere in my API implementation and run the test suite with `env UPDATE_EXPECT=1 cargo test --test wpt`, the report (`wptreport.json`) isn't updated (even if some of the tests passed)
> 
> It would be very nice if we could do something about that

This isn't feasible as `todo!()` causes a panic. What I'd recommend is perhaps writing a `js_todo!()` macro that returns some of `JsResult<T>` using `JsNativeError`.

johnyob:
> I can't run serve:
> 
> ```
> Submodule 'jstz_wpt/wpt' (git@github.com:web-platform-tests/wpt.git) registered for path 'jstz_wpt/wpt'
> Cloning into '/Users/alanmarko/projects/jstz_attempt2/jstz/jstz_wpt/wpt'...
> Submodule path 'jstz_wpt/wpt': checked out '9c5a8826e241425c5f3f53cd9426c3cc4670cf5c'
> Initialized WPT git submodule
> 192:jstz alanmarko$ cargo run --bin jstz-wpt -- serve
>     Finished dev [unoptimized + debuginfo] target(s) in 0.42s
>      Running `target/debug/jstz-wpt serve`
> Error: Failed to start wpt server
> ```

Solved while pairing with @alanmarkoTrilitech. There is a bug in the `doctor` function in `jstz_wpt` that doesn't read the `/etc/hosts` file but instead reads the `/jstz_wpt/hosts` file.

jstz_api/tests/wpt.rs
johnyob: Yes. I think perhaps making a function `register_apis` that is passed a context would be best.

jstz_api/tests/wpt.rs
johnyob: I'll investigate this more. My intention was to avoid global bindings accidentally leaking, but perhaps my implementation was incorrect.

jstz_api/tests/wptreport.json
Lesenr1: Uugh sorry I commented the wrong place.
Here is a better example from my branch [here](https://github.com/trilitech/jstz/blob/c49e6b3421c99acc72f99487fc8183f0b9f4c21f/jstz_api/tests/wptreport.json#L447-L477) for which all the subtests pass, but I still have an Err status

```
"textdecoder-byte-order-marks.any.js": {
          "Test": {
            "variations": [
              {
                "subtests": [
                  {
                    "name": "Byte-order marks: utf-8",
                    "status": "Pass",
                    "message": null
                  },
                  {
                    "name": "Byte-order marks: utf-16le",
                    "status": "Pass",
                    "message": null
                  },
                  {
                    "name": "Byte-order marks: utf-16be",
                    "status": "Pass",
                    "message": null
                  }
                ],
                "status": "Err",
                "metrics": {
                  "passed": 3,
                  "failed": 0,
                  "timed_out": 0
                }
              }
            ]
          }
        }
```

Lesenr1: 🤔 https://github.com/trilitech/jstz/blob/c49e6b3421c99acc72f99487fc8183f0b9f4c21f/jstz_api/tests/wpt.rs#L321-L327

johnyob: In this case, it might be due to the fact that the test harness doesn't call the completion callback for some reason 🤔

johnyob: In this case, I will modify it to have a `Option`al status

johnyob:
@Lesenr1 This PR is now ready to merge. Please could to manually test the changes I've made before we merge :)

jstz_api/tests/wpt.rs
johnyob: Added this function ^

jstz_api/tests/wpt.rs
johnyob: I decided that your suggestion is a better fit for bundling

jstz_api/tests/wptreport.json
johnyob: Added the `WptTestStatus::Null` variant

PR #178, opened 2023-11-15
Docs: Asset bridge and deposits

johnyob:
# Context

**Related Tasks**: [Guide doc: Deposits](https://app.asana.com/0/1205770721173531/1205770721324127/f)

# Description

This PR:
- Adds some output for running `jstz bridge deposit`
- Adds a quick start for jstz's asset bridge and depositing using the CLI. 

Documentation is kept short as this is likely to change in the future. 

# Manually testing the PR

```sh
npm run docs:dev
```

### DIFF
--- a/docs/bridge.md
+++ b/docs/bridge.md
@@ -1,0 +1,41 @@
+# 💸 Asset Bridge
+
+`jstz` maintains a persistent ledger of all accounts and their balances of L2 tez (stored as mutez).
+
+The `jstz` _bridge_ implements a bridge protocol that allows to transfer fungible tokens (namely [CTEZ](https://ctez.app/)) from Tezos to the `jstz` rollup (and back *soon*™️).
+
+::: danger
+⚠️ Withdrawals from `jstz` to Tezos is not supported ⚠️
+:::
+
+## Quick Start
+
+The `jstz` CLI empowers you to effortlessly transfer assets between a Tezos address (`tz1`) and a `jstz` L2 address (`tz4`) using the provided `bridge` commands.
+
+To deposit assets from a Tezos address to a `jstz` L2 address, run the following command:
+
+```bash
+jstz bridge deposit --from <TZ1_ADDRESS/ALIAS> --to <TZ4_ADDRESS/ALIAS> --amount <AMOUNT>
+```
+
+Replace `<TZ1_ADDRESS/ALIAS>` with the source Tezos address or alias (managed by `octez-client`), `<TZ4_ADDRESS/ALIAS>` with the destination `jstz` address, and `<AMOUNT>` with the quantity of CTEZ to deposit.
+
+For example, running:
+
+```bash
+jstz bridge deposit --from tz1faswCTDciRzE4oJ9jn2Vm2dvjeyA9fUzU \
+    --to tz4N7y3T2e2dfCyHB1Ama68jnt3Fps7Ufu6d \
+    --amount 42
+```
+
+sucessfully deposits 42 CTEZ from `tz1faswCTDciRzE4oJ9jn2Vm2dvjeyA9fUzU` to the `tz4N7y3T2e2dfCyHB1Ama68jnt3Fps7Ufu6d` `jstz` address, outputting:
+
+```
+Deposited 42 CTEZ to tz4N7y3T2e2dfCyHB1Ama68jnt3Fps7Ufu6d
+```
+
+## How it Works?
+
+::: danger
+⚠️ Under construction ⚠️
+:::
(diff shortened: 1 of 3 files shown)


docs/bridge.md
```diff
@@ -0,0 +1,41 @@
...
+
+`jstz` maintains a persistent ledger of all accounts and their balances of L2 tez (stored as mutez).
+
+The `jstz` _bridge_ implements a bridge protocol that allows to transfer fungible tokens (namely [CTEZ](https://ctez.app/)) from Tezos to the `jstz` rollup (and back *soon*™️).
+
+::: danger
+⚠️ Withdrawals from `jstz` to Tezos is not supported ⚠️
+:::
+
+## Quick Start
```
alanmarkoTrilitech, line 11: Overall this seems to be a bit redundant as a very similar description is already in the CLI.

alanmarkoTrilitech (changes requested):
This diff seems a bit unfinished. It essentially doesn't add much more information than what is already in the CLI docs and it seems to me better to instead just extend the CLI doc with some description and then adding a "How it works?" doc later, so that we don't have to maintain two different descriptions of the interface.

johnyob:
> It essentially doesn't add much more information than what is already in the CLI docs, therefore it seems to me better to instead just extend the CLI doc and then adding a "How it works?" doc later, so that we don't have to maintain two different descriptions of the interface.

The CLI doc will likely evolve into a reference with examples, rather like: https://developers.cloudflare.com/workers/wrangler/commands/. 

This doc is specifically designed to explain the asset bridge (which will be rather complex in future) + to quickly demonstrate how to transfer assets between L1 and L2. Additionally, repetition is acceptable in documentation (we want to optimise for reader experience over redundancy).

alanmarkoTrilitech:
> The CLI doc will likely evolve into a reference with examples, rather like: https://developers.cloudflare.com/workers/wrangler/commands/.
> 
> This doc is specifically designed to explain the asset bridge (which will be rather complex in future) + to quickly demonstrate how to transfer assets between L1 and L2. Additionally, repetition is acceptable in documentation (we want to optimise for reader experience over redundancy).

My understanding was that after we write the documentation, the core parts will be edited/possibly rewritten by a technical writer to make it readable and understandable for developers(as a perspective of someone reading/using the docs might be very different from those writing them) and he will then optimize for readability. 

I was hoping, that since jstz should be simple to use, that in the final versions we could just make a very simple tutorial on a one longer page, and the JS developer can just scroll through it and understand everything without having to click through anything(besides API docs when needed).

Also it seems to me like there are two different goals that the docs trying to achieve - both explaining how to use the asset bridge and how it works. My understanding is that it wouldn't be too complicated to use deposit+withdrawal of assets(not sure if there are any other assets going to be included besides tez). Of course the "How it works?" might be more complex, but it is not clear to me why the developers would need to read through it. 

Can you please clarify it?

johnyob:
> My understanding was that after we write the documentation, the core parts will be edited/possibly rewritten by a technical writer to make it readable and understandable for developers (as a perspective of someone reading/using the docs might be very different from those writing them) and he will then optimize for readability.

The timeline (not set in stone) for jstz will likely be as follows:
- The mondaynet launch(es) (using WASM rollups) (Q4/2023)
- The alphanet launch(es) (on mondaynet using RISC-V rollups) (Q1/2024?)
- TezDev demo (Q2/2024)
- The betanet launch (likely on P/Q-net once RISC-V rollups have matured)
- The mainnet launch

It is unclear when developer success will aid us in writing/rewrite our documentation, but it will not be in time for the mondaynet launch this quarter. Given this, we still need to optimize the documentation we currently write for readability. 

> I was hoping, that since jstz should be simple to use, that in the final versions we could just make a very simple tutorial on a one longer page, and the JS developer can just scroll through it and understand everything without having to click through anything(besides API docs when needed). 

A single page seems optimistic. What we'll eventually end up with is a set of simple step-by-step tutorials in a 'getting started' section.

> Also it seems to me like there are two different goals that the docs trying to achieve - both explaining how to use the asset bridge and how it works. My understanding is that it wouldn't be too complicated to use deposit+withdrawal of assets(not sure if there are any other assets going to be included besides tez). Of course the "How it works?" might be more complex, but it is not clear to me why the developers would need to read through it.

> My understanding is that it wouldn't be too complicated

I think this assumption could be fatal. Tezos's documentation has historically taken this premise, often to the detriment of DX. While it is fairly simple to us, it may not be for our users (existing Web3 devs _+ Web2 devs_).

> but it is not clear to me why the developers would need to read through it.

While most developers wouldn't read through the 'How it works' section, I still think it is important to document and detail critical parts of our infrastructure (both internally for developer success and for our early users).

alanmarkoTrilitech (approved):
LGTM! I'm not sure though if "managed by `octez-client`" is correct as the aliases are currently stored just locally by jstz CLI.

johnyob:
> I'm not sure though if "managed by octez-client" is correct as the aliases are currently stored just locally by jstz CLI.

L1 aliases (accounts w/ addresses beginning w/ tz1) are managed by octez-client, whereas jstz CLI only managed accounts beginning w/ tz4.

PR #198, opened 2023-11-23
Node & API : Filter logs from `kernel.log` by address 

ryutamago:
# Context

[Filter kernel.log by address](https://app.asana.com/0/1205770721173531/1205770721347931/f)

# Description

This PR enables the jstz_node endpoint `logs/smart_function/stream` to stream the log relevant to the provided address of the smart function. the endpoint will stream logs in the a json format. 
Example:
```
data: {"contract_address":{"Tz4":"tz4EWCakzUFzj5gHsZZ7dWAkHANxLXTBdpwf"},"request_id":"b9ca156c0ddea66dd8848c5ae298ddf2ac31631eb61f3d48e797bc7d060aa8c7","level":"log","text":"log"}
```

In repl, we keep the pretty log style with the unicode prefix.
 
# Manually testing the PR

1. run the sandbox
2. run the node
3. deploy examples/logs.js 

4. curl http://localhost:8933/logs/SMART_FUNCTION/stream

Additionally, to test that the disconnected clients properly get removed, i printed out the clients address with the number of responsive senders. we can see that once the clients are disconnected, the entry clients in the broadcast map is removed properly after a few ping.

https://github.com/trilitech/jstz/assets/128799322/1dba2b68-96fe-45ec-b4fc-4ee0321ffea2


jstz_api/src/console.rs
```diff
@@ -22,10 +22,41 @@ use boa_engine::{
...
+use serde_json;
 use tezos_smart_rollup::prelude::debug_msg;
 
+#[derive(Serialize, Deserialize)]
+struct JSONLog {
+    function_address: String,
+    request_id: String,
+    level: String,
+    text: String,
+}
```
johnyob, line 38: Rust naming: `JSONLog` to `JsonLog`. 
Additionally I'd prefer to avoid having `JSON` in the name, since in future the encoding is likely to change. Perhaps `LogRecord`?

jstz_api/src/console.rs
```diff
@@ -135,15 +189,24 @@ fn formatter(data: &[JsValue], context: &mut Context<'_>) -> JsResult<String> {
     }
 }
 
-#[derive(Trace, Finalize)]
+#[derive(Finalize)]
 struct Console {
     groups: Vec<String>,
+    // TODO: Remove these once `Jstz` object is implemented
+    pub contract_address: PublicKeyHash,
+    pub operation_hash: Blake2b,
```
johnyob, line 197: I think both of these have dedicated types, namely `Address` and `OperationHash`

jstz_api/src/console.rs
```diff
@@ -234,6 +297,7 @@ impl Console {
         context: &mut Context<'_>,
     ) -> JsResult<()> {
         LogMessage::Warn(formatter(data, context)?).log(rt, self);
+        LogMessage::Warn(formatter(data, context)?).json_log(rt, self);
         Ok(())
     }
```
johnyob, line 302: Instead of doing both of logs, I think we should have a feature flag to determine whether we want 'pretty' logs or 'json' logs

jstz_cli/src/repl.rs
```diff
@@ -36,7 +36,13 @@ pub fn exec(self_address: Option<String>, cfg: &Config) -> Result<()> {
...
     let realm_clone = rt.realm().clone();
 
-    realm_clone.register_api(ConsoleApi, rt.context());
+    realm_clone.register_api(
+        ConsoleApi {
+            contract_address: address.clone(),
+            operation_hash: Default::default(),
+        },
+        rt.context(),
+    );
```
johnyob, line 45: I'd prefer to avoid the use of `Default::default()` here. I'm trying to think of a better approach to avoid dummy values in `jstz_api` for the CLI (since it is a code smell). 

We could try having `ConsoleApi` as an enum which is either `Cli` or `Proto { contract_address, operation_hash }`?

jstz_proto/src/executor/contract.rs
```diff
@@ -156,8 +174,9 @@ impl Script {
         &self,
         contract_address: Address,
         context: &mut Context<'_>,
+        hash: &OperationHash,
```
johnyob, line 177: Style: `Context<'_>` should _always_ be the last argument

jstz_proto/src/executor/contract.rs
```diff
@@ -248,12 +267,13 @@ impl Script {
         address: &Address,
         request: &JsValue,
         context: &mut Context<'_>,
+        operation_hash: &OperationHash,
```
johnyob, line 270: Ditto

jstz_node/src/services/logs/broadcaster.rs
```diff
@@ -50,42 +51,57 @@ impl Broadcaster {
...
+                    .send(sse::Event::Comment("ping".into()))
+                    .await
+                    .is_ok()
+                {
+                    // TODO: why cannot we avoid cloning? clients are already moved?
+                    responsive_clients
+                        .entry(address.clone())
+                        .or_insert(Vec::default())
+                        .push(sender);
+                }
```
johnyob, line 68: To understand the borrowing issue here, we can annotate the owners as follows:
```rust
let clients = self.clients.lock().clone(); // <-- owned HashMap<Address, Vec<..>>

// ...

for (address, senders) in clients { // client moved here into iterator, address & sender are owned
  for sender in senders { // senders moved here, sender is owned
    if sender.send(..).await.is_ok() { // sender is immutably borrowed here
      responsive_clients
        .entry(address) // address is moved here, BUT it can only be moved once and Rust cannot ensure this loop only iterates once!
        .or_default()
        .push(sender);
    }
  }
}
```

Thus we get a borrow checking error. To satisfy the borrow checker, we need to hoist the entry call:
```rust

        for (address, senders) in clients {
            let responsive_senders = responsive_clients.entry(address).or_default();

            for sender in senders {
                if sender
                    .send(sse::Event::Comment("ping".into()))
                    .await
                    .is_ok()
                {
                    responsive_senders.push(sender);
                }
            }
        }
```

jstz_node/src/services/logs/broadcaster.rs
```diff
@@ -50,42 +51,57 @@ impl Broadcaster {
...
+    ) -> Sse<InfallibleStream<ReceiverStream<sse::Event>>> {
         let (tx, rx) = mpsc::channel(10);
 
         tx.send(sse::Data::new("connected").into()).await.unwrap();
 
-        self.clients.lock().push(tx);
+        self.clients
+            .lock()
+            .entry(address)
+            .or_insert(Vec::default())
```
johnyob, line 87:
```suggestion
            .or_default()
```

jstz_node/src/services/logs/broadcaster.rs
ryutamago: thanks, that makes sense! We also need to make sure that the empty vector for stale clients will get wiped out so perhaps the following would work?
```
        for (address, senders) in clients {
            if senders.is_empty() {
                continue;
            }
            let responsive_senders = responsive_clients.entry(address).or_default();
            for sender in senders {
                if sender
                    .send(sse::Event::Comment("ping".into()))
                    .await
                    .is_ok()
                {
                    responsive_senders.push(sender);
                }
            }
        }
```

johnyob: Yes, additionally if no senders are responsive, then the entry shouldn't be inserted.

jstz_api/src/console.rs
```diff
@@ -22,10 +22,47 @@ use boa_engine::{
...
 use tezos_smart_rollup::prelude::debug_msg;
 
+pub const LOG_PREFIX: &str = "[JSTZ:SMART_FUNCTION:LOG] ";
+
+#[derive(Serialize, Deserialize)]
+pub struct LogRecord {
+    pub contract_address: PublicKeyHash,
+    request_id: String,
+    level: String,
+    text: String,
```
johnyob, line 39: Could they all be public?

jstz_proto/src/executor/contract.rs
```diff
@@ -133,8 +133,21 @@ impl Script {
...
+    ) {
         register_web_apis(self.realm(), context);
+        // TODO: Register console API in `register_web_apis` once `Jstz` object is implemented
+        self.realm().register_api(
+            jstz_api::ConsoleApi::Proto {
+                contract_address: contract_address.clone(),
+                operation_hash: operation_hash.clone(),
+            },
+            context,
+        );
```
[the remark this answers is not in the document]
johnyob: Update: This doesn't cause the console API to be registered twice. But it does permit the user to revert logging back to the Proto (JSON) style of logging instead of the Cli (Pretty) style of logging. This will be fixed in a future PR.

johnyob (approved):
LGTM 🎉

PR #199, opened 2023-11-23
Bugfix: Ensure sandbox config clears when closing sandbox

johnyob:
# Context

Currently the sandbox occasionally reports:
```
...
Saving sandbox config
^CReceived signal 2, shutting down...
Error: Command "/Users/ajob410/tezos/jstz/octez-client" "-base-dir" "/var/folders/5j/sd7sw3657ld425146jzz54xr0000gn/T/nix-shell.BPq0My/octez_clientLOHfDI" "-endpoint" "http://127.0.0.1:18730" "bake" "for" "--minimal-timestamp" failed:
```
This error results in the sandbox config not being cleared, meaning when running `sandbox start` following this, 
the CLI complains that the sandbox is already running. 

This heisenbug can be forced with the following diff:
```diff
diff --git a/jstz_cli/src/sandbox/daemon.rs b/jstz_cli/src/sandbox/daemon.rs
index ddabe5b..d91173f 100644
--- a/jstz_cli/src/sandbox/daemon.rs
+++ b/jstz_cli/src/sandbox/daemon.rs
@@ -249,12 +249,12 @@ impl OctezThread {
 
         let thread_handle: JoinHandle<Result<()>> = thread::spawn(move || {
             loop {
+                f(&cfg)?;
+
                 if shutdown_rx.try_recv().is_ok() {
                     break;
                 }
 
-                f(&cfg)?;
-
                 sleep(Duration::from_secs(1));
             }
 ```

# Description

This PR fixes the issue of the baker failing when shutting down due to the node shutting down.

Additionally this PR fixes some logging around `logs/client.log`, namely ensuring that the logs are retained every time the bake command is ran.

# Manually testing the PR

```sh
```
and then ctrl-c-ing it with the above diff applied.

### DIFF
--- a/jstz_cli/src/octez.rs
+++ b/jstz_cli/src/octez.rs
@@ -36,6 +36,8 @@ impl OctezClient {
             &format!("http://127.0.0.1:{}", cfg.octez_node_rpc_port),
         ]);
 
+        command.env("TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER", "Y");
+
         Ok(command)
     }
 
@@ -128,9 +130,7 @@ impl OctezClient {
         Ok(())
     }
 
-    pub fn bake(cfg: &Config, log_file: &PathBuf, options: &[&str]) -> Result<String> {
-        let log_file = File::create(log_file)?;
-
+    pub fn bake(cfg: &Config, log_file: &File, options: &[&str]) -> Result<String> {
         output(
             Self::command(cfg)?
                 .arg("bake")
--- a/jstz_cli/src/sandbox/daemon.rs
+++ b/jstz_cli/src/sandbox/daemon.rs
@@ -1,5 +1,5 @@
 use std::{
-    fs,
+    fs::{self, File},
     io::Write,
     process::{Child, Command},
     sync::mpsc::{self, Sender},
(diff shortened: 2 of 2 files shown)


jstz_cli/src/octez.rs
```diff
@@ -36,6 +36,8 @@ impl OctezClient {
             &format!("http://127.0.0.1:{}", cfg.octez_node_rpc_port),
         ]);
 
+        command.env("TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER", "Y");
```
alanmarkoTrilitech, line 39: Can you please write to the PR description or a comment what exactly does this do?

jstz_cli/src/sandbox/daemon.rs
```diff
@@ -119,12 +119,10 @@ fn init_client(cfg: &Config) -> Result<()> {
...
 }
 
-fn client_bake(cfg: &Config) -> Result<()> {
-    OctezClient::bake(
-        cfg,
-        &cfg.jstz_path.join("logs/client.log"),
-        &["for", "--minimal-timestamp"],
-    )?;
+fn client_bake(cfg: &Config, log_file: &File) -> Result<()> {
+    // SAFETY: When a baking fails, then we want to silently ignore the error and
```
alanmarkoTrilitech, line 123: Question: what are the reasons while baking can fail?

alanmarkoTrilitech (approved):
LGTM! Can you please just write some more details to the PR description about what caused the bug and how the PR fixes it?

jstz_cli/src/sandbox/daemon.rs
johnyob: The node is down, or inconsistent context, or many other (undocumented) reasons 😅 😩

jstz_cli/src/octez.rs
johnyob: `TEZOS_CLIENT_UNSAFE_DISABLE_DISCLAIMER` disables the a disclaimer that is logged on each octez-client command when using a 'unsafe network' such as a local sandbox (or a testnet)

PR #234, opened 2023-12-04
Rust 1.73.0 toolchain

vapourismo:
Depends: #238 

To work with the latest Rollup SDK's RISC-V target you need Rust 1.73.0. This upgrades the Rust toolchain.

There appear to be some formatting changes relating to let-else bindings.

While upgrading I took the liberty of upgrading to a `rust-toolchain.toml` file to consolidate toolchain-related information. This has an impact on CI given the `action-rs/toolchain` action does not support this and is no longer maintained. Fortunately, GitHub runner images come with rustup by default.


.prettierignore
```diff
@@ -9,3 +9,6 @@
 **/wpt
 **/manifest.json
 **/wptreport.json
+
+# Lock files
+**/*.lock
```
johnyob, line 14: I'm assuming this is to stop `flake.lock` from being formatted? Or were there other `.lock` files that prettier was formatting?

johnyob (commented):
Thank you for updating our `rust-toolchain` file -- I've been meaning to get around to that for ages 😅 

Looks like the `action-rs/toolchain` action became unmaintained 2 days after I wrote the CI pipeline 😩

.prettierignore
vapourismo: Yes, in this case it was the `flake.lock`.

vapourismo:
> Could you rebase on `main` (now that #238 is merged) to ensure the CI is green

Rebased and tested.

johnyob (approved):
LGTM 🎉

PR #235, opened 2023-12-04
Health: Move all crates to `crates/`

johnyob:
# Context

**Related Tasks**: [Use `/crates` for crates directory](https://app.asana.com/0/1205770721173531/1205870630092075/f)

To conform to existing standards in larger Rust projects, we will be moving all crates to `crates/`. 
There is already precedence for this style of folder separation in `jstz` e.g TypeScript packages in `packages/`.

Additionally this clears the root folder (which is slowly becoming more cluttered 😬 )

# Description

This PR moves all crates to `crates/`. Additionally to fit better with smart contract practices, all contracts (e.g. from `jstz_bridge`) have been moved to `contracts/`. 

All commit history is retained using `git mv` (although this requires a `--follow` flag when using `git log` manually). 

# Manually testing the PR

None required :)


.github/dependabot.yml
```diff
@@ -6,18 +6,6 @@ updates:
       interval: daily
     commit-message:
       prefix: "Deps (NPM)"
-  - package-ecosystem: npm
```
alanmarkoTrilitech, line 9: Why are these being removed? Shouldn't there be new settings for new directories?

alanmarkoTrilitech (approved):
LGTM, but please address the question regarding the dependabot changes before merging.

.github/dependabot.yml
johnyob: Dependabot should be able to traverse our file structure to find relevant manifest files for NPM (`package.json`), Cargo (`Cargo.toml`) and GitHub Actions (.yml file in `.github/workflows`)

johnyob: Actually I'm mistaken, only the nightly version of dependabot does this 😔 I will update the config to have all the relevant paths

johnyob: @alanmarkoTrilitech updated

alanmarkoTrilitech (approved):

PR #253, opened 2023-12-05
Fix Darwin CC flags dilemma when targeting WASM and other architectures

vapourismo:
Not all architectures in Clang support `-mcpu`. Therefore it is better to not use those machine-specific flags at all when cross-compiling to other architectures.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,6 +32,23 @@
               frameworks
             );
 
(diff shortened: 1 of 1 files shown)


johnyob:
LGTM, tested locally

johnyob (approved):

PR #254, opened 2023-12-05
Implemented `UnderlyingSource` and added a trivial `ReadableStream` constructor

xavierm02:
# Description

Implemented `UnderlyingSource`, and added a trivial `ReadableStream` constructor that only transforms its first input into an `UnderlyingSource`.

A few other things were added to make this possible / easier:
- A wrapper type `JsFn<T, I, O>` for `JsFunction` that allows to be more explicit about the JS function we expect to have, and to automatically handle the `TryFromJs` and `IntoJs` conversion for inputs and outputs.
- A `Todo` placeholder type for types from the spec that have not been defined yet.
- Type aliases that allow referring to some types by their WebIDL names.

# Manual testing

```sh
cargo run --bin jstz -- repl
>> new ReadableStream()
thread 'main' panicked at 'not yet implemented', jstz_api/src/stream/readable/mod.rs:39:9
```

```sh
cargo run --bin jstz -- repl
>> new ReadableStream({type:"bytes"})
thread 'main' panicked at 'not yet implemented', jstz_api/src/stream/readable/mod.rs:39:9
```

```sh
cargo run --bin jstz -- repl
>> new ReadableStream({type:"invalid"})
Uncaught TypeError: invalid is not a valid value for enumeration ReadableStreamType.
```

One can also add
```rust
underlying_source.cancel(Some(boa_engine::js_string!("reason").into()), context);
```
in the `constructor` of `ReadableStreamClass` and then test it with
```sh
cargo run --bin jstz -- repl
>> new ReadableStream({cancel:function(reason){console.log(reason);}})
[🪵] reason
thread 'main' panicked at 'not yet implemented', jstz_api/src/stream/readable/mod.rs:42:9
Uncaught TypeError: invalid is not a valid value for enumeration ReadableStreamType.
```

# Checklist

### DIFF
--- a/crates/jstz_api/src/stream/readable/underlying_source.rs
+++ b/crates/jstz_api/src/stream/readable/underlying_source.rs
@@ -1,0 +1,389 @@
+//! [Streams Standard - § 4.2.3. The underlying source API][https://streams.spec.whatwg.org/#underlying-source-api]
+
+use boa_engine::{
+    object::builtins::JsPromise, value::TryFromJs, Context, JsNativeError, JsObject,
+    JsResult, JsValue,
+};
+use boa_gc::{custom_trace, Finalize, Trace};
+use jstz_core::{
+    impl_into_js_from_into, js_fn::JsFn, native::JsNativeObject, value::IntoJs,
+};
+use std::str::FromStr;
+
+use crate::idl;
+
+use crate::stream::tmp::*;
+
+/// [Streams Standard - § 4.2.3.][https://streams.spec.whatwg.org/#underlying-source-api]
+/// > ```
+/// > dictionary UnderlyingSource {
+/// >   UnderlyingSourceStartCallback start;
+/// >   UnderlyingSourcePullCallback pull;
+/// >   UnderlyingSourceCancelCallback cancel;
+/// >   ReadableStreamType type;
+/// >   \[EnforceRange\] unsigned long long autoAllocateChunkSize;
+/// > };
+/// > ```
+#[derive(Debug)]
+pub struct UnderlyingSource {
+    /// A reference to the [`JsObject`] from which the [`UnderlyingSource`] was build, used as `this` parameter when calling the methods of the [`UnderlyingSource`].
+    ///
+    /// [Streams Standard - § 4.2.4.][https://streams.spec.whatwg.org/#rs-prototype]
+    /// > Note: We cannot declare the underlyingSource argument as having the UnderlyingSource type directly, because doing so would lose the reference to the original object. We need to retain the object so we can invoke the various methods on it.
+    pub this: JsObject,
+
+    /// [Streams Standard - § 4.2.3.][https://streams.spec.whatwg.org/#dom-underlyingsource-start]
+    /// > **`start(controller)`, of type UnderlyingSourceStartCallback**
+    /// >
+    /// > A function that is called immediately during creation of the ReadableStream.
+    /// >
+    /// >  Typically this is used to adapt a push source by setting up relevant event listeners, as in the example of § 10.1 A readable stream with an underlying push source (no backpressure support), or to acquire access to a pull source, as in § 10.4 A readable stream with an underlying pull source.
+    /// >
+    /// >  If this setup process is asynchronous, it can return a promise to signal success or failure; a rejected promise will error the stream. Any thrown exceptions will be re-thrown by the ReadableStream() constructor.
+    pub start: Option<UnderlyingSourceStartCallback>,
+
+    /// [Streams Standard - § 4.2.3.][https://streams.spec.whatwg.org/#dom-underlyingsource-pull]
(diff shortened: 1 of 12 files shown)


Lesenr1 (commented):
👍

Lesenr1 (approved):
LGTM

crates/jstz_core/src/js_fn.rs
```diff
@@ -0,0 +1,136 @@
...
+    _this_type: PhantomData<T>,
+    _inputs_type: PhantomData<I>,
+    _output_type: PhantomData<O>,
+}
+
+impl<T: IntoJs, const N: usize, I: IntoJsArgs<N>, O: TryFromJs> Finalize
+    for JsFn<T, N, I, O>
+{
+    fn finalize(&self) {}
+}
```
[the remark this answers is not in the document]
xavierm02: Or maybe the impl is empty because `fn finalize(&self) {}` is the default, in which case, we could indeed derive it if we don't care about the superfluous `+ Finalize` requirements for the parameters.

crates/jstz_api/src/stream/readable/underlying_source.rs
```diff
@@ -0,0 +1,408 @@
...
+    });
+}
+
+// TODO derive this implementation with a macro?
+impl TryFromJs for UnderlyingSource {
+    fn try_from_js(value: &JsValue, context: &mut Context<'_>) -> JsResult<Self> {
+        // TODO check that this function works as intended in all cases,
+        // and move it either to a new derive macro for TryFromJs, or to JsObject
+        #[allow(non_snake_case)]
+        pub fn get_JsObject_property(
```
johnyob, line 111: Why non-snake-case? I'm not too sure why the deviation from standard style is helpful here

crates/jstz_api/src/stream/readable/underlying_source.rs
```diff
@@ -0,0 +1,408 @@
...
+            obj: &JsObject,
+            name: &str,
+            context: &mut Context<'_>,
+        ) -> JsResult<JsValue> {
+            let key = PropertyKey::from(js_string!(name));
+            let key2 = key.clone();
+            let has_prop = obj.has_property(key, context)?;
+            if has_prop {
+                obj.get(key2, context)
+            } else {
```
johnyob, line 121:
```suggestion
            let key = PropertyKey::from(js_string!(name));
            let has_prop = obj.has_property(key.clone(), context)?;
            if has_prop {
                obj.get(key, context)
            } else {
```

crates/jstz_core/src/js_fn.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+/// A `JsFn<T, N, I, O>` is a `JsFunction` tagged with some Rust types used to handle the `TryFromJs` and `IntoJs` conversions automatically:
+/// - `T` is the type of the `this` parameter;
+/// - `N` is the arity;
+/// - `I` is a tuple `(I1, ..., IN)` that contains the types of the parameters;
+/// - `O` is the type of the output.
+#[derive(Debug)]
+pub struct JsFn<T: IntoJs, const N: usize, I: IntoJsArgs<N>, O: TryFromJs> {
```
johnyob, line 48: On reflection, I think the const-level generic `N` adds more effort than it is worth. 

Removing it would give us `JsFn<T, I, O>` which:
- I think is clearer
- It is also the type I'd expect once associated const-types are permitted in Rust (https://github.com/rust-lang/rust/issues/60551), resulting a more localized refactor. 

The downside of such a change is that we'd have return a `Vec` for `IntoJsArgs`. But I think is an acceptable given that it will eventually be changed once associated const-types are in Rust.

johnyob: Also another suggestion would be to have `T` default to `JsValue`: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#default-generic-type-parameters-and-operator-overloading

crates/jstz_core/src/js_fn.rs
johnyob, line 59:
```suggestion
impl<T: IntoJs, const N: usize, I: IntoJsArgs<N>, O: TryFromJs> Finalize
    for JsFn<T, N, I, O>
{
}
```

crates/jstz_api/src/stream/readable/underlying_source.rs
xavierm02: What would you prefer? `get_jsobject_property`?

crates/jstz_core/src/js_fn.rs
johnyob: Hmmm, yeah I guess that is a bit odd. Perhaps we can just wrap over this type with a macro that will substitute `JsValue` for `T` if 2 arguments are given

PR #255, opened 2023-12-05
Cli: implement `jstz logs trace <sf>`

ryutamago:
# Context
This PR Implements `jstz logs trace <sf>`, allowing users to listen to logs from a given sf on cli
Additionally, i've added an extra field in the cli config that specifies the host of jstz node. The default should be localhost (e.g.
```
  {
   ...
  "jstz_node_host": "127.0.0.1",
  "jstz_node_port": 8933,
 ```

[Task link](https://app.asana.com/0/1205770721173533/1205770721347936)

# Description

# Manually testing the PR

Deploy a function,
run `jstz logs trace <sf>` to connect to the jstz node and run the function

https://github.com/trilitech/jstz/assets/128799322/9d162569-d85f-4d50-bda0-aff45ec46101

### DIFF
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -7,6 +7,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Result};
+use jstz_proto::context::account::Address;
 use serde::{Deserialize, Serialize};
 
 use crate::account::account::{Account, AliasAccount};
@@ -97,6 +98,18 @@ impl AccountConfig {
 
     pub fn list_all(&self) -> Vec<(&String, &Account)> {
         self.accounts.iter().collect()
+    }
+
+    pub fn get_address(&self, address_or_alias: &str) -> Result<Address> {
+        if let Ok(address) = Address::from_base58(address_or_alias) {
+            return Ok(address);
+        }
+
+        if let Ok(account) = self.get(address_or_alias) {
+            return Ok(account.address().clone());
+        }
+
+        Err(anyhow!("Invalid alias or address"))
     }
 }
 
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -110,6 +110,8 @@ pub struct Config {
     pub octez_node_port: u16,
     /// The port of the octez RPC node
     pub octez_node_rpc_port: u16,
+    /// The host of the jstz node
+    pub jstz_node_host: String,
     /// The port of the jstz node
     pub jstz_node_port: u16,
     /// Sandbox config (None if sandbox is not running)
@@ -155,6 +157,7 @@ impl Config {
             octez_path: PathBuf::from_str(".").unwrap(),
             octez_node_port: 18731,
             octez_node_rpc_port: 18730,
+            jstz_node_host: "127.0.0.1".to_string(),
             jstz_node_port: 8933,
             sandbox: None,
             accounts: AccountConfig::default(),
--- a/crates/jstz_cli/src/logs/mod.rs
+++ b/crates/jstz_cli/src/logs/mod.rs
@@ -1,5 +1,6 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::Subcommand;
+use jstz_crypto::public_key_hash::PublicKeyHash;
 
 use crate::config::Config;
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: crates/jstz_cli/src/logs/mod.rs [hunk 1]
```diff
@@ -9,36 +10,17 @@ mod trace;
...
-        /// View console.error messages
-        #[arg(long)]
-        error: bool,
-        /// View contract creations
-        #[arg(long)]
-        contract: bool,
-        /// Add custom search strings
-        #[arg(long, value_name = "custom_strings")]
-        custom: Vec<String>,
+        #[arg(value_name = "SMART_FUNCTIOIN")]
```
alanmarkoTrilitech, line 13: SMART_FUNCTION

I think it would be best to use ADDRESS or SMART_FUNCTION_ADDRESS for consistency with other naming - e.g. self address

crates/jstz_cli/src/logs/trace.rs
```diff
@@ -1,71 +1,33 @@
...
+use anyhow::Result;
+use futures_util::stream::StreamExt;
+use jstz_api::LogRecord;
+use jstz_proto::context::account::Address;
+use reqwest_eventsource::{Event, EventSource};
+
+pub async fn exec(address: Address, cfg: &Config) -> Result<()> {
+    let url = format!(
+        "http://{}:{}/logs/{}/stream",
+        cfg.jstz_node_host,
```
alanmarkoTrilitech, line 11: Can you please change it also in other parts of the code to use this config value? e.g. jstz_cli/src/jstz.rs

crates/jstz_cli/src/logs/mod.rs
```diff
@@ -9,36 +10,17 @@ mod trace;
...
-        Command::Trace {
-            log,
-            info,
-            warn,
-            error,
-            contract,
-            custom,
-        } => trace::exec(log, info, warn, error, contract, custom, cfg),
+        Command::Trace { address } => {
+            let address = PublicKeyHash::from_base58(&address)
```
alanmarkoTrilitech, line 21: Can you please make it so that it works with aliases, so that the address value can either be an alias or an address? 
Using AccountConfig (cfg.accounts.get), you can check whether an account exists for a particular address. If it does, use the account, otherwise try to convert it into an address.

alanmarkoTrilitech (changes requested):

crates/jstz_cli/src/logs/mod.rs
```diff
@@ -1,5 +1,7 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use clap::Subcommand;
+use jstz_crypto::public_key_hash::PublicKeyHash;
+use jstz_proto::context::account::Address;
```
johnyob, line 4: Use of both `PublicKeyHash` and `Address` is a code smell (since `Address` should be an alias of `PublicKeyHash`)

crates/jstz_cli/src/logs/mod.rs
```diff
@@ -9,36 +11,29 @@ mod trace;
...
+    if let Ok(address) = PublicKeyHash::from_base58(smart_function) {
+        return Ok(address);
+    }
+
+    if let Ok(account) = cfg.accounts.get(smart_function) {
+        return Ok(account.address().clone());
+    }
+
+    Err(anyhow!("Invalid smart function alias or address"))
+}
```
johnyob, line 39: @alanmarkoTrilitech I think we have logic for this elsewhere in the CLI. Could @ryutamago use it?

crates/jstz_cli/src/logs/trace.rs
```diff
@@ -1,71 +1,33 @@
...
+            Ok(Event::Message(message)) => {
+                if let Ok(log_record) = serde_json::from_str::<LogRecord>(&message.data) {
+                    println!("{}", serde_json::to_string_pretty(&log_record).unwrap());
+                }
+            }
+            Err(err) => {
+                println!("Event source error: {}", err);
+                event_source.close();
+            }
+        }
```
johnyob, line 30: Would be nice to have pretty logs here, instead of JSON

crates/jstz_cli/src/logs/mod.rs
alanmarkoTrilitech: I don't think there is exactly this variant where you get address from either account or alias - there are methods to get either an account from alias or use the self-address one. I think @ryutamago can move this method to AccountConfig in jstz_cli/config.rs and make it general so that it does not apply just to smart function addresses (it can be named e.g. just get_address)

crates/jstz_cli/src/logs/mod.rs
```diff
@@ -9,36 +11,29 @@ mod trace;
...
-            log,
-            info,
-            warn,
-            error,
-            contract,
-            custom,
-        } => trace::exec(log, info, warn, error, contract, custom, cfg),
+        Command::Trace { smart_function } => {
+            let address = get_address_for_smart_function(&smart_function, cfg)?;
+            trace::exec(address, cfg).await
```
alanmarkoTrilitech, line 24: Nit: to match the style of other commands, it would be better to have just trace::exec(address, cfg).await here that does 'let address  = ...' parsing inside of it, similarly to main.rs.

crates/jstz_cli/src/logs/trace.rs
ryutamago: As discussed, I will work on this in the following PR :)

crates/jstz_cli/src/config.rs
```diff
@@ -98,6 +99,18 @@ impl AccountConfig {
...
+    pub fn get_address(&self, address_or_alias: &str) -> Result<Address> {
+        if let Ok(address) = Address::from_base58(address_or_alias) {
+            return Ok(address);
+        }
+
+        if let Ok(account) = self.get(address_or_alias) {
+            return Ok(account.address().clone());
+        }
+
+        Err(anyhow!("Invalid smart function alias or address"))
```
alanmarkoTrilitech, line 112: Since this is defined in AccountConfig, it would be better if the error messages don't mention smart functions, so that it can be more easily reused elsewhere.

```suggestion
        Err(anyhow!("Invalid alias or address"))
```

alanmarkoTrilitech (approved):
LGTM! I've just added a suggestion regarding a small nit in get_address method.

crates/jstz_cli/src/config.rs
ryutamago: nice catch!

PR #257, opened 2023-12-06
fix(cli): fix paths in sandbox

johnyob:
# Context

Currently running `sandbox start` gets stuck on:
```
Configuring sandbox... done
Initializing octez-node configuration... done
Generating identity... done
Starting node... done
```

This is because the paths for the sandbox parameter files changed.

# Description

This PR simply fixes the paths to the parameter files (and any other broken paths)

# Manually testing the PR

```sh
```


johnyob:
> Also I've noticed that there is a warning generated in jstz_core, which should not happen in main(it can be fixed in another PR):

I've also noticed this. I'll fix in another PR

PR #267, opened 2023-12-11
Cli & Api & Proto: Introduce `ConsoleLog` trait

ryutamago:
# Context
Currently the console logging methods aren't flexible. We also had an issue where the loggings method gets reverted (e.g. calling a contract in repl would change the logging method from `pretty` to `json`. 

**Related Tasks**: [Task](https://app.asana.com/0/1205770721173533/1206083271582090) 

# Description
This PR
* introduces the `JsLog` trait that provides a way to control how logging is done in the js console object. There are two implementations: `JsonLogger` for proto and `PrettyLog` for repl.
* `set_js_logger` can be called to set the type of logger before running the script 
* Introduced a new global variable `TraceData` in HostDefined that contains both the`contract_address` and `operation_hash`. This way we don't have to store those fields in the ConsoleApi or ContractApi
* Pretty print the log in logs trace <sf>

# Manually testing the PR
Proto: 
* start the sandbox and jstz_node
* deploy a contract
* run `jstz logs trace <sf>` to listen to the logs
* run the deployed contract to see if the logs are properly working

https://github.com/trilitech/jstz/assets/128799322/0fbd007f-50e4-4f02-bbde-99514956c265

Repl: 
* console.log / error / info  in repl


crates/jstz_api/src/console.rs
```diff
@@ -23,97 +21,66 @@ use boa_engine::{
...
-impl FromStr for LogRecord {
-    type Err = serde_json::Error;
-    fn from_str(json: &str) -> Result<Self, Self::Err> {
-        serde_json::from_str(json)
+use jstz_core::value::IntoJs;
+use log::js_log;
+pub use log::{set_console_logger, ConsoleLog, LogData, LogLevel};
+
+mod log {
+    use boa_engine::{Context, JsNativeError, JsResult};
```
johnyob, line 29: Suggest to use `use super::...` here

crates/jstz_api/src/console.rs
```diff
@@ -23,97 +21,66 @@ use boa_engine::{
...
-    }
-}
-
-impl FromStr for LogRecord {
-    type Err = serde_json::Error;
-    fn from_str(json: &str) -> Result<Self, Self::Err> {
-        serde_json::from_str(json)
+use jstz_core::value::IntoJs;
+use log::js_log;
+pub use log::{set_console_logger, ConsoleLog, LogData, LogLevel};
```
johnyob, line 26: Personal preference, but I usually prefer to read the module linearly, meaning that these statements would make more sense _after_ `mod log`

crates/jstz_api/src/console.rs
```diff
@@ -23,97 +21,66 @@ use boa_engine::{
...
 
-    fn message(&self) -> &str {
-        match self {
-            LogMessage::Error(msg) => msg,
-            LogMessage::Warn(msg) => msg,
-            LogMessage::Info(msg) => msg,
-            LogMessage::Log(msg) => msg,
-        }
+    // The implementor of this trait controls how console.log/warn/error etc is handled.
+    pub trait ConsoleLog {
```
johnyob, line 59: Perhaps rename to `JsLog`, this trait isn't just limited to being used for the `console` API

crates/jstz_api/src/console.rs
```diff
@@ -23,97 +21,66 @@ use boa_engine::{
...
+            match self {
+                LogLevel::ERROR => '🔴',
+                LogLevel::WARN => '🟠',
+                LogLevel::INFO => '🟢',
+                LogLevel::LOG => '🪵',
             }
         }
     }
+    #[derive(Serialize, Deserialize)]
+    pub struct LogData {
```
johnyob, line 52: I would've suggested `LogRecord`, but I assume this was chosen to avoid the name collision with `LogRecord` in `jstz_proto`

crates/jstz_proto/src/executor/contract/console_logger.rs
johnyob, line 1: I would've probably put this in `/api`, although coming to think of it, it doesn't really belong there 🤔

crates/jstz_api/src/console.rs
ryutamago: yes, i'm open to renaming it if you have better suggestion...

crates/jstz_proto/src/executor/contract/console_logger.rs
ryutamago: yes i guess the logging is still part of the `proto` so perhaps we could put them all in `jstz_proto/src/logger` ? I could add the `request_logger` for persistent logging there too. wdyt ? 🤔

crates/jstz_proto/src/executor/contract.rs
```diff
@@ -131,23 +134,14 @@ impl Script {
         Ok(Self(module))
     }
 
-    // TODO: we need to be able to specify the type of console API (Proto vs Cli),
-    // With current implementation, calling a contract in CLI will revert the logging back to Proto
     fn register_apis(
         &self,
         contract_address: Address,
-        context: &mut Context<'_>,
```
sam-finch-tezos, line 139: What's the purpose of this change?

sam-finch-tezos (approved):
lgtm

crates/jstz_proto/src/executor/contract.rs
ryutamago: i'm sticking to the conventions to place context at the end, similar to how we place runtime at the start :)

PR #268, opened 2023-12-12
Api: Support `Random` API

sam-finch-tezos:
# Description

This PR enables the random api for jstz. 
We monkeypatch Boa's Math object with our own implementation of random().
Note this is not currently spec compliant. We use the contract address and the operation hash as a seed, so  calling the same contract programmatically twice in a row will generate the same sequence of random numbers. 
The simplest fix would be to wait for risc-v and add the tick count into the seed

**Related issue**: [Enable random api](https://app.asana.com/0/1205770721173531/1205770721333453/f)

# Manual testing

```sh
# with sandbox running
jstz deploy examples/pi.js
# copy the new address
jstz run "tezos://`pbpaste`"
# pi = 3.141 // random approximation s.d. 0.05.  
```

# Checklist

- [x] Internal documentation has been added (if appropriate)


johnyob (approved):
LGTM 🥳 

If we wanted to avoid this monkey patching in the future, I believe `getrandom` permits a way to set a global random generator

PR #269, opened 2023-12-12
Health: Add `clippy` linting

ryutamago:
# Context
This PR introduces a new command `make lint` that uses clippy and apply them during pre-push / pipeline.

Additionally, i refactored the code so that there will be no linting warnings. The first commit is auto generated by cargo clippy and i manually fixed the linting issues in the last commit.

Make sure you run `scripts/install-hooks.sh` to install the pre-push-hook script. 

**Asana Task**: [Add linting to ci/pre-push hook](https://app.asana.com/0/1205770721173531/1206077181649731/f)

# Manually testing the PR

* pipeline: we can see [here](https://github.com/trilitech/jstz/actions/runs/7185008193/job/19567349006?pr=269) that the lint job throws error
* pre-push-hook: run `scripts/install-hooks.sh` to install the pre-push-hook script and try pushing with warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,4 +72,4 @@ fmt-check: fmt-nix-check fmt-rust-check fmt-js-check
 
 .PHONY: lint
 lint:
-	@cargo clippy -- -D warnings -A clippy::let_underscore_future -A clippy::module_inception -A clippy::op_ref -A clippy::manual_strip -A clippy::missing_safety_doc -A clippy::slow_vector_initialization -A clippy::empty_loop
+	@cargo clippy -- -D warnings -A clippy::let_underscore_future -A clippy::module_inception -A clippy::op_ref -A clippy::manual_strip -A clippy::missing_safety_doc -A clippy::slow_vector_initialization -A clippy::empty_loop -A clippy::expect-fun-call
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,7 @@ test:
 	@cargo test
 
 .PHONY: check
-check:
-	@cargo fmt --check
-	@cargo clippy --all-targets -- --deny warnings
+check: lint fmt
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -72,4 +72,4 @@ fmt-check: fmt-nix-check fmt-rust-check fmt-js-check
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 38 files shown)


Makefile
```diff
@@ -71,3 +71,7 @@ fmt: fmt-nix fmt-rust fmt-js
 
 .PHONY: fmt-check
 fmt-check: fmt-nix-check fmt-rust-check fmt-js-check
+
+.PHONY: lint
+lint:
+	@cargo clippy -- -D warnings -A clippy::let_underscore_future -A clippy::module_inception -A clippy::op_ref -A clippy::manual_strip -A clippy::missing_safety_doc -A clippy::slow_vector_initialization -A clippy::empty_loop
```
johnyob, line 77: I don't think is the preferred way of configuring clippy

https://doc.rust-lang.org/stable/clippy/configuration.html

johnyob:
It would also be good to split the existing `make check` target into calling `fmt` and `lint`

crates/jstz_api/tests/wpt.rs
```diff
@@ -214,15 +214,15 @@ impl jstz_core::Api for TestHarnessReportApi {
             let value = context
                 .global_object()
                 .get(js_string!(name), context)
-                .expect(&format!("globalThis.{} is undefined", name));
+                .unwrap_or_else(|_| panic!("globalThis.{} is undefined", name));
```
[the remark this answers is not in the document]
ryutamago: i can silence this ling rule :)

Makefile
ryutamago: as discussed, clippy config file doesn't support all the linting rules sadly :(

crates/jstz_core/src/realm.rs
```diff
@@ -303,10 +301,9 @@ impl HostDefined {
                 host_defined,
                 Attribute::all(),
             )
-            .expect(&format!(
-                "{:?} object should only be defined once",
-                Self::NAME
-            ))
+            .unwrap_or_else(|_| {
+                panic!("{:?} object should only be defined once", Self::NAME)
+            })
```
johnyob, line 306: Revert change:
```suggestion
            .expect(&format!(
                "{:?} object should only be defined once",
                Self::NAME
            ))
```

johnyob (approved):
LGTM

PR #272, opened 2023-12-14
Health: Fix panics (low hanging fruit)

sam-finch-tezos:
# Context
This PR replaces several uses of `.expect()` and `.unwrap()` which could cause the rollup to panic. 

**Related Tasks**: [Analyze all possible panics](https://app.asana.com/0/1205770721173531/1206072089048483/f)

# Description
Calls to `.expect` and `.unwrap` are replaced wherever possible by converting to a result.  This reduces instances where the rollup may panic.

* Focus is on the areas where the rollup kernel would panic, `jstz_cli` and `jstz_node` may still panic.
* Panics caused by a programmer error in rust (such as interactions with `HostDefined` are not addressed. 
* There is no claim that this is exhaustive. Only the easy fixes have been addressed.

### DIFF
--- a/crates/jstz_api/src/console.rs
+++ b/crates/jstz_api/src/console.rs
@@ -41,15 +41,13 @@ pub struct LogRecord {
 
 impl Display for LogRecord {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(
-            &serde_json::to_string(self).expect("Failed to convert LogRecord to string"),
-        )
+        f.write_str(&serde_json::to_string(self).map_err(|_| std::fmt::Error)?)
     }
 }
 
 impl LogRecord {
-    pub fn from_string(json: &str) -> Self {
-        serde_json::from_str(json).expect("Failed to deserialize JSONLog")
+    pub fn try_from_string(json: &str) -> Option<Self> {
+        serde_json::from_str(json).ok()
     }
 }
 
--- a/crates/jstz_api/tests/wpt.rs
+++ b/crates/jstz_api/tests/wpt.rs
@@ -214,15 +214,15 @@ impl jstz_core::Api for TestHarnessReportApi {
             let value = context
                 .global_object()
                 .get(js_string!(name), context)
-                .expect(&format!("globalThis.{} is undefined", name));
+                .unwrap_or_else(|_| panic!("globalThis.{} is undefined", name));
 
             let function = value
                 .as_callable()
-                .expect(&format!("globalThis.{} is not callable", name));
+                .unwrap_or_else(|| panic!("globalThis.{} is not callable", name));
 
             function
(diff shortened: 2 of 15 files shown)


crates/jstz_api/tests/wpt.rs
```diff
@@ -214,15 +214,15 @@ impl jstz_core::Api for TestHarnessReportApi {
             let value = context
                 .global_object()
                 .get(js_string!(name), context)
-                .expect(&format!("globalThis.{} is undefined", name));
+                .unwrap_or_else(|_| panic!("globalThis.{} is undefined", name));
```
johnyob, line 217: Is this the linter? @ryutamago

crates/jstz_node/src/services/logs/mod.rs
```diff
@@ -60,7 +60,8 @@ impl LogsService {
                 line = lines.next_line() => {
                     if let Ok(Some(msg)) = line {
                         if msg.starts_with(LOG_PREFIX) {
-                            let log = LogRecord::from_string(&msg[LOG_PREFIX.len()..]);
+                            let log = LogRecord::try_from_string(&msg[LOG_PREFIX.len()..])
+                                .unwrap();
```
johnyob, line 64: Would prefer a `.expect` here

johnyob (approved):
LGTM 

I think there is some discussion to be had around having `JsError`s for internal invariants being broken / when to prefer `.expect` over `Result` and vice-versa

crates/jstz_api/src/console.rs
```diff
@@ -41,15 +41,13 @@ pub struct LogRecord {
...
     }
 }
 
 impl LogRecord {
-    pub fn from_string(json: &str) -> Self {
-        serde_json::from_str(json).expect("Failed to deserialize JSONLog")
+    pub fn try_from_string(json: &str) -> Option<Self> {
+        serde_json::from_str(json).ok()
     }
 }
```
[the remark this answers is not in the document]
sam-finch-tezos: Yes, yes it could!

crates/jstz_api/tests/wpt.rs
sam-finch-tezos: I think it's `clippy fix`. I'll change it back.

sam-finch-tezos:
> LGTM
> 
> I think there is some discussion to be had around having `JsError`s for internal invariants being broken / when to prefer `.expect` over `Result` and vice-versa

Definitely. I think the basic points are
* Broken internal invariants should be `.expect` and should theoretically never occur.
* If it's possible to cause it from javascript (even if you have to try really hard) it should always be a `JsError`
* If in doubt prefer `JsError` because if youu're wrong it should never occur.. 

For this one I've taken the view that if I can't convince myself in about 10 minutes that it will never be thrown then it's a JsError.  I have no doubt I got the call wrong on a couple of occasions.

PR #277, opened 2023-12-15
Health: Catch and log panics

sam-finch-tezos:
# Context

This is a draft PR demonstrating how panics can be caught and logged at the entry point.


johnyob (commented):
Oh, this is v nice!

Need to test locally, but looks v good :)

sam-finch-tezos:
> Oh, this is v nice!
> 
> Need to test locally, but looks v good :)

It's definitely a draft at this stage. I'm reasonably satisfied it's safe, but it's a Friday afternoon. Plus I think we should do something better with the panics than a little explosion emoji.

johnyob:
Closing as discussed with @sam-finch-tezos in sprint planning meeting

PR #323, opened 2024-01-19
CLI: Instant highlighting

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1205771213627876

# Description
This PR adds instant highlighting in REPL. This instead of highlighting Javascript after hitting enter, highlights the code as it is written.

# Manually testing the PR
Writing something in repl (cargo run --bin jstz -- repl).

### DIFF
--- a/crates/jstz_cli/src/repl.rs
+++ b/crates/jstz_cli/src/repl.rs
@@ -114,8 +114,7 @@ pub fn exec(self_address: Option<String>, cfg: &Config) -> Result<()> {
     }
     set_js_logger(&PrettyLogger);
 
-    let mut rl =
-        Editor::<JsHighlighter, _>::new().expect("Failed to create a new editor.");
+    let mut rl = Editor::<JsHighlighter, _>::new()?;
     rl.set_helper(Some(JsHighlighter::new()));
 
     let mut mock_hrt = MockHost::default();
--- a/crates/jstz_cli/src/repl.rs
+++ b/crates/jstz_cli/src/repl.rs
@@ -19,6 +19,7 @@ use rustyline::{
 use std::borrow::Cow;
 use tezos_smart_rollup_mock::MockHost;
 
+use std::fmt::Write;
 use syntect::{
     easy::HighlightLines,
     highlighting::{Style, Theme, ThemeSet},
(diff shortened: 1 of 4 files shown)


crates/jstz_cli/src/repl.rs
```diff
@@ -5,41 +5,98 @@ use jstz_api::{
...
+
+        // Find JavaScript syntax and a theme
+        let syntax = ss.find_syntax_by_extension("js").unwrap();
+        let theme = &ts.themes["base16-ocean.dark"];
+
+        JsHighlighter {
+            ss: ss.clone(),
+            syntax: syntax.clone(),
+            theme: theme.clone(),
+        }
```
johnyob, line 50: Possibly unnecessary `.clone`s

crates/jstz_cli/src/repl.rs
```diff
@@ -57,7 +114,9 @@ pub fn exec(self_address: Option<String>, cfg: &Config) -> Result<()> {
     }
     set_js_logger(&PrettyLogger);
 
-    let mut rl = Editor::<(), _>::new().expect("Failed to create a new editor.");
+    let mut rl =
```
johnyob, line 117: Perhaps propagate this error instead of using `.expect`

johnyob (commented):
lgtm, need to test locally before approval

crates/jstz_cli/src/repl.rs
alanmarkoTrilitech: It seems to actually require clone, as ss/syntax/them should not be references.

johnyob: Indeed, having looked at the `syntect` docs, I don't think the clones for `syntax` and `theme` can be removed. However `ss.clone()` can be removed.

```suggestion
        JsHighlighter {
            syntax: syntax.clone(),
            theme: theme.clone(),
            ss,
        }
```

crates/jstz_cli/src/repl.rs
```diff
@@ -5,41 +5,97 @@ use jstz_api::{
...
+            let color = style.foreground;
+            let color_string = format!("\x1b[38;2;{};{};{}m", color.r, color.g, color.b);
 
+            result.push_str(&color_string);
+            result.push_str(text);
+
+            result.push_str("\x1b[0m");
+        }
+
+        result
```
johnyob, line 66: Slightly more Rusty version:

```rust
        styles
            .iter()
            .map(|&(style, text)| {
                let color = style.foreground;
                format!(
                    "\x1b[38;2;{};{};{}m{}\x1b[0m",
                    color.r, color.g, color.b, text
                )
            })
            .collect()
 ```
 
 (GitHub won't let me propose this as a suggestion due to the deleted lines in the diff)

johnyob (approved):
LGTM

Only some minor suggestions left

crates/jstz_cli/src/repl.rs
alanmarkoTrilitech: This exact suggestion doesn't pass lint. It needs fold instead of map.

PR #320, opened 2024-01-17
Make jstz kernel compile to RISC-V target

victor-dumitrescu:
# Context

This PR makes the jstz kernel compile to the `rust-std-riscv64gc-unknown-hermit` target, as part of the development of the RISC-V runtime project in Octez. The current integration can be found [here](https://gitlab.com/tezos/tezos/-/tree/master/src/risc_v/jstz).

# Description

The main change in this PR is replacing the use of BLS12-381 signatures with Ed25519. This is necessary for the RISC-V project because the `blst` library cannot currently be compiled to the Hermit target.

# Manually testing the PR

PR #333, opened 2024-01-24
Node: Proxy external inbox messages via `jstz-node`

johnyob:
# Context

**Related Tasks**: [Proxy external messages from `jstz-node` to the rollup node](https://app.asana.com/0/1205770721173531/1206249753769937/f) 

This PR introduces a proxy for external inbox messages to the jstz rollup using the `jstz-node`. This avoids
the CLI directly communicating with the `octez-node`. 

The only direct`octez-node` dependencies are in the sandbox (not removable for now) and L1 account deposits. 

# Description

This PR adds a `POST` endpoint to the `/operations` service, permitting the injection of operations via the `jstz-node` instead of using the `octez-node`. 

**Note**: There seems to be an issue with the JSON serialization and deserialization of BLS signatures with `actix-web`, this could be fixed by #320. For now we just directly pass bytes, but in future (when implementing IDL definitions for `jstz-node`) we will provide the correct de+serializers. 

# Manually testing the PR

Startup the sandbox and deploy / run any smart function :)

### DIFF
--- a/crates/jstz_cli/src/jstz.rs
+++ b/crates/jstz_cli/src/jstz.rs
@@ -3,6 +3,7 @@ use std::time::Duration;
 use anyhow::anyhow;
 use anyhow::Result;
 use jstz_api::KvValue;
+use jstz_proto::operation::SignedOperation;
 use jstz_proto::{context::account::Nonce, operation::OperationHash, receipt::Receipt};
 use reqwest::StatusCode;
 use tokio::time::sleep;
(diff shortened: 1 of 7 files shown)


crates/octez/src/rollup.rs
```diff
@@ -103,6 +103,31 @@ impl OctezRollupClient {
...
+                &external_messages
+                    .into_iter()
+                    .map(hex::encode)
+                    .collect::<Vec<String>>(),
+            )
+            .send()
+            .await?;
+
+        if res.status() == 200 {
+            // TODO: Should we ignore the response?
```
alanmarkoTrilitech, line 124: I don't think we should ignore it - the client should know if the operation wasn't successful.

crates/jstz_cli/src/jstz.rs
```diff
@@ -140,7 +141,22 @@ impl JstzClient {
             }
 
             // tokio sleep
-            sleep(Duration::from_millis(100)).await;
+            sleep(Duration::from_millis(200)).await;
```
alanmarkoTrilitech, line 144: What motivated the change to 200?

alanmarkoTrilitech (approved):
LGTM, I just left some minor comments.

crates/octez/src/rollup.rs
johnyob: Yes, this is captured by the `else` part of this `if` block. The comment here refers to whether we should be ignoring the response from the rollup (which IIRC is a list of operation hashes)

crates/jstz_cli/src/jstz.rs
johnyob: I noticed that at 100ms we were spamming the jstz-node, happy to change it back

PR #337, opened 2024-01-24
Cli: Remove `jstz_path` from CLI config

johnyob:
# Context

**Related Tasks**: [Removing dependence on jstz source location in CLI](https://app.asana.com/0/1205770721173531/1206345416248634/f)

# Description

This PR introduces a Rust build script that removes the need for `jstz_path` in the CLI.

This introduces an inadvertent semantic changes -- the `/logs` directory used by jstz CLI will be created in the current working directory:
- if you run `cargo run --bin jstz` in the root directory for the jstz repository, this should have no effect, but 
- if you run `cargo run --bin jstz` in a subdirectory `A`, this will create `A/logs` directory

In future, the `logs` directory will be in `/tmp` once we start the `jstz-node` automatically when running `jstz sandbox start`. 

# Manually testing the PR

Running the sandbox and checking various features still work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,26 @@
 .PHONY: all
-all: build test check
+all: test build check
+
+.PHONY: build
+build: build-cli-kernel
+	@cargo build --release
 
 .PHONY: build-bridge
 build-bridge:
 	@ligo compile contract contracts/jstz_bridge.mligo \
 		--module "Jstz_bridge" > contracts/jstz_bridge.tz
 
-.PHONY: build
-build:
+.PHONY: build-kernel
+build-kernel:
 	@cargo build --package jstz_kernel --target wasm32-unknown-unknown --release
+
+.PHONY: build-cli-kernel
+build-cli-kernel: build-kernel
+	@cp target/wasm32-unknown-unknown/release/jstz_kernel.wasm crates/jstz_cli/jstz_kernel.wasm
+
+.PHONY: build-cli
+build-cli: build-cli-kernel
+	@cargo build --package jstz_cli --release
 
 .PHONY: build-deps
 build-deps:
@@ -28,8 +40,8 @@ check: lint fmt
 .PHONY: clean
 clean:
 	@cargo clean
-	rm -f result
-	rm -rf logs
+	@rm -f result
+	@rm -rf logs
 
 .PHONY: fmt-nix-check
 fmt-nix-check:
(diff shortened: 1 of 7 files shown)


crates/jstz_cli/src/repl/debug_api/account.rs
```diff
@@ -0,0 +1,133 @@
+use std::ops::Deref;
```
alanmarkoTrilitech, line 1: I guess this was just moved from previous debug_api in the jstz_cli - not sure why Github shows it as new files. 

Not sure if I am very supportive of this movement - I think it would be better to rather have all APIs in a single place - e.g. jstz_api.  In this way it is quite hidden and I think especially if someone new joins jstz development in the future it may be easier to intuitively understand the code if all APIs can be seen together. We may also decide at some point to enable debugging with DebugAPI in the actual .js contracts.

alanmarkoTrilitech: Ah I see now why it is not showing as movement - there are actually two debug_api folders now. Please remove one of them.

crates/jstz_cli/src/repl/js_logger.rs
```diff
@@ -0,0 +1,24 @@
+use boa_engine::Context;
+use jstz_api::js_log::{JsLog, LogData};
+use jstz_core::runtime::with_global_host;
+use tezos_smart_rollup::prelude::debug_msg;
+
+pub struct PrettyLogger;
+
```
alanmarkoTrilitech, line 7: Please mention addition of PrettyLogger and its functionality in the PR description.

crates/jstz_cli/build.rs
```diff
@@ -0,0 +1,62 @@
+use std::{
```
alanmarkoTrilitech, line 1: Can you please provide some more description of what is the expected behaviour of running this?

alanmarkoTrilitech (changes requested):
A more detailed description for the PR would help. It also seems to be doing more than just removing jstz_path - e.g. introduction of PrettyLogger and movement of debug_api.

johnyob:
> A more detailed description for the PR would help. It also seems to be doing more than just removing jstz_path - e.g. introduction of PrettyLogger and movement of debug_api.

You're right, I will split this PR into two PRs

Makefile
```diff
@@ -61,6 +69,7 @@ fmt: fmt-nix fmt-rust fmt-js
 .PHONY: fmt-check
 fmt-check: fmt-nix-check fmt-rust-check fmt-js-check
 
+# Clippy builds the CLI, in order to build the CLI we need to build the kernel
 .PHONY: lint
-lint:
+lint: build-kernel
```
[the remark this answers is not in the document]
johnyob: Clippy must expand all macros _before_ linting. Since `jstz_kernel.wasm` is a build-time dependency (explicitly required for expanding macros), then we must build the kernel before linting the CLI

crates/jstz_cli/src/repl/js_logger.rs
johnyob: `PrettyLogger` is already part of [`main`](https://github.com/trilitech/jstz/blob/main/crates/jstz_cli/src/js_logger.rs)

crates/jstz_cli/src/repl/debug_api/account.rs
johnyob: Likely an artefact from moving the folder. Will likely disappear once the PR is split

johnyob: Resolving since diff nolonger exists

PR #347, opened 2024-01-26
feat(node): node runner

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1206345416213081

# Description

This PR makes jstz-node run automatically after sandbox start is executed.

# Manually testing the PR


crates/jstz_node/src/node_runner.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Endpoint defailts for the `jstz-node`
+const ENDPOINT: (&str, u16) = ("127.0.0.1", 8933);
+
+pub async fn run_node(
+    rollup_node_rpc_addr: String,
+    rollup_node_rpc_port: u16,
+    rollup_endpoint: Option<String>,
+    kernel_file_path: String,
+) -> io::Result<()> {
+    env_logger::init_from_env(Env::default().default_filter_or("info"));
```
johnyob, line 19: We may want to disable the logger in the sandbox, wdyt?

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -245,8 +245,15 @@ pub fn main(cfg: &mut Config) -> Result<()> {
     println!("Saving sandbox config");
     cfg.save()?;
 
+    // Run node
+    let jstz_node = OctezThread::from_child(
+        Command::new("cargo")
+            .args(["run", "--bin", "jstz-node"])
+            .spawn()?,
+    );
+
```
johnyob, line 254: We shouldn't be using `cargo` like this, since this won't be packaged correctly on the user's installation (only works in our local dev environment)

crates/jstz_node/src/main.rs
```diff
@@ -32,44 +27,18 @@ struct Args {
     #[arg(short, long)]
     rollup_endpoint: Option<String>,
 
-    #[arg(long)]
+    #[arg(long, default_value = DEFAULT_KERNEL_FILE_PATH)]
     kernel_file_path: String,
 }
 
 #[actix_web::main]
```
johnyob, line 34: To avoid issues in future, we should explicitly switch to `#[tokio::main]` (which `actix` uses internally)

crates/jstz_node/src/main.rs
```diff
@@ -1,26 +1,21 @@
...
+    },
+    node_runner::run_node,
+};
 pub use error::{Error, Result};
 
+mod config;
 mod error;
+mod node_runner;
 mod services;
 mod tailed_file;
```
johnyob, line 17: I think these can be dropped and you simply import them from the crate defined in `jstz_node` (introducing `lib.rs` now defines a crate + a binary in `jstz_node`)

crates/jstz_cli/src/sandbox/daemon.rs
alanmarkoTrilitech: yes, we need to compile it and then change to just 'jstz-node'. For now however, I think it can stay like this and we can create a separate task for changing it.

johnyob: I'd be keen on tracking it in this PR since the other changes in this PR would be mostly unnecessary if this was our desired approach

crates/jstz_node/src/node_runner.rs
alanmarkoTrilitech: I've added an option to disable it.

What is the actual purpose of it?

crates/jstz_node/src/main.rs
alanmarkoTrilitech: we need both crate and binary since we can still run just jstz-node, or run it from the sandbox

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -254,6 +257,26 @@ pub fn main(cfg: &mut Config) -> Result<()> {
...
+                println!("Jstz node started 🎉");
+                let _ = run_node(
+                    DEFAULT_ROLLUP_NODE_RPC_ADDR.to_string(),
+                    DEFAULT_ROLLUP_RPC_PORT,
+                    None,
+                    DEFAULT_KERNEL_FILE_PATH.to_string(),
+                    false,
+                )
+                .await;
+            })
```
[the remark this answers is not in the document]
alanmarkoTrilitech: This is part of another PR.

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -254,6 +257,26 @@ pub fn main(cfg: &mut Config) -> Result<()> {
...
+            .await
+            .expect("Local task panicked");
+        })
+        .await;
```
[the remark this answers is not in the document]
alanmarkoTrilitech: Jstz node apparently cannot be moved between threads because it involves some non-thread safe behaviour, so  LocalSet is essentially creating a context for it so that it can be run.

johnyob:
Added a review commit to avoid back-and-forth (feel free to revert if you disagree with the changes):

TL;DR:
- Improved crate/bin separation between `jstz-node` and `jstz_node`
- Re-introduced a cli flag (seems to have been dropped somewhere) for defining the jstz node endpoint of the `jstz-node` 
- Factored out running jstz-node logic into `run_jstz_node` in the sandbox

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -17,9 +13,14 @@ use std::{
     time::Duration,
 };
 use tempfile::TempDir;
+use tokio::task;
 
 use crate::config::{Config, SandboxConfig, SANDBOX_OCTEZ_SMART_ROLLUP_PORT};
 
+const SANDBOX_JSTZ_NODE_ADDR: &str = "127.0.0.1";
+const SANDBOX_JSTZ_NODE_PORT: u16 = 8933;
+const SANDBOX_OCTEZ_SMART_ROLLUP_ADDR: &str = "127.0.0.1";
```
alanmarkoTrilitech, line 22: Now we have it defined it two places which is not ideal. Ideally there should be a shared library for these constants.

crates/jstz_node/Cargo.toml
```diff
@@ -24,7 +24,7 @@ parking_lot = "0.12.1"
 reqwest = { version = "0.11.22", features = ["json"] }
 serde = { version = "1.0.183", features = ["derive"] }
 thiserror = "1.0.56"
-tokio = { version = "1.33.0", features = ["fs"] }
+tokio = { version = "1.33.0", features = ["fs", "rt-multi-thread"] }
```
alanmarkoTrilitech, line 27: what does this do?

alanmarkoTrilitech (commented):
The changes look good to me, except those constants.

You will have to approve it though.

crates/jstz_cli/src/sandbox/daemon.rs
johnyob: They don't necessarily have to be the same -- the node has defaults for it's CLI and we have defaults for the sandbox. 

They're only currently the same because the node was designed with the sandbox in mind

crates/jstz_node/Cargo.toml
johnyob: Tokio seemed to give me a compiler error without this feature flag

johnyob (approved):
Please squash before merging (especially removing the 'review' commit)

PR #349, opened 2024-01-29
Health: Move repl-related files to a sub directory in `jstz_cli`

johnyob:
# Context

# Description

This PR collects all REPL relates files in `jstz_cli` into a submodule, this includes:
- `repl.rs`
- The REPL debug APIs (these APIs will never / extremely unlikely to be used anywhere else)
- The `js_logger` -- the console logger

# Manually testing the PR


alanmarkoTrilitech (approved):
I'm approving it, however I still think it is worth considering having all APIs (in this case debug API) separated together with other APIs. We don't know if we wouldn't want to use this or similar API at some point also for debugging .js files with smart functions. It would also make it easier to intuitively understand the code structure, as REPL is more of a terminal/"frontend" part and debug API is an API.

johnyob:
> I'm approving it, however I still think it is worth considering having all APIs (in this case debug API) separated together with other APIs. We don't know if we wouldn't want to use this or similar API at some point also for debugging .js files with smart functions. It would also make it easier to intuitively understand the code structure, as REPL is more of a terminal/"frontend" part and debug API is an API.

Thank you 🙏 Currently I prefer separating the APIs across crates based on their responsibilities:
- `jstz_api` for common Web Standard APIs
- `jstz_proto` for jstz (remote) specific APIs 
- `jstz_cli/repl` for jstz REPL (local) specific APIs

This improves coupling, reusability, testing (when we eventually implement those 😶), etc. For me the most notable benefit is that we have a compile-time guarantee that none of the REPL APIs are compiled with the `jstz_kernel`. 

However, separating by responsibility has the disadvantage that it decreases code locality. I'm open to other approaches to achieving the compile-time guarantee (such as using feature flags). 

Additionally I see the REPL more than just a 'terminal' component, since it must also define the context for the JavaScript runtime + all the APIs that are loaded.

PR #350, opened 2024-01-30
docs: update jstz installation doc

Frank-en-stein:
Updated jstz docs/Installation.md with the latest installation steps.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -10,8 +10,7 @@ Currently, `jstz` can only be installed by building from sources using [Rust](ht
 
 ## Building from Source
 
-Below are instruction on how to build `jstz` from source. Additionally, this section
-details how to install Octez, which is used for our local sandbox.
+Below are instruction on how to build `jstz` from source. Additionally, this section details how to install all the prerequisites needed to build `jstz` from sources.
 
 ### Cloning the Repository
 
@@ -25,6 +24,36 @@ git clone https://github.com/trilitech/jstz.git
 Both `jstz` and Octez are packaged with Nix, a package manager and system configuration tool that makes building from sources easy! See the [Nix docs](https://nixos.org/download.html) for instructions for your system. Additionally, ensure [Nix flakes are enabled](https://nixos.wiki/wiki/Flakes#Enable_flakes).
 :::
 
+#### LLVM 🛠️
+
+::: code-group
+
+```sh [MacOS]
+brew install llvm
+export CC="$(brew --prefix llvm)/bin/clang"
+```
+
+```sh [Ubuntu]
+sudo apt-get install clang-11
+export CC=clang-11
+```
+
+```sh [Fedora]
+sudo dnf install clang
+export CC=clang
+```
+
+```sh [Arch Linux]
+pacman -S clang
+export CC=clang
+```
+
+```sh [Nix]
+nix-env -iA nixpkgs.llvm
+```
+
+:::
+
 #### Rust 🦀
 
 > `jstz` requires a specific release of Rust. The version of Rust required is specified in the `rust-toolchain` file.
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -13,68 +13,82 @@ Currently, `jstz` can only be installed by building from sources using [Rust](ht
...
-:::
+### llvm 📜
 
-#### Rust 🦀
+Install llvm. For instance, on MacOS, run:
 
-> `jstz` requires a specific release of Rust. The version of Rust required is specified in the `rust-toolchain` file.
+```sh
+brew install llvm
+```
```
johnyob, line 33: I'd opt for avoiding OS specific installations (either we provide both Linux & MacOS instructions) or we link to a page about installing them

docs/installation.md
```diff
@@ -13,68 +13,82 @@ Currently, `jstz` can only be installed by building from sources using [Rust](ht
...
 ```sh
 git clone https://github.com/trilitech/jstz.git
+git clone https://gitlab.com/tezos/tezos.git
 ```
 
 ### Prerequisites 📋
 
-::: tip  
-Both `jstz` and Octez are packaged with Nix, a package manager and system configuration tool that makes building from sources easy! See the [Nix docs](https://nixos.org/download.html) for instructions for your system. Additionally, ensure [Nix flakes are enabled](https://nixos.wiki/wiki/Flakes#Enable_flakes).
-:::
```
johnyob, line 26: Nix isn't necessarily a pre-requisite, but an optional one that makes things slightly easier (especially for those on MacOS). I'd prefer to keep it in the 'tips'

docs/installation.md
```diff
@@ -13,68 +13,82 @@ Currently, `jstz` can only be installed by building from sources using [Rust](ht
...
 
-### Cloning the Repository
+### Cloning the Repositories
+
+Get the source code for both `jstz` and Octez.
 
 ```sh
 git clone https://github.com/trilitech/jstz.git
+git clone https://gitlab.com/tezos/tezos.git
 ```
```
johnyob, line 23: You won't always have to build Octez from sources. 

Historically we did was due to floating point support and we only currently require sources since the `master` version of Octez is buggy

docs/installation.md
```diff
@@ -13,68 +13,82 @@ Currently, `jstz` can only be installed by building from sources using [Rust](ht
...
 
-After Nix successfully builds Octez (it takes a long time 🕣), the Octez binaries will be accessable from `result`.
-:::
+After Nix successfully builds Octez (which takes a long time 🕣), the Octez binaries will be available in the `result/bin/` directory. Copy the following binaries to the `jstz` root direactory:
+
+- octez-client
+- octez-node
+- octez-smart-rollup-node
+- octez-smart-rollup-wasm-debugger
```
johnyob, line 79: `octez-smart-rollup-wasm-debugger` should be dropped. Not required for the sandbox and is no-longer a useful tool for us

docs/installation.md
```diff
@@ -88,8 +102,7 @@ You can locate the resulting build artifact at `/target/wasm32-unknown-unknown/r
 You can now start the sandbox with:
 
 ```sh
-make build-installer
-cargo run -- sandbox start
+cargo run --bin jstz -- sandbox start
```
johnyob, line 105: Some of the build commands have changed since this was written 😅

I think it should be:
```sh
make build-cli-kernel
```

docs/installation.md
Frank-en-stein: That makes sense but should we keep it like this until we're out of the woods? Leaving building from source out of the docs as tribal knowledge would be harder for anyone trying to onboard to jstz.

johnyob: I'm not suggesting we leave it out, but cloning from sources for `tezos/tezos` is a step in installing Octez in the Octez documentation. 

We should avoid duplication between these two docs (only offering suggestions e.g. nix that might aid the user)

johnyob (commented):
See review commit. 

TL;DR:
- Keeping Nix optional 
- Adding a code-group for LLVM installation instructions
- Updating commit shas / branch names to tezos/tezos

docs/installation.md
```diff
@@ -26,13 +24,39 @@ git clone https://github.com/trilitech/jstz.git
 Both `jstz` and Octez are packaged with Nix, a package manager and system configuration tool that makes building from sources easy! See the [Nix docs](https://nixos.org/download.html) for instructions for your system. Additionally, ensure [Nix flakes are enabled](https://nixos.wiki/wiki/Flakes#Enable_flakes).
 :::
 
-### Clang and llvm 📜
+#### LLVM 🛠️
+
```
Frank-en-stein, line 28: ìs it better to have a copy of the installation guide? I was looking for modularised section but I suppose can't do it in md

johnyob (approved):

PR #352, opened 2024-01-30
Make jstz kernel compile to RISC-V target (electric boogaloo)

johnyob:
> Superseded PR from #320 

# Context

This PR makes the jstz kernel compile to the `rust-std-riscv64gc-unknown-hermit` target, as part of the development of the RISC-V runtime project in Octez. The current integration can be found [here](https://gitlab.com/tezos/tezos/-/tree/master/src/risc_v/jstz).

# Description

The main change in this PR is replacing the use of BLS12-381 signatures with Ed25519. This is necessary for the RISC-V project because the `blst` library cannot currently be compiled to the Hermit target.

# Manually testing the PR


alanmarkoTrilitech (approved):

alanmarkoTrilitech (commented):
Actually, needs to pass the last remaining check

alanmarkoTrilitech (changes requested):
Needs to pass the last check.

alanmarkoTrilitech (approved):

PR #360, opened 2024-02-05
feat(cli): separate errors into 'user errors' and 'internal errors'

johnyob:
# Context

Improves CLI error reporting

# Description

This PR:
1. Introduces an `error.rs` file that defines the CLI's result + error type (anyhow w/ internal constructors such as `UserError`).
2. Updates the rest of the CLI to use this error module.
3. Improves various error messages in the CLI

# Manually testing the PR

Running a selection of commands with erroneous outputs

### DIFF
--- a/crates/jstz_cli/src/logs/trace.rs
+++ b/crates/jstz_cli/src/logs/trace.rs
@@ -1,9 +1,9 @@
-use crate::Config;
-use anyhow::Result;
 use futures_util::stream::StreamExt;
 use jstz_api::js_log::LogLevel;
 use jstz_proto::js_logger::LogRecord;
 use reqwest_eventsource::{Event, EventSource};
+
+use crate::{error::Result, Config};
 
 const DEFAULT_LOG_LOG_LEVEL: LogLevel = LogLevel::LOG;
 
@@ -24,7 +24,7 @@ pub async fn exec(
 
     while let Some(event) = event_source.next().await {
         match event {
-            Ok(Event::Open) => println!("Connection open with {}", url),
+            Ok(Event::Open) => println!("Event source opened."),
             Ok(Event::Message(message)) => {
                 if let Ok(log_record) = serde_json::from_str::<LogRecord>(&message.data) {
                     let LogRecord { level, text, .. } = log_record;
@@ -34,8 +34,8 @@ pub async fn exec(
                 }
             }
             Err(err) => {
-                println!("Event source error: {}", err);
                 event_source.close();
+                eprintln!("Event source closed with an error: {}", err);
             }
         }
     }
--- a/crates/jstz_cli/src/sandbox/daemon.rs
+++ b/crates/jstz_cli/src/sandbox/daemon.rs
@@ -7,7 +7,6 @@ use std::{
     time::Duration,
 };
 
-use anyhow::Result;
 use jstz_rollup::{
     deploy_ctez_contract, rollup::make_installer, BootstrapAccount, BridgeContract,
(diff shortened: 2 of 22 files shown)


crates/jstz_cli/src/utils.rs
```diff
@@ -9,8 +9,8 @@ pub fn from_file_or_id(data_or_file: String) -> String {
 }
 
 fn get_stdin() -> String {
-    let lines: Result<Vec<_>, _> = io::stdin().lines().collect();
-    lines.expect("Can't read from stdin").join("\n")
+    let lines: io::Result<Vec<_>> = io::stdin().lines().collect();
+    lines.expect("Failed to read lines from stdin").join("\n")
```
sam-finch-tezos, line 13: Should we definitely be using expect here? It doesn't really matter in the CLI, but an error might be better.

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -228,7 +230,7 @@ fn start_sandbox(cfg: &Config) -> Result<(OctezThread, OctezThread, OctezThread)
 pub fn main(cfg: &mut Config) -> Result<()> {
     // 1. Check if sandbox is already running
     if cfg.sandbox.is_some() {
-        return Err(anyhow::anyhow!("Sandbox is already running!"));
+        bail_user_error!("Sandbox is already running!");
```
sam-finch-tezos, line 233:
```suggestion
        bail_user_error!("The sandbox is already running!");
```

crates/jstz_cli/src/logs/trace.rs
```diff
@@ -34,8 +34,8 @@ pub async fn exec(
                 }
             }
             Err(err) => {
-                println!("Event source error: {}", err);
                 event_source.close();
+                eprintln!("Event source error, with an error: {}", err);
```
sam-finch-tezos, line 38: Is this really better than the original?

sam-finch-tezos (approved):
lgtm; a couple of minor sugestions.
I think long term It would be better to have interactive recoveries eg "you are already logged in as {current_alias}, logout {current_alias} and log in as {new_alias}? Y/n".  But lets keep this for now.

johnyob:
> I think long term It would be better to have interactive recoveries eg "you are already logged in as {current_alias}, logout {current_alias} and log in as {new_alias}? Y/n". But lets keep this for now.

Dw this is coming

crates/jstz_cli/src/utils.rs
johnyob: We can probably use a `Result` here. Will fix 👍

crates/jstz_cli/src/logs/trace.rs
johnyob: It should've been `Event source closed with an error: {}`. Will fix

crates/jstz_cli/src/sandbox/daemon.rs
johnyob: Fixed 👍 (also changed the `sandbox stop` error to match)

crates/jstz_cli/src/logs/trace.rs
johnyob: Fixed

crates/jstz_cli/src/utils.rs
johnyob: Fixed, it's slightly more messy. But it works :)

sam-finch-tezos (approved):

PR #361, opened 2024-02-05
refactor(cli): misc config improvements

johnyob:
# Context

# Description

**Dependencies**: #360 

This PR introduces various improves to the handling of accounts:
- Renaming `Owned` and `Alias` accounts to `User` and `Deployment` accounts
- Introducing a `AddressOrAlias` type for the CLI arguments
- Removing some old CLI flags (self-address, referer) + renaming current flags to more sensible / consistent names
- `Config::load()` will now always be at the start of the entry-points of commands, instead of `main`
- Use `register_*_apis` functions from `jstz_proto` with default / config values

# Manually testing the PR


sam-finch-tezos (approved):
lgtm

crates/jstz_cli/src/config.rs
```diff
@@ -1,118 +1,172 @@
...
+    if let Ok(value) = env::var("JSTZ_HOME") {
+        PathBuf::from(value)
+    } else {
+        dirs::home_dir()
+            .expect("Could not find home directory")
+            .join(".jstz")
+    }
+}
+
+// Represents a collection of accounts: users or 'deployments' (aka smart functions)
```
alanmarkoTrilitech, line 29: How about User&Function?

crates/jstz_cli/src/config.rs
```diff
@@ -1,118 +1,172 @@
...
+            if !self.contains(alias.as_str()) {
+                bail!(
+                    "Cannot set current account to '{}', account not found.",
+                    alias
+                );
+            }
+
+            if let Some(Account::Deployment(_)) = self.accounts.get(alias) {
+                bail!(
+                    "Cannot set current account to '{}', it is a deployment.",
```
alanmarkoTrilitech, line 79:
```suggestion
                    "Cannot set current account to '{}', account belongs to a smart function.",
```

crates/jstz_cli/src/config.rs
```diff
@@ -1,118 +1,172 @@
...
-            return Ok(address);
+impl AddressOrAlias {
+    pub fn resolve(&self, cfg: &Config) -> Result<Address> {
+        match self {
+            AddressOrAlias::Address(address) => Ok(address.clone()),
+            AddressOrAlias::Alias(alias) => {
+                let account = cfg
+                    .accounts
+                    .get(alias)
+                    .ok_or_else(|| user_error!("User/deployment '{}' not found. Please provide a valid address or alias.", alias))?;
```
alanmarkoTrilitech, line 137:
```suggestion
                    .ok_or_else(|| user_error!("User/Function account '{}' not found. Please provide a valid address or alias.", alias))?;
```

crates/jstz_cli/src/config.rs
```diff
@@ -136,51 +189,31 @@ pub struct SandboxConfig {
...
-            octez_node_rpc_port: 18730,
-            jstz_node_port: 8933,
-            private: (),
-        }
-    }
-}
+pub const SANDBOX_OCTEZ_NODE_PORT: u16 = 18731;
+pub const SANDBOX_OCTEZ_NODE_RPC_PORT: u16 = 18730;
+pub const SANDBOX_JSTZ_NODE_PORT: u16 = 8933;
+pub const SANDBOX_OCTEZ_SMART_ROLLUP_PORT: u16 = 8932;
```
alanmarkoTrilitech, line 199: yeah, I guess it makes sense to have these hard-coded and not allowing the user to change the values.

crates/jstz_cli/src/config.rs
```diff
@@ -136,51 +189,31 @@ pub struct SandboxConfig {
...
+    pub fn load() -> Result<Self> {
         let path = Self::path();
 
         let config = if path.exists() {
             let json = fs::read_to_string(&path)?;
-            serde_json::from_str(&json)
-                .map_err(|e| Error::new(ErrorKind::InvalidData, e))?
+
+            serde_json::from_str(&json).map_err(|_| {
+                user_error!("Your configuration file is improperly configured.")
```
alanmarkoTrilitech, line 215: Would be good to write some instructions about how to configure it.

crates/jstz_cli/src/config.rs
```diff
@@ -189,25 +222,25 @@ impl Config {
...
 
     /// Save the configuration to the file
-    pub fn save(&self) -> std::io::Result<()> {
+    pub fn save(&self) -> Result<()> {
         let path = Self::path();
-        if let Some(parent) = path.parent() {
-            fs::create_dir_all(parent)?;
+
+        if !path.exists() {
+            if let Some(parent) = path.parent() {
```
alanmarkoTrilitech, line 229: what is the purpose of this? it seems to me like this should essentialy happen only if the user has deleted the config while running. it would be nice if it showed some error/notification in this case.

crates/jstz_cli/src/main.rs
```diff
@@ -85,38 +83,37 @@ enum Command {
...
-        Command::Logout {} => account::logout(cfg),
-        Command::WhoAmI {} => account::whoami(cfg),
-        Command::Kv(kv_command) => kv::exec(kv_command, cfg).await,
+        } => run::exec(url, http_method, gas_limit, json_data).await,
+        Command::Repl { account } => repl::exec(account),
+        Command::Logs(logs) => logs::exec(logs).await,
+        Command::Login { alias } => account::login(alias),
+        Command::Logout {} => account::logout(),
+        Command::WhoAmI {} => account::whoami(),
+        Command::Kv(kv_command) => kv::exec(kv_command).await,
```
alanmarkoTrilitech, line 107: at some point I want to also reorder and categorize these so that more commonly used commands like deploy/run are on the top.

crates/jstz_cli/src/repl/mod.rs
```diff
@@ -98,55 +98,45 @@ impl Completer for JsHighlighter {
...
+const DEFAULT_RANDOM_SEED: u64 = 42;
+
+pub fn exec(account: Option<AddressOrAlias>) -> Result<()> {
+    let cfg = Config::load()?;
+
+    let address = match account {
+        Some(account) => account.resolve(&cfg)?.clone(),
+        None => DEFAULT_SMART_FUNCTION_ADDRESS
+            .parse()
+            .expect("`DEFAULT_SMART_FUNCTION_ADDRESS` is a valid address."),
```
alanmarkoTrilitech, line 112:
```suggestion
            .expect("`DEFAULT_SMART_FUNCTION_ADDRESS` is an invalid address."),
```

crates/jstz_cli/src/term/emoji.rs
```diff
@@ -0,0 +1,4 @@
+use console::Emoji;
+
+pub static WARN: Emoji = Emoji("▲", "");
```
alanmarkoTrilitech, line 3: isn't there a better WARN emoji?

crates/jstz_cli/src/term/emoji.rs
```diff
@@ -0,0 +1,4 @@
...
+pub static ERROR: Emoji = Emoji("✕", "");
```
alanmarkoTrilitech, line 4: same for errot

alanmarkoTrilitech (changes requested):
This is a very large PR that should ideally be split into 3 smaller ones. In the interest of time I am not going to ask for it, but please do so next time for other similar PRs. 

I see that it has no manual testing section, so before merging please ensure that it was tested. Overall the changes seem reasonable, but I can't test them all so I believe we should have 1-2 day long internal ecosystem dogfooding session to properly test everything before wider dogfooding release as a lot of changes are being introduced quickly.

johnyob:
> This is a very large PR that should ideally be split into 3 smaller ones. In the interest of time I am not going to ask for it, but please do so next time for other similar PRs.

I concur, writing bits of the configure is a bit messy. Thanks for not requiring it (since this would delay merging this PR) 🙏

crates/jstz_cli/src/account.rs
```diff
@@ -0,0 +1,290 @@
+use bip39::{Language, Mnemonic, MnemonicType};
```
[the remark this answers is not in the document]
johnyob: The folder would simply be reduced to `account/mod.rs`. At this point it is better practice to simply have `account.rs`.

crates/jstz_cli/src/account.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        Ok(Self {
+            address,
+            secret_key: sk,
+            public_key: pk,
+        })
+    }
+}
+
+fn add_deployment(alias: String, address: Address) -> Result<()> {
```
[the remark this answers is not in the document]
johnyob: I think `User` is an apt description for key owners -- you and I can have multiple AWS/GCP accounts, but each account is still a `User` account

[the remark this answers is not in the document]
johnyob: `Deployment` I concede is slightly more confusing -- but also fits the current purpose of aliases, namely to alias deployed smart functions (deployments)

crates/jstz_cli/src/config.rs
johnyob: Function could work

crates/jstz_cli/src/config.rs
johnyob: Currently we don't have a docs link for config. Once we do, we can link it in this user error.

crates/jstz_cli/src/config.rs
johnyob: No, perhaps the config never existed in the first place (which is common on startup)

crates/jstz_cli/src/main.rs
johnyob: 👍 sgtm

crates/jstz_cli/src/term/emoji.rs
johnyob: Yes, but the benefit of this emoji is that it works well with ANSI colours

crates/jstz_cli/src/term/emoji.rs
johnyob: Ditto for the cross, works well with ANSI colours

johnyob:
> I see that it has no manual testing section, so before merging please ensure that it was tested. Overall the changes seem reasonable, but I can't test them all so I believe we should have 1-2 day long internal ecosystem dogfooding session to properly test everything before wider dogfooding release as a lot of changes are being introduced quickly.

We can add this to the sprint planning today, I think it is an excellent idea to do a day or 2 of QA

crates/jstz_cli/src/config.rs
johnyob: I went with `SmartFunction`, a bit more specific for our use case

alanmarkoTrilitech (approved):
LGTM! Please test it a bit before merging as this is a large PR and doesn't have manual testing section.

PR #366, opened 2024-02-06
feat(cli): daemonizing sandbox 😈

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1206345416213096

# Description
This PR daemonizes the sandbox 😈. Additionally it adds a new progress bar and redirects logging of both node and sandbox to file. 
There seems to be a bug with stopping - this will be solved in further PR.

# Manually testing the PR
```
cargo build --release
jstz sandbox start
jstz deploy examples/counter.js
```


crates/jstz_cli/src/sandbox/mod.rs
```diff
@@ -10,20 +10,19 @@ mod daemon;
 use crate::config::Config;
 
 #[derive(Subcommand)]
-pub enum Command {
+pub enum SandboxCommand {
```
johnyob, line 13: Any reason for the rename?

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -284,3 +324,66 @@ pub async fn main(cfg: &mut Config) -> Result<()> {
...
+        let stdout_file = OpenOptions::new()
+            .create(true)
+            .write(true)
+            .truncate(true)
+            .open(path.clone())?;
+
+        let mut child = Command::new(std::env::current_exe()?)
+            .args(["sandbox", "start", "--no-daemon"])
+            .stdout(Stdio::from(stdout_file))
+            .spawn()?;
```
johnyob, line 342: While this is essentially what forking a process is, it doesn't adhere to the standard way of creating a daemon. 

Did you look into using `daemonize`?

johnyob:
The progress bars are a nice touch. I'd really like to avoid using `Command` for anything but wrapping over external binaries (like `octez`) -- there is definitely a better approach by using unix libraries to create the daemon

crates/jstz_cli/src/sandbox/mod.rs
alanmarkoTrilitech: it was conflicting with process::Command, but I've later decided to move it to daemon.rs, so I guess it could be changed back to Command.

crates/jstz_cli/src/sandbox/daemon.rs
alanmarkoTrilitech: yes, I've spend significantly more time on trying to make it work through daemonize compared to this version. However, I've encountered multiple issues with daemonize - for example it isn't very compatible with LocalSet and how we run jstz-node. It also had some networking and configuration issues.

My guess is that it probably possible to make it somehow work through daemonize - however it would take more time to get it work and doesn't really functionally change it compared to this version, therefore I would suggest we keep it as is now, focus on more high priority features before the doogfooding release and then we can look at it again later if you really want.

johnyob: Just spent 10 mins playing around with this -- it isn't an issue with `LocalSet` or `jstz-node`, but the tokio runtime. 

Namely daemonize uses fork under the hood, but the tokio runtime breaks after forking (in the child process). To support this we'd need to carefully re-write bits of the sandbox to initialize the runtime _after_ forking.

crates/jstz_cli/src/sandbox/mod.rs
```diff
@@ -10,20 +10,19 @@ mod daemon;
...
     /// Starts a sandbox.
-    Start,
+    Start {
+        /// Do not daemonize the process.
+        #[clap(long, short, default_value = "false")]
+        no_daemon: bool,
+    },
```
johnyob, line 19: The standard flag for this is `--detach` (or `-d` for short). By default it is `false`, when set it detaches (daemonizes) the process

johnyob:
Also is it possible to fix the sandbox stopping bug in this PR -- without the ability to easily cleanup the sandbox, testing locally is painful

alanmarkoTrilitech:
> Also is it possible to fix the sandbox stopping bug in this PR -- without the ability to easily cleanup the sandbox, testing locally is painful

well yes, but a bit later in the week. I wanted to fix the issue together with adding jstz sandbox reset command. 
Also it seems to work if you just run sanbox stop twice instead of just once.

crates/jstz_cli/src/sandbox/mod.rs
alanmarkoTrilitech: The point of the flag is to NOT daemonize  the process. I don't think there is any standard for it though a possible alternative is --attach. However, since there is no point in using it I have decided to hide it as it is only used internally.

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -153,7 +168,13 @@ fn generate_identity(cfg: &Config) -> Result<()> {
     Ok(())
 }
 
-fn init_node(cfg: &Config) -> Result<()> {
+const MAX_PROGRESS: u32 = 16;
```
johnyob, line 171: Maybe a comment on how this constant was produced?

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -457,3 +467,99 @@ pub async fn main(cfg: &mut Config) -> Result<()> {
...
+                break;
+            }
+
+            if let Ok(Some(status)) = child.try_wait() {
+                progress_bar
+                    .finish_with_message(format!("Sandbox failed to start: {:}", status));
+                break;
+            }
+
+            thread::sleep(Duration::from_millis(100));
```
johnyob, line 536: Note to self: the mix of tokio + threading is becoming dangerous (and causing issues)

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -1,23 +1,34 @@
...
 use tokio::task;
 
+// Todo: @alanmarkoTrilitech - make it so that it behaves like a log and outputs to terminal when the env flag is set
+macro_rules! debug {
+    ($($arg:tt)*) => {
+        println!($($arg)*);
+        io::stdout().flush().unwrap();
+    };
+}
+
```
johnyob, line 31: Fix before merging

crates/octez/src/thread.rs
```diff
@@ -44,6 +44,27 @@ impl OctezThread {
...
+                Ok(())
+            }
+        });
+
+        Self {
+            shutdown_tx,
+            thread_handle,
+        }
+    }
+
```
johnyob, line 67: Is this used?

crates/jstz_node/src/config.rs
johnyob, line 1: Empty file

crates/jstz_cli/src/sandbox/mod.rs
```diff
@@ -17,15 +19,41 @@ use crate::{config::Config, error::bail_user_error};
...
+                break;
+            }
+            Err(e) => {
+                // An unexpected error occurred
+                bail_user_error!("Failed to kill the sandbox process: {:?}", e)
+            }
+        }
+    }
     Ok(())
 }
```
johnyob, line 58: Is this used?

crates/jstz_cli/src/sandbox/mod.rs
johnyob: There is a standard, namely _all_ long-lived processes don't automatically detach unless explicitly specified. 

`--detach` is inspired by `docker run --detach`

For all examples of not implicitly detaching long-lived processes, simply look at:
- `wrangler dev`
- `vercel dev`
- Most HTTP dev servers for frontends (e.g. react, vue, etc)

johnyob (changes requested):

crates/octez/src/node.rs
```diff
@@ -39,7 +39,7 @@ impl OctezNode {
             self.octez_node_dir.to_str().expect("Invalid path"),
             "--net-addr",
             http_endpoint,
-            "--rpc-addr",
+            "--local-rpc-addr", // TODO: @alanmarkoTrilitech --local-rpc-addr is not present in future versions of octez-node
```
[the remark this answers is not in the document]
alanmarkoTrilitech: The purpose of adding --singleprocess is to make it so that octez-validator is run together with octez-node, thus killing it is easier. This only works when --local-rpc-addr is provided, as we have decided to work with a specific octez commit instead of master. In later versions --rpc-addr should be used as it was merged into a single flag. I don't think it would be productive to make the switch now, we can do it though right before releasing the crate for dog fooding.

crates/jstz_cli/src/sandbox/daemon.rs
alanmarkoTrilitech: added

crates/jstz_cli/src/sandbox/daemon.rs
alanmarkoTrilitech: output is now redirected to file

crates/octez/src/thread.rs
alanmarkoTrilitech: no, this stayed there from previous versions - removed now. would be nice if the compiler detected it.

crates/jstz_cli/src/sandbox/mod.rs
alanmarkoTrilitech: yes, it doesn't get killed immediately.

crates/jstz_cli/src/sandbox/mod.rs
alanmarkoTrilitech: changed it so that it does both --detach and without --detach (required some changes to how stopping/logging is handled). However thinking about it, I don't like this change  much - using it detached seems to me much better and easier to use to me and should be a default option IMO. I would rather have --attach flag and for sandbox start without any flag I would just output a line saying that it was started in background and can be killed with sandbox stop.

crates/jstz_cli/src/sandbox/mod.rs
```diff
@@ -1,52 +1,44 @@
...
 pub enum Command {
     /// 🎬 Starts the sandbox.
-    Start,
+    Start {
+        /// Detach the process to run in the background.
+        #[clap(long, short, default_value = "false")]
+        detach: bool,
+        /// Run the sandbox in the background without showing any output.
+        #[clap(long, short, default_value = "false", hide = true)]
+        background: bool,
```
johnyob, line 21: Two flags? What is the difference here

crates/jstz_cli/src/sandbox/mod.rs
johnyob: I think we'll simply have to agree to disagree here

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -1,23 +1,37 @@
...
 use tempfile::TempDir;
 use tokio::task;
 
+macro_rules! debug {
+    ($file:expr, $($arg:tt)*) => {
+        writeln!($file, $($arg)*).expect("Failed to write to log file");
+        $file.flush().expect("Failed to flush log file");
+    };
+}
+
```
johnyob, line 34: I think it is possible to avoid this hacky debug macro by using `env_logger` to configure the output to a file

crates/jstz_cli/src/sandbox/mod.rs
johnyob: It appears --background is an internal flag for starting the sandbox as another process. It would be good to document that it is internal

examples/counter.js
```diff
@@ -1,6 +1,8 @@
 const KEY = "counter";
 
 const handler = () => {
+  let a = [];
+  a[1].foo();
```
johnyob, line 5: Remove

crates/jstz_cli/src/sandbox/mod.rs
alanmarkoTrilitech: It is hidden, thus it won't be shown in help or documentation as there is no reason for the user to be able to access it.

johnyob: I'm aware, what I meant was that we should document that this flag _wont_ be used by the user (the documentation is for us, not the user)

crates/jstz_cli/src/sandbox/daemon.rs
johnyob: Added ticket to fix in the future

alanmarkoTrilitech (commented):
the commit changes look good to me, except that the ci complains.
Maybe printing something like - "sandbox is running in background, use jstz sandbox stop to stop" would be useful after the progressbar disappears.
 
You will have to approve the pr though.

johnyob:
Example output:
image

johnyob (approved):
lgtm 🚀 

Commit history needs clearing up before merging -- I recommend squashing into 1/2 commits (1 for the dameonizing and 1 for the progress bar)

PR #370, opened 2024-02-06
fix(node): Allow JSON `SignedOperations` for `POST /operations`

johnyob:
# Context

# Description

This PR fixes a known issue about `jstz-node` -- namely that `POST /operations` accepts bytes as opposed to a `SignedOperation`. 

This was due to an issue with the serializer and deserializer for `SignedOperation` while it was using tz4 addresses (from `tezos_crypto_rs`). Following #352, we now use tz1 addresses (which have a working serializer & deserializer 😊). 

# Manually testing the PR

With the sandbox + jstz node running:
```sh
cargo run --bin jstz -- deploy examples/counter.js
```

Inspect the jstz-node logs for a `POST /operations` call.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1924,7 +1924,6 @@ version = "0.1.0"
 dependencies = [
  "ansi_term",
  "anyhow",
- "bincode",
  "boa_engine",
  "boa_gc",
  "bs58",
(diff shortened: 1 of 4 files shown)


alanmarkoTrilitech (approved):
LGTM!

PR #384, opened 2024-02-09
Node: fix running the server in main

ryutamago:
# Context
Executing the jstz_node binary outside of sandbox environment leads to a panic. This problem arises when spawning the broadcaster future that expects to be called inside the `LocalSet`, a set of task executed on the same thread.

# Description
This PR fixes the issue by replacing the tokio's runtime with the #[actix_web::main].

# Manually testing the PR

run the jstz_node binary as well as the sandbox to see if it works.


johnyob (approved):
lgtm

PR #387, opened 2024-02-11
feat: implemented `CountQueuingStrategy`

xavierm02:
# Description

This implements part of the [queuing strategy](https://streams.spec.whatwg.org/#qs-api) section of the Stream WebAPI. The main goal was to implement `CountQueuingStrategy`, but since most of the code could be shared through simple macros, most of `ByteLengthQueuingStrategy` is also implemented (with the only remaining being finishing the implementation of its size algorithm).

# Manual testing

```sh
nix develop
cargo run --bin jstz -- repl

>> new CountQueuingStrategy()
✕  ERROR  Uncaught TypeError: CountQueuingStrategy constructor: At least 1 argument required, but only 0 passed
>> new CountQueuingStrategy(1)
✕  ERROR  Uncaught TypeError: Missing required highWaterMark property
>> new CountQueuingStrategy({})
✕  ERROR  Uncaught TypeError: Missing required highWaterMark property

>> strat = new CountQueuingStrategy({highWaterMark: 1})

>> strat.highWaterMark
1
>> strat.highWaterMark = 2 // Does nothing because highWaterMark is readonly
2
>> strat.highWaterMark
1

>> strat.size()
1
>> strat.size(3)
1
>> strat.size({})
1

>> bad_strat = new CountQueuingStrategy({highWaterMark: -1}) // Creating this object is fine
>> bad_strat.highWaterMark
-1
>> new ReadableStream({}, bad_strat) // But using it in ReadableStream's constructor causes an error
✕  ERROR  Uncaught RangeError: Invalid highWaterMark

>> new ReadableStream({}, strat)
thread 'main' panicked at crates/jstz_api/src/stream/readable/mod.rs:67:13:
not yet implemented: SetUpReadableStreamDefaultControllerFromUnderlyingSource
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
# Checklist

### DIFF
--- a/crates/jstz_api/src/stream/queuing_strategy/builtin.rs
+++ b/crates/jstz_api/src/stream/queuing_strategy/builtin.rs
@@ -1,0 +1,216 @@
+use crate::{
+    idl,
+    stream::{
+        queuing_strategy::size::{
+            ByteLengthQueuingStrategySizeAlgorithm, CountQueuingStrategySizeAlgorithm,
+        },
+        tmp::get_jsobject_property,
+    },
+};
+use boa_engine::{
+    js_string, object::Object, property::Attribute, value::TryFromJs, Context, JsArgs,
+    JsNativeError, JsResult, JsValue, NativeFunction,
+};
+use boa_gc::{Finalize, GcRefMut, Trace};
+use jstz_core::{
+    accessor,
+    js_fn::JsCallableWithoutThis,
+    native::{Accessor, ClassBuilder, JsNativeObject, NativeClass},
+};
+
+/// [Streams Standard - § 7.1.][https://streams.spec.whatwg.org/#qs-api]
+/// > ```
+/// > dictionary QueuingStrategyInit {
+/// >   required unrestricted double highWaterMark;
+/// > };
+/// > ```
+pub struct QueuingStrategyInit {
+    pub high_water_mark: idl::UnrestrictedDouble,
+}
+
+impl TryFromJs for QueuingStrategyInit {
+    fn try_from_js(value: &JsValue, context: &mut Context<'_>) -> JsResult<Self> {
+        let this = value.to_object(context)?;
+        let high_water_mark: idl::UnrestrictedDouble =
+            get_jsobject_property(&this, "highWaterMark", context)?
+                .try_js_into::<Option<idl::UnrestrictedDouble>>(context)?
+                .ok_or_else(|| {
+                    JsNativeError::typ()
+                        .with_message("Missing required highWaterMark property")
+                })?;
+        // TODO: .try_js_into::<Option<idl::UnrestrictedDouble>>(context) is slightly wrong. "arandomstring" should map to "NaN"
+        Ok(QueuingStrategyInit { high_water_mark })
+    }
+}
+
+/// Streams Standard - § 7.3. The CountQueuingStrategy class][https://streams.spec.whatwg.org/#cqs-class]
+/// > A common queuing strategy when dealing with streams of generic objects is to simply count the number of chunks that have been accumulated so far, waiting until this number reaches a specified high-water mark. As such, this strategy is also provided out of the box.
+///
+/// [Streams Standard - § 7.3.1.][https://streams.spec.whatwg.org/#countqueuingstrategy]
+/// > ```
+/// > [Exposed=*]
+/// > interface CountQueuingStrategy {
+/// >   constructor(QueuingStrategyInit init);
+/// >
+/// >   readonly attribute unrestricted double highWaterMark;
+/// >   readonly attribute Function size;
+/// > };
+/// > ```
+///
+#[derive(Trace, Finalize)]
+pub struct CountQueuingStrategy {
+    pub high_water_mark: idl::UnrestrictedDouble,
+}
+
+impl CountQueuingStrategy {
+    pub fn size_algorithm(&self) -> CountQueuingStrategySizeAlgorithm {
+        CountQueuingStrategySizeAlgorithm::ReturnOne
+    }
+}
+
+#[derive(Trace, Finalize)]
+pub struct ByteLengthQueuingStrategy {
+    pub high_water_mark: idl::UnrestrictedDouble,
+}
+
+impl ByteLengthQueuingStrategy {
+    pub fn size_algorithm(&self) -> ByteLengthQueuingStrategySizeAlgorithm {
+        ByteLengthQueuingStrategySizeAlgorithm::ReturnByteLengthOfChunk
+    }
+}
+
+macro_rules! impl_for_builtin_queuing_strategy_struct {
+    ($struct_name: ident, $struct_name_as_str: expr) => {
+        impl $struct_name {
+            pub fn new(init: QueuingStrategyInit) -> Self {
+                $struct_name {
+                    high_water_mark: init.high_water_mark,
+                }
+            }
+
+            pub fn try_from_js(value: &JsValue) -> JsResult<GcRefMut<Object, Self>> {
+                value
+                    .as_object()
+                    .and_then(|obj| obj.downcast_mut::<Self>())
+                    .ok_or_else(|| {
+                        JsNativeError::typ()
+                            .with_message(format!(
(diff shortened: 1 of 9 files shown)


crates/jstz_api/src/stream/readable/mod.rs
```diff
@@ -38,9 +44,28 @@ impl NativeClass for ReadableStreamClass {
...
-            Option::<UnderlyingSource>::try_from_js(args.get_or_undefined(0), context)?;
-        let _ = underlying_source;
-        todo!()
+            Option::<UnderlyingSource>::try_from_js(args.get_or_undefined(0), context)?
+                .unwrap_or_else(|| todo!());
+        let queuing_strategy =
+            Option::<QueuingStrategy>::try_from_js(args.get_or_undefined(1), context)?
+                .unwrap_or_default();
+
+        // TODO Perform ! InitializeReadableStream(this).
```
alanmarkoTrilitech, line 53: Can these todos stay here?

crates/jstz_api/src/stream/queuing_strategy/mod.rs
```diff
@@ -0,0 +1,67 @@
+//! [Streams Standard - § 7. Queuing strategies][https://streams.spec.whatwg.org/#qs]
+//!
+//! A few things to keep in mind:
+//!
+//! - The queuing strategy given to a stream constructor can be either builtin one (i.e. an instance of `CoutQueuingStrategy` or `ByteLengthQueuingStrategy`)
```
alanmarkoTrilitech, line 5:
```suggestion
//! - The queuing strategy given to a stream constructor can be either builtin one (i.e. an instance of `CountQueuingStrategy` or `ByteLengthQueuingStrategy`)
```

alanmarkoTrilitech (commented):
Hi @xavierm02,

It looks good to me, I just have a few questions. 

Firstly, could you please also provide some more detailed context for these changes? I understand that this is based on the specification, however there is also a reference implementation in https://github.com/whatwg/streams repo. Do you aim to follow this implementation more closely or primarily just trying to make it spec compliant? If trying to follow the implementation, I think it would be nice to refer to it in comments and point out what are the differences if any. 

Also, is it possible to test whether it is actually spec compliant with WPT tests?

alanmarkoTrilitech (commented):
Also, I am not sure if this was explored at some point, but isn't it possible to sometimes just implement a class in Javascript(or use a reference implementation) and just evaluate through boa in case that it doesn't require access to anything internal that needs to be implemented in rust like Kv storage? 

@johnyob

crates/jstz_api/src/stream/readable/mod.rs
xavierm02: Yes, I partially implemented the `ReadableStream` constructor to allow testing `UnderlyingSource` and `QueuingStrategy`, but since the implementation is not complete I added `todo!` and `TODO` comments (with `TODO` comments for everything that should be added before the parts used for the tests). Whoever will eventually implement this will probably know that these steps are missing, but I think having a comment there doesn't hurt

xavierm02:
> Also, I am not sure if this was explored at some point, but isn't it possible to sometimes just implement a class in Javascript(or use a reference implementation) and just evaluate through boa in case that it doesn't require access to anything internal that needs to be implemented in rust like Kv storage?
> 
> @johnyob

I recall talking about this with @johnyob, and he preferred a Rust implementation (probably for performance reasons, but I'm not sure).

crates/jstz_api/src/stream/queuing_strategy/builtin.rs
```diff
@@ -0,0 +1,216 @@
...
+impl TryFromJs for QueuingStrategyInit {
+    fn try_from_js(value: &JsValue, context: &mut Context<'_>) -> JsResult<Self> {
+        let this = value.to_object(context)?;
+        let high_water_mark: idl::UnrestrictedDouble =
+            get_jsobject_property(&this, "highWaterMark", context)?
+                .try_js_into::<Option<idl::UnrestrictedDouble>>(context)?
+                .ok_or_else(|| {
+                    JsNativeError::typ()
+                        .with_message("Missing required highWaterMark property")
+                })?;
```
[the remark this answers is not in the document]
xavierm02: Hm. The whole `impl TryFromJs for QueuingStrategyInit` should be derived. The `get_jsobject_property` function is just something I added as a temporary fix because the derived `TryFromJs` didn't work for `UnderlyingSource`, and it seems to also not work for `QueuingStrategyInit`: running `strat = new CountQueuingStrategy({highWaterMark: 1})` yields an error `✕  ERROR  Uncaught TypeError: cannot get property high_water_mark of value` if I use
```
#[derive(TryFromJs)]
pub struct QueuingStrategyInit {
    pub high_water_mark: idl::UnrestrictedDouble,
}
```
(which may be caused by the difference between Rust naming convensions and JS ones, i.e. `high_water_mark` vs `highWaterMark`, but I haven't looked into it)

johnyob:
> I recall talking about this with @johnyob, and he preferred a Rust implementation (probably for performance reasons, but I'm not sure).

Yes, IIRC the primary reason behind a Rust implementation was:
1. Performance (Streams are used in most Web API specs)
2. Avoiding evaluating large polyfills on function load (Boa doesn't offer snapshots)
3. Integrating with DAL

Having investigated SpiderMonkey more (the engine we're switching to) -- 
1. Can be solved as SpiderMonkey provides a _partial_ implementation of the WHATWG spec for streams and 
2. Could be solved with snapshots
3. Can be solved by FFI stubs to SpiderMonkey's C++ stream impl.

johnyob (approved):

xavierm02:
> Hi @xavierm02,
> 
> It looks good to me, I just have a few questions.
> 
> Firstly, could you please also provide some more detailed context for these changes? I understand that this is based on the specification, however there is also a reference implementation in https://github.com/whatwg/streams repo. Do you aim to follow this implementation more closely or primarily just trying to make it spec compliant? If trying to follow the implementation, I think it would be nice to refer to it in comments and point out what are the differences if any.
> 
> Also, is it possible to test whether it is actually spec compliant with WPT tests?

The spec already contains a sort of reference implementation (though in a non-executable language) that's more precise that the JS reference implementation, so I follow this and not the JS implementation. (Initially, [I wanted to the spec algorithms as closely as possible](https://tezos-dev.slack.com/archives/C061SSDBN69/p1700840136229009), but @johnyob preferred deviating from the spec (in terms of implementation, not externally observable behavior) to make things more Rusty.) The main difference between the spec and this implementation for `QueuingStrategy` is the use of structs and traits where the spec uses inheritance, methods and functions, but that's just the expected Rust way of doing things and is not really specific to `QueuingStrategy`.

I haven't looked at WPT tests, but I don't think they would be useful yet: we can't even create a `ReadableStream` yet, only turn the arguments of its constructor into Rust structs.

alanmarkoTrilitech:
> > I recall talking about this with @johnyob, and he preferred a Rust implementation (probably for performance reasons, but I'm not sure).
> 
> Yes, IIRC the primary reason behind a Rust implementation was:
> 
> 1. Performance (Streams are used in most Web API specs)
> 2. Avoiding evaluating large polyfills on function load (Boa doesn't offer snapshots)
> 3. Integrating with DAL
> 
> Having investigated SpiderMonkey more (the engine we're switching to) --
> 
> 1. Can be solved as SpiderMonkey provides a _partial_ implementation of the WHATWG spec for streams
> 2. Could be solved with snapshots (which SpiderMonkey does provide)
> 3. Can be solved by FFI stubs to SpiderMonkey's C++ stream impl.

ok, makes sense. what does it imply for the future once we switch to spidermonkey?  could we then use mostly Javascript implementation for new APIs?

crates/jstz_api/src/stream/queuing_strategy/builtin.rs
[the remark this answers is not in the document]
xavierm02: See https://tezos-dev.slack.com/archives/C061SSDBN69/p1701430478351919

PR #397, opened 2024-02-14
Node: persistent logging feature

ryutamago:
# Context

The presistent logging gives warning in the sadbox environement due to the deterministc nature of the request id. we want to be able to disable the persistnet logging
[Task](https://app.asana.com/0/1205763124050213/1206544923875669) 

# Description

Add a feature to be able to turn on the persistent log.

# Manually testing the PR

run the sandbox and do 
`sqlite3 ~/.jstz/log.db "SELECT * FROM log"`
to see if the flushing is diabled after running the contract

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ jstz_api = { path = "crates/jstz_api" }
 jstz_kernel = { path = "crates/jstz_kernel" }
 jstz_proto = { path = "crates/jstz_proto" }
 jstz_crypto = { path = "crates/jstz_crypto" }
-jstz_node = { path = "crates/jstz_node" }
+jstz_node = { path = "crates/jstz_node"}
 jstz_cli = { path = "crates/jstz_cli" }
 jstz_wpt = { path = "crates/jstz_wpt" }
 octez = { path = "crates/octez" }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ jstz_api = { path = "crates/jstz_api" }
 jstz_kernel = { path = "crates/jstz_kernel" }
 jstz_proto = { path = "crates/jstz_proto" }
 jstz_crypto = { path = "crates/jstz_crypto" }
-jstz_node = { path = "crates/jstz_node" }
+jstz_node = { path = "crates/jstz_node", features = ["persistent-logging"]}
```
johnyob, line 29: Avoid enabling `persistent-logging` by default here

crates/jstz_cli/Cargo.toml
```diff
@@ -15,7 +15,7 @@ jstz_proto.workspace = true
 jstz_core.workspace = true
 jstz_api.workspace = true
 jstz_rollup.workspace = true
-jstz_node.workspace = true
+jstz_node = { path = "../jstz_node"}
```
johnyob, line 18: Switch back to `jstz_node.workspace = true`

Cargo.toml
```diff
@@ -26,7 +26,7 @@ jstz_api = { path = "crates/jstz_api" }
 jstz_kernel = { path = "crates/jstz_kernel" }
 jstz_proto = { path = "crates/jstz_proto" }
 jstz_crypto = { path = "crates/jstz_crypto" }
-jstz_node = { path = "crates/jstz_node" }
+jstz_node = { path = "crates/jstz_node"}
```
johnyob, line 29: Nit: Strange the formatter didn't catch this, but this diff is unnecessary. Recommend reverting this

johnyob (approved):
LGTM.

Please fix the nit + squash the commit history before merging

PR #405, opened 2024-02-19
docs(api): update and improve runtime API docs

johnyob:
# Context

**Related Tasks**: [API doc: Update & Review](https://app.asana.com/0/1205770721173531/1206249753816554/f)

# Description

This PR updates the runtime API docs:
- adding various missing methods
- removes old / outdated information
- improves styling consistency across the API documentation

# Manually testing the PR

```
npm run docs:dev
```


docs/api/headers.md
```diff
@@ -29,11 +29,11 @@ console.log(authHeader); // "Bearer 123456789"
 
 ## Referer
 
-The `Referer` header is a special header that is automatically set by `jstz` when it makes a request. The value of the `Referer` header is the `tz4` address of the smart function (or account) that made the request.
+The `Referer` header is a special header that is automatically set by `jstz` when it makes a request. The value of the `Referer` header is the `tz1` address of the smart function (or account) that made the request.
 
 ```typescript
 async function handler(request: Request): Promise<Response> {
-    const referer = request.headers.get("Referer"); // "tz4..."
+    const referer = request.headers.get("Referer"); // "tz1..."
```
alanmarkoTrilitech, line 36: do you know about any docs that say what exactly should be the difference between use of tz1, ... tz4 addresses?

johnyob: They're just different keypair algorithms -- tz1 is ed25519 and tz4 is BLS. Not aware of any docs apart from reading the code of the Tezos protocol

docs/api/encoding.md
```diff
@@ -3,7 +3,9 @@
 An implementation of the Web standard encoding API.
 The Encoding API provides a mechanism for handling text in various character encodings, including legacy non-UTF-8 encodings.
 
-::: danger ⚠️ jstz's implementation is not fully spec compliant ⚠️ :::
+::: danger
+⚠️ `jstz`'s implementation is not fully spec compliant ⚠️
```
sam-finch-tezos, line 7: Can we say why not or would the list of reasons be embarrassingly long.

sam-finch-tezos (approved):
lgtm

docs/api/encoding.md
johnyob: For encoding I think it is simply the fact that we don't support text encoder / decoder streams

johnyob: But in other cases, the list would be embarrasingly long 😬

PR #412, opened 2024-02-20
feat(cli): rearranging commands + new help messages

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1206577693474288

# Description
This PR rearranges commands in help and creates an easier to read help messages by grouping commands.

# Manually testing the PR
jstz help


johnyob (changes requested):
While I understand the purpose of this PR, I'm not a fan of it's implementation as I'd prefer to somehow find a way to group the commands using clap instead of monkey-patching the help command like this. 

What I can agree on is the ordering of the commands -- I propose a compromise PR which swaps the order of the `Command` constructors to fit the order of commands listed in `print_custom_help()`

alanmarkoTrilitech:
closed in favour of alan@cli/help_grouping

PR #414, opened 2024-02-21
feat(cli): deployment streamlined

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1206577693474286

# Description
This PR includes account login and sandbox start in deploy. Additionally it simplifies messages and fixes a few bugs with config not being saved.

# Manually testing the PR
jstz deploy examples/counter.js

### DIFF
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -168,15 +168,14 @@ pub fn login(alias: String) -> Result<()> {
     }
 }
 
-pub fn login_quick() -> Result<()> {
-    let cfg = Config::load()?;
+pub fn login_quick(cfg: &mut Config) -> Result<()> {
     if cfg.accounts.current_user().is_none() {
         let account_alias: String = Input::new()
                 .with_prompt("You are not logged in. Please type the account name that you want to log into or create as new")
                 .interact()?;
 
         login(account_alias)?;
-        println!();
+        info!("");
     }
     Ok(())
 }
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -168,6 +168,19 @@ pub fn login(alias: String) -> Result<()> {
     }
 }
 
+pub fn login_quick() -> Result<()> {
+    let cfg = Config::load()?;
+    if cfg.accounts.current_user().is_none() {
+        let account_alias: String = Input::new()
+                .with_prompt("You are not logged in. Please type the account name that you want to log into or create as new")
+                .interact()?;
+
+        login(account_alias)?;
+        println!();
+    }
+    Ok(())
+}
+
 pub fn logout() -> Result<()> {
     let mut cfg = Config::load()?;
 
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -406,6 +406,15 @@ impl Config {
         Ok(network?.clone())
     }
 
+    pub fn network_name(&self, name: &Option<NetworkName>) -> Result<NetworkName> {
+        match name {
+            Some(name) => Ok(name.clone()),
+            None => self.networks.default_network.clone().ok_or_else(|| {
+                user_error!("No default network found in the config file. Please specify the `--network` flag or set the default network in the config file.")
+            }),
+        }
+    }
+
     fn lookup_network(&self, name: &NetworkName) -> Result<Network> {
         match name {
             NetworkName::Custom(name) => {
--- a/crates/jstz_cli/src/deploy.rs
+++ b/crates/jstz_cli/src/deploy.rs
@@ -1,4 +1,5 @@
 use boa_engine::JsError;
+use dialoguer::Confirm;
 use dialoguer::Input;
 use jstz_proto::{
     context::account::ParsedCode,
@@ -13,7 +14,7 @@ use crate::{
     error::{bail, bail_user_error, user_error, Result},
     sandbox::daemon,
     term::styles,
-    utils::{get_file_name_from_path, read_file_or_input_or_piped},
+    utils::read_file_or_input_or_piped,
 };
 
 pub async fn exec(
@@ -32,30 +33,34 @@ pub async fn exec(
         || (network.is_none() && cfg.networks.default_network == Some(NetworkName::Dev)))
         && cfg.sandbox.is_none()
     {
-        daemon::main(true, false, &mut cfg).await?;
-        info!(
-            "Use `{}` to start from a clear sandbox state.\n",
-            styles::command("jstz sandbox restart --detach")
-        );
-        cfg = Config::load()?;
+        info!("No sandbox is currently running.");
+
+        let proceed = Confirm::new()
+            .with_prompt(format!("Start the sandbox in daemon mode now? Tip: Use '{}' for an interactive session instead.", styles::command("jstz sandbox start")))
+            .default(true)
+            .interact()?;
+
+        if proceed {
+            // User confirmed, start the sandbox
+            daemon::main(true, false, &mut cfg).await?;
+            info!(
+                "Use `{}` to start from a clear sandbox state.\n",
+                styles::command("jstz sandbox restart --detach")
+            );
+        } else {
+            bail_user_error!(
+                "Please start the sandbox before deploying a smart function to dev network."
+            );
+        }
     }
 
     // Get the current user and check if we are logged in
-    let (user_name, user) = {
-        if cfg.accounts.current_user().is_none() {
-            let account_alias: String = Input::new()
-                    .with_prompt("You are not logged in. Please type the account name that you want to log into or create as new")
-                    .interact()?;
-
-            account::login(account_alias)?;
-            println!();
-
-            cfg = Config::load()?;
-        }
-        cfg.accounts.current_user().ok_or(user_error!(
-            "Failed to setup the account. Please try `jstz login`."
-        ))?
-    };
+    account::login_quick()?;
+    cfg = Config::load()?;
+    let (user_name, user) = cfg.accounts.current_user().ok_or(user_error!(
+        "Failed to setup the account. Please try `{}`.",
+        styles::command("jstz login")
+    ))?;
 
     // 1. Check if smart function account already exists
     if let Some(name) = &name {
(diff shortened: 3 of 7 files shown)


johnyob (changes requested):
Personally I don't think the login, naming and sandbox features this PR implements is particularly helpful in their current state.
1. The CLI will already prompt users to use `jstz login` if they are not currently logged in.
    This PR introduces inconsistencies in the login-prompting across other commands -- however, the changes would be good if we regained this consistency.
2. I don't see value in the automatic name generation; users can effortlessly supply a meaningful name using the '--name' flag. I believe it's acceptable for users to use an address directly unless they wish to specify a name.
3. Users should be aware of networks and whether they are running or not. Additionally they should be comfortable using the sandbox commands directly. 

    The analogy to other CLI's here is the `dev` command. Wrangler, denoctl, etc don't automatically start a dev environment if the user tries deploying a function.

johnyob:
>  There are clearly at least some features and ideas in this PR that are useful - for example it simplifies the output, so that it doesn't show the huge json log and shows the command to run the next function. 

This is true, there are two features in this PR that I do find useful:
1. The improvements in the command output
2. Automatically logging the user in (provided this feature was replicated on other commands as well)

> Furthermore, I have mentioned before that I would like to make these and other changes in jstz-dev and there was a general agreement.

There was agreement to points 2, 4 (i) and 5 from https://tezos-dev.slack.com/archives/C061SSDBN69/p1706626070925749?thread_ts=1706101009.261069&cid=C061SSDBN69. There was no agreement/concensus on:
- 'running sandbox start automatically if you run deploy and sandbox is not running"
- 'having "sandbox restart" that combines start+stop'
- 'naming deployed sf alias automatically based on the name of the .js file'

I would've preferred more discussion around these points before implementing them in a PR. 

> Well, maybe you don't see the value in it

I do see the value in generating names, my primary issue with the implementation is that across sandboxes + networks, the namespace is shared. Meaning that you'll often have dead/old smart function entries in the config. By not naming smart functions by default, we avoid this issue. 

> We can't just force users to be comfortable using complicated commands

I'd hardly call starting the sandbox environment 'complicated'

> Regarding the point that the analogy here are other dev commands like Wrangler, denoctl etc. I understand that they are good for inspiration and making analogies with cloud, however they also don't strike me as a particularly easy to use interface.

I think we'll agree to disagree here. While it's true that these tools aren't perfect, their presence and usage within the developer community suggest a certain level of familiarity and acceptance. By aligning our CLI with established tools, we tap into the collective experience of developers who are accustomed to similar conventions. This can significantly enhance the DX by reducing the learning curve for users already familiar with those tools. It also promotes a sense of consistency and standardization within the development ecosystem, making it easier for users to switch between different tools seamlessly.

johnyob:
> I think we could agree on the following compromise:
> - The improvements in the command output stays
> - Automatically logging the user in with replication for other commands (mainly run)

Agreed

> for generating names we either merge the current version with names as the behaviour is the same as currently with --name flag (it throws an error if the function doesn't exist for the network so I don't think it is such an issue)
or fix accounts to work properly when network changes and then add function naming
or change the created name to "network_name", e.g. "dev_counter", "mainnet_mynft"

I'd favour for fixing names later and merging this feature after it is fixed

> The sandbox would not start automatically, but there would instead be a message asking if the user wants to start it in background with reference to "sandbox start" command to start non-daemonized version.

Agreed that printing `jstz sandbox start` would be a good message if the network is `dev` and the sandbox hasn't been started

crates/jstz_cli/src/deploy.rs
```diff
@@ -1,25 +1,59 @@
...
     network: Option<NetworkName>,
 ) -> Result<()> {
     let mut cfg = Config::load()?;
 
-    let (_, user) = cfg.accounts.current_user().ok_or(user_error!(
-        "You are not logged in. Please run `jstz login`."
+    // Load sandbox if network is Dev or it is None and cfg network is dev, and sandbox is not already loaded
+    if (network == Some(NetworkName::Dev)
+        || (network.is_none() && cfg.networks.default_network == Some(NetworkName::Dev)))
+        && cfg.sandbox.is_none()
```
johnyob, line 28: I believe there should be a function in `Config` for getting the current network name

crates/jstz_cli/src/config.rs
```diff
@@ -242,7 +242,7 @@ struct Network {
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct NetworkConfig {
     // if None, the users have to specify the network in the command
-    default_network: Option<NetworkName>,
+    pub default_network: Option<NetworkName>,
```
johnyob, line 245: If you use the `Config` methods, you should be able to avoid making this public

crates/jstz_cli/src/account.rs
```diff
@@ -168,6 +168,19 @@ pub fn login(alias: String) -> Result<()> {
...
 
+pub fn login_quick() -> Result<()> {
+    let cfg = Config::load()?;
+    if cfg.accounts.current_user().is_none() {
+        let account_alias: String = Input::new()
+                .with_prompt("You are not logged in. Please type the account name that you want to log into or create as new")
+                .interact()?;
+
+        login(account_alias)?;
+        println!();
```
johnyob, line 179:
```suggestion
        info!();
```

crates/jstz_cli/src/account.rs
```diff
@@ -168,6 +168,19 @@ pub fn login(alias: String) -> Result<()> {
...
+    }
+    Ok(())
+}
+
```
johnyob, line 183: Is it possible to take `cfg` as a parameter?

crates/jstz_cli/src/deploy.rs
```diff
@@ -1,25 +1,59 @@
...
+        } else {
+            bail_user_error!(
+                "Please start the sandbox before deploying a smart function to dev network."
+            );
+        }
+    }
+
+    // Get the current user and check if we are logged in
+    account::login_quick()?;
+    cfg = Config::load()?;
```
johnyob, line 53: `cfg.reload()`?

crates/jstz_cli/src/deploy.rs
```diff
@@ -1,25 +1,59 @@
...
+            info!(
+                "Use `{}` to start from a clear sandbox state.\n",
+                styles::command("jstz sandbox restart --detach")
+            );
+        } else {
+            bail_user_error!(
+                "Please start the sandbox before deploying a smart function to dev network."
+            );
+        }
+    }
```
johnyob, line 49: I thought we agreed not to automatically start the sandbox?

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -480,6 +480,7 @@ pub async fn run_sandbox(cfg: &mut Config) -> Result<()> {
     run_jstz_node().await?;
     OctezThread::join(vec![baker, rollup_node, node])?;
 
+    let mut cfg = Config::load()?;
```
johnyob, line 483: `cfg.reload()`

crates/jstz_cli/src/deploy.rs
alanmarkoTrilitech: this doesn't start it automatically - it asks it with a message whether they want to start it in background and says how to start the interactive session with just "jstz sandbox start".

crates/jstz_cli/src/deploy.rs
alanmarkoTrilitech: there is currrently no function like this - I've added a new one

crates/jstz_cli/src/config.rs
alanmarkoTrilitech: done

crates/jstz_cli/src/account.rs
alanmarkoTrilitech: done

crates/jstz_cli/src/account.rs
alanmarkoTrilitech: yes, done

crates/jstz_cli/src/deploy.rs
alanmarkoTrilitech: done

crates/jstz_cli/src/sandbox/daemon.rs
alanmarkoTrilitech: done

johnyob (changes requested):
Everything looks good modulo the sandbox start

crates/jstz_cli/src/deploy.rs
alanmarkoTrilitech: I've removed the interaction so that it moves quicker, but I don't see any issue with it and would prefer adding it later at least during development as it makes testing much more pleasant.

crates/jstz_cli/src/deploy.rs
```diff
@@ -90,9 +99,24 @@ pub async fn exec(
...
+    // Show message showing how to run the smart function
+    // TODO: add --trace flag
+    let network_flag = match network {
+        Some(network) => format!(" --network {}", network),
+        None => "".to_string(),
+    };
+    info!(
+        "Run with `{}{}{}`",
+        styles::command("jstz run "),
+        styles::url(format!("tezos://{}/<args>", address)),
```
johnyob, line 116: Maybe remove the `<args>` (user will be always remove this)

crates/jstz_cli/src/deploy.rs
```diff
@@ -7,21 +7,35 @@ use jstz_proto::{
...
+        );
+    }
 
-    let (_, user) = cfg.accounts.current_user().ok_or(user_error!(
-        "You are not logged in. Please run `jstz login`."
+    // Get the current user and check if we are logged in
+    account::login_quick(&mut cfg)?;
+    cfg.reload()?;
+    let (user_name, user) = cfg.accounts.current_user().ok_or(user_error!(
+        "Failed to setup the account. Please try `{}`.",
```
johnyob, line 37: Consistency with other messages

```suggestion
        "Failed to setup the account. Please run`{}`.",
```

johnyob (commented):
Overall LGTM

Only some nits to fix

crates/jstz_cli/src/deploy.rs
alanmarkoTrilitech, line 116:
```suggestion
        styles::url(format!("tezos://{}/", address)),
```

crates/jstz_cli/src/deploy.rs
```diff
@@ -7,21 +7,35 @@ use jstz_proto::{
...
     name: Option<String>,
     network: Option<NetworkName>,
 ) -> Result<()> {
     let mut cfg = Config::load()?;
+    // Load sandbox if the selected network is Dev and sandbox is not already loaded
+    if cfg.network_name(&network)? == NetworkName::Dev && cfg.sandbox.is_none() {
+        bail_user_error!(
+            "No sandbox is currently running. Run {} for an interactive session or {} to run in the background.",
+            styles::command("jstz sandbox start"),
+            styles::command("jstz sandbox start --detach")
```
alanmarkoTrilitech, line 29:
```suggestion
            styles::command("jstz sandbox start")
```

johnyob (approved):
lgtm 🚢 🚀

PR #415, opened 2024-02-21
fix: check smart function code parses before continuing

sam-finch-tezos:
# Context
Currently if we deploy a contract it is not checked whether it parses or not. This means the kernel panics when the contract is called.  This PR fixes that by raising an error when the contract is deployed. 

**Related Tasks**: [Add check to see if JS parses when deploying contract](https://app.asana.com/0/1205770721173531/1205770721309638/f)

# Description
* Introduces a new type `ParsedCode` as a witness that the code has been parsed.
* Changes the deployment logic to uses ParsedCode and not String.
* Updates the CLI to pass ParsedCode not strings so that we can perform the checks before deploying.

# Manually testing the PR

```bash
# for the cli
echo "fn default(request: Request) -> Response { response::new() }" | jstz deploy # parse error

# for the rollup
jstz deploy --name test_deploy examples/deploy.js
jstz run tezos://test_deploy
```

### DIFF
--- a/crates/jstz_proto/src/context/account.rs
+++ b/crates/jstz_proto/src/context/account.rs
@@ -1,6 +1,10 @@
-use std::fmt::Debug;
+use std::{
+    fmt::{self, Display},
+    result,
+};
 
 use crate::error::{Error, Result};
+use boa_engine::{Context, JsError, JsResult, Module, Source};
 use jstz_core::{
     host::HostRuntime,
     kv::{Entry, Transaction},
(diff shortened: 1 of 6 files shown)


crates/jstz_proto/src/context/account.rs
```diff
@@ -33,11 +34,37 @@ impl ToString for Nonce {
...
+    }
+}
+impl Display for ParsedCode {
+    fn fmt(
+        &self,
+        formatter: &mut std::fmt::Formatter<'_>,
+    ) -> std::result::Result<(), std::fmt::Error> {
+        Display::fmt(&self.0, formatter)
+    }
+}
```
johnyob, line 52: nit: usually we avoid `std::...` and prefer `fmt::Formatter` and `fmt::Result<()>`

johnyob (approved):
lgtm 🚀 

Only 1 nit to fix :)

alanmarkoTrilitech (approved):
LGTM!

PR #416, opened 2024-02-21
feat(cli): Add explicit error message for long smart functions

sam-finch-tezos:
# Context
As we have not yet implemented the DAL there is a restriction on the maximum length of a smart function.
This change adds amore helpful error message. 

**Related Tasks**: [Error when deploying contract](https://app.asana.com/0/1205770721173531/1206256178382438/f) 

# Manually testing the PR

```bash
jstz deploy examples/long_function.js
```


crates/jstz_cli/src/deploy.rs
```diff
@@ -41,6 +44,9 @@ pub async fn exec(
 
     let code = read_file_or_input_or_piped(code)?
         .ok_or(user_error!("No function code supplied. Please provide a filename or pipe the file contents into stdin."))?;
+    if code.len() > MAX_CODE_LENGTH {
+        bail_user_error!("The data availability layer os not yet available. Smart functions are currently restricted to {MAX_CODE_LENGTH} bytes");
```
sam-finch-tezos, line 48:
```suggestion
        bail_user_error!("The data availability layer is not yet available. Smart functions are currently restricted to {MAX_CODE_LENGTH} bytes");
```

crates/jstz_cli/src/deploy.rs
```diff
@@ -16,6 +16,9 @@ pub async fn exec(
     name: Option<String>,
     network: Option<NetworkName>,
 ) -> Result<()> {
+    // maximum size of code until the DAL is implemented
+    const MAX_CODE_LENGTH: usize = 3915;
```
johnyob, line 20: How was this derived?

Also shouldn't this be a top-level constant?

crates/jstz_cli/src/deploy.rs
```diff
@@ -41,6 +44,9 @@ pub async fn exec(
 
     let code = read_file_or_input_or_piped(code)?
         .ok_or(user_error!("No function code supplied. Please provide a filename or pipe the file contents into stdin."))?;
+    if code.len() > MAX_CODE_LENGTH {
```
johnyob, line 47: `code.len()` doesn't necessarily give the number of bytes, but the number of utf-8 codepoints

sam-finch-tezos: That's worth fixing! Good spot.

crates/jstz_cli/src/deploy.rs
sam-finch-tezos: It was literally derived by trial and error. 
It doesn't actually do anything too bad if this check is omitted. It just has a less helpful error message. ("Failed to post operation")
So this change is only really giving an explanation for a common user error and noting it's due to be fixed.
The DAL stuff would happen in `jstz_node` when the operation is serialised and we can decide whether to send it directly or use the DAL.
For now I wanted to keep that change as superficial and local as possible.

sam-finch-tezos: It's not a top level constant because it has no meaning outside this function. (50:50 decision omp)

johnyob (approved):

alanmarkoTrilitech (approved):
lgtm! the function doesn't seem particularly long to me though - was it really the case before that they had to be shorter than this?

PR #422, opened 2024-02-23
feat(cli): use path for octez binaries and default missing config fields

sam-finch-tezos:
# Context

Currently the octez path is stored as a string which defaults to "" (which is a bit strange).
Missing fields in the config will cause the sandbox to refuse to work.
**Related Tasks**: [Fix jstz configuration defaults](https://app.asana.com/0/1205770721173531/1206345416213107/f)

# Description
* `octez_path` is now stored as an option.
* Config fields equal to their default values are no longer stored and may be omitted making the config less fragile.

# Next Steps
[Add jstz config command.](https://app.asana.com/0/1205770721173531/1206614216286655/f)

### DIFF
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -67,7 +67,9 @@ pub struct SmartFunction {
 
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
 pub struct AccountConfig {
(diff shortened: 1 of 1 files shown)


alanmarkoTrilitech (approved):
LGTM! Can you please remove the merge commit?

PR #426, opened 2024-02-26
fix(node): collapsible match fix

alanmarkoTrilitech:
# Context
Cargo build fails outside of nix shell due to an issue with clippy::collapsible-match

# Description
'-'->'_'.

# Manually testing the PR
Cargo build


ryutamago (changes requested):
i thought i removed this line before... 
I think you can remove this line and make check shuld still pass

alanmarkoTrilitech:
You are right, it seems to work without it.

ryutamago (approved):
LGTM!

PR #437, opened 2024-03-05
fix(core): fix native class constructor method

johnyob:
# Context

**Related Tasks**: [Panics in URL / URLSearchParams](https://app.asana.com/0/1205770721173531/1206682403279323/f)

# Description

This PR fixes the `NativeClass` constructor method, namely the `constructor` doesn't have the `this` value to hand until the `JsNativeObject` is built. 

This was causes dereferencing panics since the *target* was being passed as `this`. The _target_ is the `JsValue` that is used e.g. for `new Animal`, the `JsValue` would be `Animal`. This obviously doesn't have the same Rust type as the `Self::Instance` type. 

The solution is to split the constructor into a *data* constructor and an *object* constructor.

# Manually testing the PR

This panic'd previously
```sh
jstz repl
>> const url = new URL("http://example.com/?");
>> url.searchParams.delete("param1")
>> url.search
null
```

### DIFF
--- a/crates/jstz_api/src/encoding/text_encoder.rs
+++ b/crates/jstz_api/src/encoding/text_encoder.rs
@@ -85,8 +85,8 @@ impl NativeClass for TextEncoderEncodeIntoResult {
 
     const NAME: &'static str = "TextEncoderEncodeIntoResult";
 
-    fn constructor(
-        _this: &JsNativeObject<TextEncoderEncodeIntoResult>,
+    fn data_constructor(
+        _target: &JsValue,
         _args: &[JsValue],
         _: &mut Context<'_>,
     ) -> JsResult<TextEncoderEncodeIntoResult> {
@@ -255,8 +255,8 @@ impl NativeClass for TextEncoderClass {
 
     const NAME: &'static str = "TextEncoder";
 
-    fn constructor(
-        _this: &JsNativeObject<TextEncoder>,
+    fn data_constructor(
+        _target: &JsValue,
         _args: &[JsValue],
         _: &mut Context<'_>,
     ) -> JsResult<TextEncoder> {
--- a/crates/jstz_api/src/url/mod.rs
+++ b/crates/jstz_api/src/url/mod.rs
@@ -71,7 +71,6 @@ impl Url {
     ///
     /// [spec] https://url.spec.whatwg.org/#dom-url-url
     pub fn new(
-        this: &JsNativeObject<Self>,
         url: String,
         base: Option<String>,
         context: &mut Context<'_>,
@@ -95,9 +94,6 @@ impl Url {
                 context,
             )?,
         };
-
-        // 7. Set `url`’s query object’s URL object to `url`.
-        url.search_params.deref_mut().set_url(this);
 
         Ok(url)
     }
@@ -448,15 +444,26 @@ impl NativeClass for UrlClass {
 
     const NAME: &'static str = "URL";
 
-    fn constructor(
-        this: &JsNativeObject<Url>,
+    fn data_constructor(
+        _target: &JsValue,
         args: &[JsValue],
         context: &mut Context<'_>,
     ) -> JsResult<Url> {
         let url: String = args.get_or_undefined(0).try_js_into(context)?;
         let base: Option<String> = args.get_or_undefined(1).try_js_into(context)?;
 
-        Url::new(this, url, base, context)
+        Url::new(url, base, context)
+    }
+
+    fn object_constructor(
+        this: &JsNativeObject<Self::Instance>,
+        _args: &[JsValue],
+        _context: &mut Context<'_>,
+    ) -> JsResult<()> {
+        // 7. Set `this`’s query object’s URL object to `this`.
+        this.deref_mut().search_params.deref_mut().set_url(this);
+
+        Ok(())
     }
 
     fn init(class: &mut ClassBuilder<'_, '_>) -> JsResult<()> {
--- a/crates/jstz_core/src/native.rs
+++ b/crates/jstz_core/src/native.rs
@@ -38,13 +38,6 @@ impl<T: NativeObject> JsNativeObject<T> {
         value.as_object().map_or(false, JsObject::is::<T>)
     }
 
-    fn from_value_unchecked(value: JsValue) -> Self {
-        Self {
-            inner: value,
-            _phantom: PhantomData,
-        }
-    }
-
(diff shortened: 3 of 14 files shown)


crates/jstz_api/src/encoding/text_encoder.rs
```diff
@@ -255,8 +255,8 @@ impl NativeClass for TextEncoderClass {
 
     const NAME: &'static str = "TextEncoder";
 
-    fn constructor(
-        _this: &JsNativeObject<TextEncoder>,
+    fn data_constructor(
+        _this: &JsValue,
```
ryutamago, line 259: isn't this supposed to be _target?

crates/jstz_core/src/native.rs
```diff
@@ -455,13 +447,25 @@ pub trait NativeClass {
     /// The attributes the class will be bound with, default is `writable`, `enumerable`, `configurable`.
     const ATTRIBUTES: Attribute = Attribute::all();
 
-    /// The constructor of the class.
-    fn constructor(
-        this: &JsNativeObject<Self::Instance>,
+    /// Creates the internal data for an instance of this class.
+    fn data_constructor(
+        target: &JsValue,
```
ryutamago, line 452: target is never used in the class that implements it, i'm curous how we would ever need it?

crates/jstz_core/src/native.rs
```diff
@@ -424,14 +417,13 @@ fn raw_constructor<T: NativeClass>(
...
         .flatten();
 
-    let native_object = T::constructor(
-        &JsNativeObject::from_value_unchecked(this.clone()),
-        args,
-        context,
-    )?;
+    let native_object = T::data_constructor(target, args, context)?;
+
     let object =
```
ryutamago, line 422: I find the naming a bit confusing here. What's the differnce between instance and native object? 

the variable `object` is of type `JsNativeObject` which is a type safe version of `native_object`?

ryutamago (changes requested):

crates/jstz_core/src/native.rs
johnyob: `native_object` here refers to the native Rust value

johnyob: `object` is a JavaScript object (wrapped using the `JsNativeObject` struct -- which gives some more type safety)

crates/jstz_core/src/native.rs
johnyob: If you used static properties of the target in the initializer

crates/jstz_api/src/encoding/text_encoder.rs
johnyob: Yes, it should be 👍

johnyob: Fixed

crates/jstz_core/src/native.rs
```diff
@@ -455,13 +447,25 @@ pub trait NativeClass {
...
+    fn data_constructor(
+        target: &JsValue,
         args: &[JsValue],
         context: &mut Context<'_>,
     ) -> JsResult<Self::Instance>;
 
+    /// Initializes the properties of the constructed object for an instance of this class.
+    ///
+    /// Useful to initial additional properties for the construced object that aren't part of the
+    /// internal data are rely on the arguments or this.
```
alanmarkoTrilitech, line 460: this comment could be improved

johnyob: In what sense? What do you not understand from the comment?

alanmarkoTrilitech: The whole sentence doesn't make much sense with the "data are rely".
There is also a grammar issue - construced->constructed.
Overall it is also slightly confusing and would benefit from simplification clearly separating what is meant for rust and what for Javascript and why we need two constructors, possibly with a small example of how the are used.

alanmarkoTrilitech (commented):
the changes seem reasonable, however there are now three differen constructor types - raw, object and data. It would be good to clearly explain their differences. Also do we always need all of them? isn't it that the native class interface could be simplified to just one constructor, and the class itself can internally solve the remaining ones?

crates/jstz_api/src/url/mod.rs
```diff
@@ -448,15 +444,26 @@ impl NativeClass for UrlClass {
...
+        Url::new(url, base, context)
+    }
+
+    fn object_constructor(
+        this: &JsNativeObject<Self::Instance>,
+        _args: &[JsValue],
+        _context: &mut Context<'_>,
+    ) -> JsResult<()> {
+        // 7. Set `this`’s query object’s URL object to `this`.
+        this.deref_mut().search_params.deref_mut().set_url(this);
```
alanmarkoTrilitech, line 464: is this really supposed to create a cycle?

crates/jstz_core/src/native.rs
johnyob: Fixed typos -- linked example usage in `Url::object_constructor`.

johnyob: We have 3 kinds of constructors:
- The `raw` constructor
- The `data` constructor
- The `object` constructor

The `raw` constructor is the `NativeFunction` that is called when we want to construct an object e.g. the function called when evaluating `new URL(...)`. 

The `data` constructor is the constructor for creating the Rust value (of type `T`) that is required to construct the `JsNativeObject<T>` -- the JavaScript representation of `T`. 

The `object` constructor is the constructor that has access to `this` and `args`; it is called after the `data` constructor since it requires `this : &JsNativeObject<T>`. This constructor is useful for setting properties that rely on `this` (e.g. the required cycle between `URL` and `URLSearchParams`)

johnyob:
> isn't it that the native class interface could be simplified to just one constructor, and the class itself can internally solve the implementation of the remaining ones?

No -- since Rust doesn't easily permit cycles. And to achieve a single constructor, we'd need to permit somehow mutating the `JsNativeObject<T>` from a default null-like value to the native object. 

We could have `T: Default`, but in most cases, our objects don't implement this. Additionally we cannot mutate a `JsObject` from a null-like object to a `NativeObject`. 

So two constructors is the best approach -- the constructor for constructing the Rust value (`data` constructor) and a constructor for using/mutating `this` (perhaps based on `args`)

johnyob:
The `raw` constructor is never exposed in this API and is merely an implementation detail for bridging our trait to the JS world

ryutamago (approved):
🔥

PR #438, opened 2024-03-06
fix(cli): bridge network + errors fix

alanmarkoTrilitech:
# Context

https://app.asana.com/0/1205770721173533/1206712613344550
https://app.asana.com/0/1205770721173533/1206712613344544
https://app.asana.com/0/1205770721173533/1206703424903095

# Description
These PR fixes several issues with jstz bridge - requiring -n dev flag,supporting bootstrap aliases, being able to deposit to a bootstrap account. It additionally improves error messages and does a small refactoring.

# Manually testing the PR

```
jstz bridge depost --from bootstrap1 --to alan --amount 1000000
```

### DIFF
--- a/crates/jstz_cli/src/bridge/deposit.rs
+++ b/crates/jstz_cli/src/bridge/deposit.rs
@@ -2,7 +2,8 @@ use log::{debug, info};
 
 use crate::{
     config::{Config, NetworkName},
-    error::Result,
+    error::{bail_user_error, Result},
+    sandbox::SANDBOX_BOOTSTRAP_ACCOUNTS,
     utils::AddressOrAlias,
 };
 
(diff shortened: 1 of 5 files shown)


crates/jstz_cli/src/bridge/deposit.rs
```diff
@@ -14,18 +16,52 @@ pub fn exec(
...
+    // Check if trying to deposit to a bootsrap account.
+    if let Some(bootstrap_account) = SANDBOX_BOOTSTRAP_ACCOUNTS
+        .iter()
+        .find(|account| account.address == to_pkh.to_string())
+    {
+        bail_user_error!(
+            "Cannot deposit to the bootstrap account '{}'.",
+            bootstrap_account.address
+        );
+    }
```
johnyob, line 39: Not too sure this is required

johnyob (commented):
Added a review commit, changes lgtm mostly.

crates/jstz_cli/src/bridge/deposit.rs
alanmarkoTrilitech: I think it is - it clearly shouldn't be possible to bridge to a bootstrap account. 

We might want to have something where the bridge contract declines it automatically. Anyway, the bridge will be reworked quite a lot in the future - e.g. added withdrawals, tokens etc.

alanmarkoTrilitech (commented):
The review commit looks good to me. YOu have to approve the PR though.

johnyob:
> I think it is - it clearly shouldn't be possible to bridge to a bootstrap account.

These are different kinds of accounts. Bridging to a bootstrap account might be desirable -- keeping the same bootstrap account addresses across Tezos and jstz. 

> We might want to have something where the bridge contract declines it automatically. Anyway, the bridge will be reworked quite a lot in the future - e.g. added withdrawals, tokens etc.

Agreed, I won't block this based on this.

johnyob (approved):
LGTM 🚢  -- please squash the review commit before merging 🚀

alanmarkoTrilitech:
> > I think it is - it clearly shouldn't be possible to bridge to a bootstrap account.
> 
> These are different kinds of accounts. Bridging to a bootstrap account might be desirable -- keeping the same bootstrap account addresses across Tezos and jstz.
> 
> > We might want to have something where the bridge contract declines it automatically. Anyway, the bridge will be reworked quite a lot in the future - e.g. added withdrawals, tokens etc.
> 
> Agreed, I won't block this based on this.

thanks for not blocking it, yeah it seems like a design decision that should require a bit of thought - for example it might be also useful to have L1/L2 addresses all disjoint

PR #452, opened 2024-03-11
docs: add instructions for cloning get-tez example and installing dependencies

johnyob:
# Context

Currently following the quick start guide command-for-command doesn't work (you get an npm error for
compiling the get-tez smart function)

# Description

Changes the quick start guide to:
- Instruct the user to clone the `jstz` repository
- Install npm dependencies for deploying the get-tez smart function

Additionally I removed the `-d` for the sandbox start since this flag doesn't work for the docker installation. 

# Manually testing the PR

```sh
npm run docs:dev
```
and following the quick start instructions

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -2,7 +2,7 @@
 
 This guide will instruct you in writing and deploying your first _smart function_ in under 10 minutes.
 
-It assumes that you have already [installed `jstz`](installation.md) and have a basic familiarity with [JavaScript](https://www.youtube.com/watch?v=lkIFF4maKMU).
+It assumes that you have already [installed `jstz`](installation.md) and have a basic familiarity with [JavaScript](https://www.youtube.com/watch?v=lkIFF4maKMU) and have `npm` (`>= 9.6.7`) installed.
 
 ## What is jstz?
 
@@ -11,7 +11,13 @@ It assumes that you have already [installed `jstz`](installation.md) and have a 
 With `jstz` you can deploy so called _smart functions_ which are operating similarly to cloud functions, while running on Tezos L2 and
 providing additional security and blockchain-specific functionality typical for smart contracts.
 
-## 1) Your First Smart Function
+## 1. Your First Smart Function
+
+First we will clone the `jstz` repository and navigate to the `get-tez` example:
+
+```sh
+git clone https://github.com/trilitech/jstz.git && cd jstz/examples/get-tez
+```
 
 In this example, the smart function provides a way to send a tez to the requester if asked politely.
 It takes a HTTP `Request` object with a message and returns a `Response` object informing whether the request succeeded.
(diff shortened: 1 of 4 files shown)


docs/quick_start.md
```diff
@@ -115,28 +121,62 @@ In addition to several [standard Web APIs](./api/index.md#web-platform-apis), `j
   Smart functions can invoke other smart functions using `fetch`, similiar to network requests in JavaScript.
   Additionally, new smart functions can be deployed by a smart function using the [`SmartFunction`](./api/smart_function.md) API.
 
-## 2) Deploying your Smart Function
+## 2. Deploying your Smart Function
 
 First we must compile our TypeScript code to JavaScript using:
 
 ```sh
+npm install
```
alanmarkoTrilitech, line 129: this should be part of the installation guide

package.json
```diff
@@ -7,7 +7,8 @@
   "workspaces": [
     "packages/*",
     "examples/fa2",
-    "examples/hello-world"
+    "examples/hello-world",
+    "examples/get-tez"
```
alanmarkoTrilitech, line 11: isn't it better to have access to the whole examples folder, including js ones?

docs/quick_start.md
```diff
@@ -115,28 +121,62 @@ In addition to several [standard Web APIs](./api/index.md#web-platform-apis), `j
...
 npm run build
 ```
 
 Once built, we can deploy the smart function to the local sandbox:
 
 ```sh
-jstz sandbox start -d
+jstz sandbox start
 jstz deploy dist/index.js
```
alanmarkoTrilitech, line 137: how about having npm run build right next to deploy and sandbox start separately? the sandbox start is run just once,  however each time you make a change to the  ts file you need to recompile it. Also, it would be good to write somewhere what is the ts file so that it is encouraged to explore it and change.

alanmarkoTrilitech (changes requested):

docs/quick_start.md
```diff
@@ -11,7 +11,13 @@ It assumes that you have already [installed `jstz`](installation.md) and have a
 With `jstz` you can deploy so called _smart functions_ which are operating similarly to cloud functions, while running on Tezos L2 and
 providing additional security and blockchain-specific functionality typical for smart contracts.
 
-## 1) Your First Smart Function
+## 1. Your First Smart Function
+
```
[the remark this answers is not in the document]
johnyob: I disagree, it should not be part of the installer. If we had the time, it would be part of the jstz CLI to have a `template` command -- but for now, manually cloning doesn't hurt

docs/quick_start.md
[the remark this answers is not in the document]
johnyob: Agreed

docs/quick_start.md
johnyob: Disagree, see above

package.json
johnyob: Not really, this is a set of globs that contain folders with `package.json`s. The JS examples don't have these.

docs/quick_start.md
johnyob: > how about having npm run build right next to deploy and sandbox start separately? the sandbox start is run just once, however each time you make a change to the ts file you need to recompile it. 

Sure

ryutamago (approved):
LGTM once @alanmarkoTrilitech's point is addressed. Great work 🙌

alanmarkoTrilitech (approved):
LGTM! 🚢

PR #453, opened 2024-03-11
feat: installer configures weeklynet

johnyob:
# Context

# Description

Adds weeklynet network config to the installer script

# Manually testing the PR

Having deleted / moved your `~/.jstz/config.json`, run:
```
source scripts/install-jstz-cli.sh
```

### DIFF
--- a/scripts/install-jstz-cli.sh
+++ b/scripts/install-jstz-cli.sh
@@ -3,6 +3,7 @@
 { # this ensures the entire script is downloaded
 
 version="v0.1.0-alpha.0"
+network="weeklynet-2024-03-06"
 container="ghcr.io/trilitech/jstz-cli:$version"
 jstz_home="$HOME/.jstz"
 
@@ -38,7 +39,13 @@ jstz_configure() {
         mkdir -p "$jstz_home"
         cat >"$jstz_home/config.json" << EOF
 {
-  "default_network": "dev"
+  "default_network": "dev",
+  "networks": {
+    "weeklynet": {
+      "octez_node_rpc_endpoint": "http://rpc.$network.teztnets.com",
+      "jstz_node_endpoint": "http://34.39.12.211:8933"
+    }
+  }
 }
 EOF
 
--- a/scripts/install-jstz-cli.sh
+++ b/scripts/install-jstz-cli.sh
@@ -66,7 +66,7 @@ EOF
             cat 1>&2 << EOF
 Unsupported shell: $shell. 
 Please manually add the alias to your shell's configuration file.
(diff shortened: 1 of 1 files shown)


ryutamago (changes requested):
shouldn't we check for the full path? 
```
if [ -d "$jstz_home/config.json" ]; then
        echo "Configuration already exists. Skipping..."
```

scripts/install-jstz-cli.sh
```diff
@@ -38,7 +39,13 @@ jstz_configure() {
...
-  "default_network": "dev"
+  "default_network": "dev",
+  "networks": {
+    "weeklynet": {
+      "octez_node_rpc_endpoint": "http://rpc.$network.teztnets.com",
+      "jstz_node_endpoint": "http://34.39.12.211:8933"
+    }
+  }
 }
 EOF
```
alanmarkoTrilitech, line 51: where is the JSTZ_VERSION defined?

alanmarkoTrilitech, line 51: The manual alias and shell alias differ

alanmarkoTrilitech (changes requested):
The network change seems good.

Added some comments regarding other changes in the install file as I couldn't do it before (please next time don't merge the review commits automatically).

I think it would be nice if the install script also checked the npm, and encouraged the user to install it if it isnt + also if it automatically cloned the examples folder that we currently have in the quick start guide(I don't think there is a need to clone the whole repo, just the folder).  

I've also had an issue with having old alias in .zshrc - not sure if it was because of JSTZ_VERSION being wrong or just having old version of it, but I guess it is fine for the first time user.

scripts/install-jstz-cli.sh
johnyob: It was moved to the `$version` variable -- it is not standard for all bash variables to be caps (just non-local variables)

johnyob: Will fix now

johnyob:
> shouldn't we check for the full path?

What do you mean by the 'full path'?

johnyob:
> I think it would be nice if the install script also checked the npm, and encouraged the user to install it if it isnt + also if it automatically cloned the examples folder that we currently have in the quick start guide(I don't think there is a need to clone the whole repo, just the folder).

I disagree. This is not the responsibility of the installer, but likely of a `template` command in the CLI

scripts/install-jstz-cli.sh
johnyob: Fixed 9169132759c687f3ca26de7139ffdc820c3ea119

ryutamago:
> > shouldn't we check for the full path?
> 
> What do you mean by the 'full path'?

i meant instead of checking if `.jstz` exists, we should check if `.jstz/config.json`exists.

alanmarkoTrilitech (approved):
:ship:

ryutamago (approved):
🔥

PR #461, opened 2024-03-12
feat(scripts): upgrading config

alanmarkoTrilitech:
Please go the the `Preview` tab and select the appropriate sub-template:

- [🚧 Web/runtime APIs template](?expand=1&template=api_dev.md)
- [📚 Web/runtime APIs documentation template](?expand=1&template=api_doc.md)
- [⬜️ Blank template](?expand=1&template=default.md)


scripts/install-jstz-cli.sh
```diff
@@ -31,13 +31,11 @@ jstz_download() {
...
-        echo "Configuration already exists. Skipping..."
-    else
-        echo -n "Configuring jstz..."
+    echo -n "Configuring jstz..."
 
-        mkdir -p "$jstz_home"
-        cat >"$jstz_home/config.json" << EOF
+    mkdir -p "$jstz_home"
+    
+    cat >"$jstz_home/config.json" << EOF
```
johnyob, line 38: This overwrites the config for each installation -- I think we should provide a way for the user to keep their config and just update the weeklynet endpoint (using sed).

...Though now I think about it, the config will essentially be useless after the update 🤔

alanmarkoTrilitech:
Updated so that it makes a backup copy only if the config file differs (prevents running the script twice in a row and losing the config.)

johnyob (approved):
lgtm

PR #470, opened 2024-03-14
feat: support for using docker for Octez

alanmarkoTrilitech:
# Context

This PR adds support for using Docker to run the octez-client, octez-node and octez-smart-rollup-node. 

# Description

The changes mainly focus around the introduction of `OctezSetup`, an enum that either allows one to use an installed version of octez or a docker image (e.g. `tezos/tezos:v20.0-rc1`). The enum has a method called `command` which will either return a command for the installed binary or a docker command that will start a container running the binary. 

# Manually Testing

With:
```
{
  "octez_setup": {
    "Docker": "v20.0-rc1"
  }
 ...
}
```

Run
```
jstz sandbox start
```

### DIFF
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -9,7 +9,7 @@ use derive_more::{From, TryInto};
 use jstz_crypto::{public_key::PublicKey, secret_key::SecretKey};
 use jstz_proto::context::account::Address;
 use log::debug;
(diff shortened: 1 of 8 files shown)


alanmarkoTrilitech:
This mostly looks good to me, but it would be good to add a more detailed description of changes so that it is a bit easier to review and polish it a bit(removing comments). I need to test it locally before approving.

Issue #473, opened 2024-03-14
🐛 Bug: jstz installer: installer attempts to reload the `$SHELL` config, even when the current shell isn't `$SHELL`


emturner:
# Description

Admittedly, I have a non-standard setup, but I can't imagine I'm the only user to do this. My login shell is bash, but the shell I'm almost always using (outside of nix-shells etc) is zsh.

### Observed behaviour

The installer attempts to write the alias to `.bashrc`, despite me running it from a zsh session. I have no issues with this, and is understandable behaviour.

Once finished, however, the installer reloads my `.bashrc`, which, among other things, breaks my zsh prompt.

### Expected behaviour

The installer should only reload the shell's configuration file if `$SHELL` matches the currently running shell.

### Steps to reproduce

From a zsh shell:

```
λ export SHELL=$(which bash)
λ source <(curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/trilitech/jstz/main/scripts/install-jstz-cli.sh)
Pulling jstz-cli docker container from GHCR...
20240313: Pulling from trilitech/jstz-cli
Digest: sha256:f320f5a2247acb08190db13429d600eef45ad29c03a76fbf47685183ecb1ef34
Status: Image is up to date for ghcr.io/trilitech/jstz-cli:20240313
ghcr.io/trilitech/jstz-cli:20240313

Configuring jstz...Configuration unchanged. No update needed.
Configuring `jstz` alias...
jstz_configure:64: permission denied: /home/emma/.bashrc
Alias added to /home/emma/.bashrc.
\u@\h \w${GUIX_ENVIRONMENT:+ [env]}\$
```

emturner:
fixed by #475

Issue #472, opened 2024-03-14
🐛 Bug: jstz install: installer does not respect `$ZDOTDIR`


emturner:
# Description

`jstz` installer writes to `~/.zshrc`, regardless of the actual value of `$ZDOTDIR`.

### Observed behaviour

The installer assumes the `.zshrc` is always placed at `$HOME/.zshrc`, which is incorrect. While it is the default, if `$ZDOTDIR` is set, the configuration is placed at `$ZDOTDIR/.zshrc`.

### Expected behaviour

The installer attempts to write the alias to `$ZDOTDIR/.zshrc`.

### Steps to reproduce

From within zsh:

```
export ZDOTDIR=~/.config/zsh && mkdir -p ~/.config/zsh
source <(curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/trilitech/jstz/main/scripts/install-jstz-cli.sh)
```

emturner:
fixed by #475

PR #475, opened 2024-03-14
fix(scripts): zdotdir + config file loading

alanmarkoTrilitech:
# Context

# Description
This PR addresses issue #472 and #473. It respects ZDOTDIR variable if set and only reloads the configuration file if the shell is matching SHELL value, in order to not break previous prompt.

# Manually testing the PR
run the install script

### DIFF
--- a/scripts/install-jstz-cli.sh
+++ b/scripts/install-jstz-cli.sh
@@ -7,6 +7,10 @@ network="weeklynet-2024-03-13"
 container="ghcr.io/trilitech/jstz-cli:$version"
 jstz_home="$HOME/.jstz"
 
+# ps -p filters on the parent process, -o comm= prints the command name (supressing the header)
+current_shell=$(basename "$(ps -p $$ -o comm=)")
+shell=$(basename "$SHELL")
+
 # (--rm): remove the container after it exits
 # (-v): the container mounts the following volumes:
 #  - /tmp for temporary files
@@ -21,7 +25,7 @@ jstz_download() {
     # Check if Docker is installed
     if ! command -v docker &> /dev/null; then
         echo "Docker is not installed. Please install Docker and try again."
-        exit 1
+        return 1
     fi
 
     # Pull the Docker CLI container from GitHub Container Registry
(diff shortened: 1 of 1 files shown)


scripts/install-jstz-cli.sh
```diff
@@ -72,22 +70,30 @@ EOF
...
-            ;;
-        *)
-            cat 1>&2 << EOF
-Unsupported shell: $shell. 
-Please manually add the alias to your shell's configuration file.
-    $shell_alias
-EOF
-            exit 1
-    esac
+    current_shell=$(ps -p $$ -ocomm=)
```
johnyob, line 73: Wouldn't it be cleaner to have:
```suggestion
    current_shell=$(basename "$0")
```

scripts/install-jstz-cli.sh
```diff
@@ -72,22 +70,30 @@ EOF
...
+
+    echo "SHELL: $SHELL"
+    echo "Detected shell: $current_shell"
+
+    if [[ $SHELL == */bash ]]; then
+        shellrc="$HOME/.bashrc"
+    elif [[ $SHELL == */zsh ]]; then
```
johnyob, line 80: Again, clearer to use `$(basename $SHELL)` and store it in a variable for these matches

scripts/install-jstz-cli.sh
```diff
@@ -72,22 +70,30 @@ EOF
...
+    fi
+
+    if ! [ -w "$shellrc" ]; then
+        echo "Warning: $shellrc is not writable. Please manually add the following alias to your shell's configuration file:"
+        echo "$shell_alias"
+
+        echo "Once you have added the alias, run the following command to reload the configuration file in your default shell:"
+        echo "    source $shellrc"
+        exit 1  # Exit the script with an error status
+    fi
```
johnyob, line 96: Rebase artifact?

johnyob (changes requested):

scripts/install-jstz-cli.sh
alanmarkoTrilitech: $0 points to install-jstz-cli.sh

johnyob: When you source? or run as a script?

alanmarkoTrilitech: with source it results in jstz_configure

scripts/install-jstz-cli.sh
alanmarkoTrilitech: yeah, should have been part of the previous PR

PR #490, opened 2024-04-17
Sandbox: improve startup robustness

emturner:
# Context

Prevents a few different cryptic errors when starting up (all of which appear as 'file not found'), namely:

- octez binaries can't be found, despite having been copied into `jstz` directory - they have to be in `PATH`
- kernel.log file not created, which causes the `TailedFile::init` to fall over. Ensure we create it earlier in the process

# Description

# Manually testing the PR

delete your `~/.jstz` folder, and try starting the sandbox with/without this change


johnyob (approved):
lgtm -- tested locally

PR #495, opened 2024-05-09
feat(rollup): private rollup

alanmarkoTrilitech:
# Context
https://app.asana.com/0/1205770721173533/1207097745265939
https://app.asana.com/0/1205770721173533/1207097745265941

# Description
This PR makes it possible to use whitelist for deployment of private rollups. It also changes some log outputs.
Additionally a new runbook was created for deploying a private rollup - https://www.notion.so/trilitech/Runbook-private-rollup-635a2b66ff5048ba9b37967275d1bf53. 

# Manually testing the PR
Use the runbook to run the rollup.


.env.example
johnyob, line 1: I wouldn't update this really with the exception of the changes to the `NETWORK` env var

crates/octez/src/client.rs
```diff
@@ -279,6 +279,8 @@ impl OctezClient {
             "--burn-cap",
             "999",
             "--force",
+            "--whitelist",
+            format!("[\"{}\"]", account).as_str(),
```
johnyob, line 283: Does this affect the local sandbox?

alanmarkoTrilitech: good point, we really need some tests :D I've tried before starting the sandbox which is ok, but apparently deploying a contract causes Timeout waiting for operation receipt. I'll look into it

alanmarkoTrilitech: it doesn't actually affect it.

johnyob (approved):
lgtm

alanmarkoTrilitech:
> lgtm -- need to test locally before merging though

ok, feel free to click on the merge button once you have done so.

PR #496, opened 2024-05-14
fix(deps): remove blst from crate tree

emturner:
# Context

Remove blst from crate graph

# Description

Prevent `blst` coming back into the crate tree, as this breaks building jstz for riscv on mac.

# Manually testing the PR

`cargo tree | grep blst`

See https://github.com/trilitech/jstz/actions/runs/9080566889/job/24952362135 for an example of the new check failing


johnyob (approved):
lgtm 🚀 🚢

PR #502, opened 2024-06-03
feat(kernel): hide 'static requirements

emturner:
# What

Hide `'static` requirements from top-level entry function

# Why

Significantly simplifies integration of jstz with parts of the sdk/riscv work. The requirement is used by jstz to play fast & loose with lifetimes internally, and is effectively an implementation detail which can be hidden away.


johnyob (commented):
lgtm 🚀 🚢

johnyob (approved):

PR #520, opened 2024-07-12
fix(sandbox): advertise octez-node rpc port instead of network port

zcabter:
# Context
Fixes the advertised octez node rpc endpoint to point to the correct RPC port

# Description 
`--net-addr` arg configures the [p2p endpoint of the node.](https://tezos.gitlab.io/user/node-configuration.html#listening-ports) whereas `--rpc-addr`[ configures the RPC endpoint](https://tezos.gitlab.io/developer/rpc.html#activating-rpc).  I think the intent was to advertise the rpc endpoint.

# Manually Testing the PR
When running sandbox, jstz will advertise the rpc endpoint to be  `http://localhost:18730` insteand of ` http://localhost:18731`. 

`./<octez-client> -E http://localhost:18730 rpc list` should work

### DIFF
--- a/crates/jstz_cli/src/sandbox/daemon.rs
+++ b/crates/jstz_cli/src/sandbox/daemon.rs
@@ -52,7 +52,7 @@ use super::{
 fn octez_node_endpoint() -> String {
     format!(
         "http://{}:{}",
-        SANDBOX_LOCAL_HOST_ADDR, SANDBOX_OCTEZ_NODE_PORT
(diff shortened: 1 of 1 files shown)


alanmarkoTrilitech:
can you please add some more context to the PR regarding why this fix is needed and what exactly does it change so that I can test it?

zcabter:
@alanmarkoTrilitech Done

alanmarkoTrilitech (approved):
LGTM! 🚢  🚀 

btw. can you please use assignees next time so that we know who should be taking care of the PR

PR #525, opened 2024-07-19
fix(octez): properly drop octez threads when done

zcabter:
# Context
This PR kills zombie octez processes started by the sandbox if the sandbox errors out or is killed before it reaches the main' `handle.join()`

# Description
Introduces a `Sandbox` struct that is responsible for performing the clean up of octez threads and reverting the config file in the event of error.  FWIW, there is definitely more that can be done to come up with a better solution but I have restrained myself from refactoring everything.

# Manually testing the PR
1. Apply the following patch to `main` branch to introduce a panic after the octez processes have started. 
```rust
diff --git a/crates/jstz_cli/src/sandbox/daemon.rs b/crates/jstz_cli/src/sandbox/daemon.rs
index 2962a9d..a9dc52b 100644
--- a/crates/jstz_cli/src/sandbox/daemon.rs
+++ b/crates/jstz_cli/src/sandbox/daemon.rs
@@ -573,6 +573,7 @@ pub async fn run_sandbox(cfg: &mut Config) -> Result<()> {
     debug!(log_file, "Saving sandbox config");
     cfg.save()?;
     // 4. Wait for the sandbox or jstz-node to shutdown (either by the user or by an error)
+    panic!();
     run_jstz_node(cfg).await?;
     sandbox.join()?;
     Ok(())
```

2. Observe that octez processes have been killed
`ps -ae | grep octez`
3. `cat ~/.jstz/config.json` show no remnants of sandbox configurations

### DIFF
--- a/crates/jstz_cli/src/sandbox/daemon.rs
+++ b/crates/jstz_cli/src/sandbox/daemon.rs
@@ -11,11 +11,13 @@ use octez::OctezThread;
 use regex::Regex;
 use std::io::Write;
 use std::{
+    cell::RefCell,
     env,
     fs::{self, File, OpenOptions},
     io::{BufRead, BufReader, Seek},
     path::PathBuf,
     process::{Child, Command, Stdio},
+    rc::Rc,
     thread::{self, sleep},
     time::Duration,
 };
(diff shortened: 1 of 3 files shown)


johnyob (approved):
lgtm 🚀

PR #527, opened 2024-07-23
fix(sandbox): fix sandbox not shutting down on ctrl+c

zcabter:
# Context

Fix bug introduced in the https://github.com/jstz-dev/jstz/pull/525 . The octez threads were not shutting down. When jstz node is shutdown, we should terminate the sandbox.

Additionally, this PR ensures that all processes are killed when any one process is killed.

# Description

# Manually testing the PR
1. Start sandbox then kill it after it is fully loaded.
2. Start sanbox then kill octez or jstz processes via `kill <pid>`

### DIFF
--- a/crates/jstz_cli/src/sandbox/daemon.rs
+++ b/crates/jstz_cli/src/sandbox/daemon.rs
@@ -1,3 +1,4 @@
+use futures::Future;
 use indicatif::{ProgressBar, ProgressStyle};
 use jstz_rollup::{rollup::make_installer, Exchanger, JstzRollup, NativeBridge};
 use nix::{
@@ -6,6 +7,10 @@ use nix::{
 };
 use octez::OctezThread;
 use regex::Regex;
+use signal_hook::{
+    consts::{SIGINT, SIGTERM},
+    iterator::Signals,
+};
 use std::io::Write;
 use std::{
     cell::RefCell,
@@ -24,7 +29,7 @@ use in_container::in_container;
 use log::info;
 use prettytable::{format::consts::FORMAT_DEFAULT, Cell, Row, Table};
 use tempfile::TempDir;
-use tokio::task;
+use tokio::task::{self, JoinHandle};
 
 macro_rules! debug {
     ($file:expr, $($arg:tt)*) => {
@@ -103,6 +108,8 @@ pub struct Sandbox {
     octez_node: Option<OctezThread>,
     octez_baker: Option<OctezThread>,
     rollup_node: Option<OctezThread>,
+    jstz_node: Option<JoinHandle<Result<()>>>,
+    task_set: task::LocalSet,
 }
 
 impl Sandbox {
@@ -113,10 +120,12 @@ impl Sandbox {
             octez_node: None,
             octez_baker: None,
             rollup_node: None,
+            jstz_node: None,
+            task_set: task::LocalSet::new(),
         }
     }
 
-    fn shutdown(&mut self) -> Result<()> {
+    pub fn shutdown(&mut self) -> Result<()> {
         if let Some(mut t) = self.rollup_node.take() {
             t.shutdown()?;
         }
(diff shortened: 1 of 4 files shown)


crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -574,7 +557,7 @@ pub async fn run_sandbox(cfg: &mut Config) -> Result<()> {
     cfg.save()?;
     // 4. Wait for the sandbox or jstz-node to shutdown (either by the user or by an error)
     run_jstz_node(cfg).await?;
-    sandbox.join()?;
+    sandbox.shutdown()?;
```
johnyob, line 560: I think this change introduces an issue where if the rollup node / baker or node dies, then the sandbox won't die either -- which was the original intent around `OctezThread::join`

johnyob (changes requested):

crates/jstz_cli/src/sandbox/daemon.rs
zcabter: Hmm, I'm not sure that even changes before https://github.com/jstz-dev/jstz/pull/525 did that. 
Anyway, I've updated `join` such that it kills all the processes whenever one is killed/dies. To do that, I had make Sandbox have its own local task set that manages both the jstz node and the culling of processes concurrently.

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -197,6 +270,24 @@ impl Sandbox {
...
         }
     }
+
+    pub fn set_jstz_node(
+        &mut self,
+        jstz_node: impl Future<Output = Result<()>> + 'static,
+    ) -> Result<()> {
+        match self.jstz_node {
+            Some(_) => {
+                debug!(self.log_file.borrow_mut(), "Error: Rollup node already set");
```
johnyob, line 280:
```suggestion
                debug!(self.log_file.borrow_mut(), "Error: jstz node already set");
```

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -131,32 +140,96 @@ impl Sandbox {
...
+        // Loop to notify when either the octez binaries or jstz node is not running
+        // or if SIGINT/SIGTERM was received. This ensures that all processes are
+        // shutdown when any process is shutdown.
+        self.task_set
+            .run_until(async move {
+                task::spawn_local(async move {
+                    'main_loop: loop {
+                        for thread in threads.iter() {
+                            if !thread.is_running() {
+                                println!(
```
johnyob, line 181: We tend to prefer the `info!` macro over `println!` (although it isn't correctly configured to log to stdout atm, but stderr instead 😔 )

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -537,7 +617,7 @@ fn format_sandbox_bootstrap_accounts() -> Table {
     table
 }
 
-pub async fn run_sandbox(cfg: &mut Config) -> Result<()> {
+pub async fn run_sandbox(cfg: &'static mut Config) -> Result<()> {
```
johnyob, line 620: I assume it needs to be static as the `Future` you pass to `set_jstz_node` is static and `cfg` is captured when creating the `Future` returned by `spawn_jstz_node`. 

But `cfg` isn't actually used apart from getting paths (which are owned). I suspect we could write `spawn_jstz_node` like:
```rust
fn spawn_jstz_node(cfg: &Config) -> impl Future<Output=Result<()>> + 'static {
  let log_path = sandbox_daemon_log_path(cfg);
  let kernel_log_path = cfg.sandbox_logs_dir().join("kernel.log");
  let mut log_file = OpenOptions::new()
        .create(true)
        .write(true)
        .append(true)
        .open(log_path.clone())?;
  
  jstz_node::run(SANDBOX_LOCAL_HOST_ADDR,
        SANDBOX_JSTZ_NODE_PORT,
        &format!(
            "http://{}:{}",
            SANDBOX_LOCAL_HOST_ADDR, SANDBOX_OCTEZ_SMART_ROLLUP_PORT
        ),
        &kernel_log_path,)
}

File: crates/jstz_cli/src/sandbox/mod.rs [hunk 5]
```diff
@@ -31,9 +31,9 @@ pub enum Command {
 }
 
 pub async fn start(detach: bool, background: bool) -> Result<()> {
-    let mut cfg = Config::load()?;
+    let cfg: &'static mut Config = Box::leak(Box::new(Config::load()?));
```
johnyob, line 34: Definitely a code smell that we're leaking memory 😅

crates/jstz_cli/src/sandbox/daemon.rs
zcabter: Changed.

crates/jstz_cli/src/sandbox/daemon.rs
zcabter: Done

crates/jstz_cli/src/sandbox/mod.rs
zcabter: 😆

johnyob (approved):
lgtm 🚀

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -735,6 +814,7 @@ pub fn stop_sandbox(restart: bool) -> Result<()> {
 }
 
 pub async fn main(detach: bool, background: bool, cfg: &mut Config) -> Result<()> {
+    let cfg: &mut Config = cfg;
```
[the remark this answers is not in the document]
zcabter: Removed 😅

zcabter:
Waiting for CI to be greento merge.

PR #530, opened 2024-08-01
feat(fa-deposit): handle fa deposit execution in proto

zcabter:
# Context
Handles the execution of an `FaDeposit` operation within the protocol. This PR does not "plug" the execution into the kernel yet but the tests should provide enough confidence of its correctness,. The plugging will be done in a follow up PR.

# Description
The `execute` handler takes an `FaDeposit` as input and produces a `Receipt` as output where the content type is `FaDepositReceiptContent`.  

When executing an FA deposit, the deposit could either be routed to user account or a proxy smart function. 
1. If the former, tickets will be added directly to the user account. 
2. If the latter, the proxy smart function is first executed, targeting the endpoint `POST /-/deposit` with body `{  receiver: ..., amount: ... , ticketHash: ... }` as described in the [Ticket Transport API](https://gitlab.com/baking-bad/tzip/-/blob/wip/029-etherlink-token-bridge/drafts/current/draft-etherlink-token-bridge/etherlink-token-bridge.md#ticket-transport-api) where the proxy should, in turn, mint the respective amount of fa tokens into the user's account. If the proxy succeeds, the kernel deposits the tickets into the proxy function's address. If it fails, tickets are deposited into the receiver's address. It will be the user's responsibility to submit a withdrawal operation of failed deposits.

# Manually testing the PR
`cargo test -p jstz_proto --lib fa_deposit`

### DIFF
--- a/crates/jstz_proto/src/executor/fa_deposit.rs
+++ b/crates/jstz_proto/src/executor/fa_deposit.rs
@@ -1,0 +1,418 @@
+use crate::{
+    context::{account::Amount, ticket_table::TicketTable},
+    executor::smart_function,
+    operation::{external::FaDeposit, RunFunction},
+    receipt::Receipt,
+    Result,
+};
+use derive_more::{Display, Error, From};
+use http::{header::CONTENT_TYPE, HeaderMap, Method, Uri};
+use jstz_api::http::body::HttpBody;
+use jstz_core::{host::HostRuntime, kv::Transaction};
+use jstz_crypto::public_key_hash::PublicKeyHash;
+use serde::{Deserialize, Serialize};
+use tezos_smart_rollup::{michelson::ticket::TicketHash, prelude::debug_msg};
+
+const FA_DEPOSIT_GAS_LIMIT: usize = usize::MAX;
+
+// TODO: https://linear.app/tezos/issue/JSTZ-36/use-cryptos-from-tezos-crypto
+// Properly represent the null address
+const NULL_ADDRESS: &str = "tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx";
+const DEPOSIT_URI: &str = "/-/deposit";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FaDepositReceiptContent {
+    pub receiver: PublicKeyHash,
+    pub ticket_balance: Amount,
+    pub run_function: Option<crate::receipt::RunFunction>,
+}
+
+#[derive(Display, Debug, Error, From)]
+pub enum FaDepositError {
+    InvalidHeaderValue,
+    InvalidUri,
+}
+
+fn deposit_to_receiver(
+    rt: &mut impl HostRuntime,
+    tx: &mut Transaction,
+    receiver: &PublicKeyHash,
+    ticket_hash: &TicketHash,
+    amount: Amount,
+) -> Result<FaDepositReceiptContent> {
+    let final_balance = TicketTable::add(rt, tx, receiver, ticket_hash, amount)?;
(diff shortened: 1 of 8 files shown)


crates/jstz_proto/src/operation.rs
```diff
@@ -140,6 +140,45 @@ pub mod external {
...
+        pub amount: Amount,
+        // Final deposit receiver address
+        pub receiver: Address,
+        // Optional proxy contract
+        pub proxy_smart_function: Option<Address>,
+        // Ticket hash
+        // Note: Can't use TicketHash (which is a typesafe wrapper around
+        // Blake2b) because it's not Serializable. Type needs to derive Serializable
+        // because it will be a child of ExternalOperation which derives Serializable.
+        // TODO: Revisit if Operation/ExternalOperation needs to be Serialiazable.
```
johnyob, line 159: Presumable we will always need to be `Operation`s and `Receipt`s to be serialisable?

crates/jstz_proto/src/executor/fa_deposit.rs
```diff
@@ -0,0 +1,406 @@
...
+            .authority(proxy_contract.to_string())
+            .path_and_query(DEPOSIT_URI)
+            .build()
+            .unwrap(),
+        method: Method::POST,
+        headers,
+        body: http_body,
+        gas_limit: FA_DEPOSIT_GAS_LIMIT,
+    }
+}
```
johnyob, line 62: It would be better to propagate the errors using a `Result` -- we should avoid `unwrap` or `expect` in the kernel code (as this may lead to panics)

crates/jstz_proto/src/executor/fa_deposit.rs
```diff
@@ -0,0 +1,406 @@
...
+    tx: &mut Transaction,
+    deposit: &FaDeposit,
+    proxy_contract: &PublicKeyHash,
+) -> Result<FaDepositReceiptContent> {
+    let run = new_run_function(deposit.to_http_body(), proxy_contract);
+    let source = PublicKeyHash::from_base58(NULL_ADDRESS).unwrap();
+    let result = smart_function::run::execute(rt, tx, &source, run, deposit.hash());
+    match result {
+        Ok(run_receipt) => {
+            if run_receipt.status_code != 200 {
```
johnyob, line 75: Should we also consider checking the range `200` to `299` (aka `response.ok`)

crates/jstz_proto/src/operation.rs
zcabter: Why do `Operation`s need to be serialisable?

zcabter: Edit: If I remove De/Serialize derives from `ExternalOperation`, build passes so it can be safely removed. Wdyt?

crates/jstz_proto/src/executor/fa_deposit.rs
zcabter: Returned result instead.

crates/jstz_proto/src/executor/fa_deposit.rs
zcabter: Done

crates/jstz_proto/src/operation.rs
johnyob: Yeah, I think removing the `De`/`Serialize` derives from `Message` and `ExternalOperation` should be safe

johnyob:
lgtm. I suggest we remove the unnecessary serialisers and then merge this 🚀

crates/jstz_proto/src/operation.rs
zcabter: Done

johnyob (approved):

PR #531, opened 2024-08-01
feat(fa-deposit): handle fa deposits in jstz kernel

zcabter:
# Context
Plugs fa deposit handling into the kernel.

# Description
Since fa deposits are done using tickets, the rollup type needs to be updated to ensure correct accounting of tickets to the rollup. 

`jstz_mock` was restructured to be more modular.

The kernel was updated to parse and read the fa deposit internal message. Since the execution always returns a receipt (since #530), the receipts are handled and written. 

Depends on #530 

# Manual testing
`cargo test -p jstz_kernel entry_fa_deposit_succeeds`


crates/jstz_kernel/src/parsing.rs
```diff
@@ -0,0 +1,137 @@
...
+use tezos_smart_rollup::michelson::{ticket::FA2_1Ticket, MichelsonContract};
+use tezos_smart_rollup::types::{Contract, PublicKeyHash as TezosPublicKeyHash};
+
+// FIXME: This should return the result instead
+pub fn ticket_hash(ticket: &FA2_1Ticket) -> Blake2b {
+    let mut bytes = Vec::new();
+    ticket.creator().bin_write(&mut bytes).unwrap();
+    ticket.contents().bin_write(&mut bytes).unwrap(); // FIXME: limit how large the content can be
+    Blake2b::from(&bytes)
+}
```
johnyob, line 15: I assume the lack of result is due to the anyhow vs error variant issue within the jstz codebase?

We should attach some issue links to these FIXME comments

crates/jstz_kernel/src/parsing.rs
johnyob, line 1: What is the responsibility split between the parsing and inbox modules now?

crates/jstz_mock/src/lib.rs
```diff
@@ -1 +1,49 @@
...
+    amount: u32,
+    content: (u32, Option<Vec<u8>>),
+) -> FA2_1Ticket {
+    let ticket_content = MichelsonPair(
+        MichelsonNat::from(content.0),
+        MichelsonOption::<MichelsonBytes>(content.1.clone().map(MichelsonBytes)),
+    );
+
+    Ticket::new(
+        Contract::Originated(ticketer.clone()),
```
johnyob, line 25: nit: I think most of the `.clone`s here can be avoided. Destructuring content should avoid the clone above, and I think you can simply move `ticketer` since it's owned and not used elsewhere

crates/jstz_mock/src/host.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn rt(&mut self) -> &mut MockHost {
+        &mut self.0
+    }
+}
+
+pub trait MockInternalMessage {
+    fn source(&self) -> PublicKeyHash;
+    fn sender(&self) -> ContractKt1Hash;
+    fn smart_rollup(&self) -> Option<SmartRollupAddress>;
+}
```
johnyob, line 82: Shouldn't this trait reside in the `message` module (`message/mod.rs`)?

crates/jstz_kernel/src/parsing.rs
zcabter: Yeah, I get derive errors when I wrap it as a `Jstz_proto` result variants

crates/jstz_kernel/src/parsing.rs
zcabter: `parsing` exposes functions that parse Michelson types into Jstz types.
`inbox` exposes functions for interacting with the inbox. So far, there's only `read_message`

crates/jstz_mock/src/host.rs
zcabter: Yep, moved

crates/jstz_mock/src/lib.rs
zcabter: `.clone`s removed

crates/jstz_kernel/src/parsing.rs
zcabter: Btw, removed this func as we now use the TicketHash directly from `FA2_1Ticket`

crates/jstz_kernel/src/parsing.rs
johnyob: I concur with this split. I ask since I think in future we may want to move functions from `inbox.rs` to `parsing.rs` if that is the split we're aiming for :)

johnyob (approved):
lgtm 🚀

PR #537, opened 2024-08-08
fix(crypto): prevent panic in `Signature::verify` due to invalid signatures

johnyob:
# Context

I experienced a kernel panic when testing out a smart function locally. 
The offending line was an `.unwrap()` when verifying signatures, due to an invalid signature.

Note: I'm not too sure how an invalid signature occurred, I suspect a mismatch between serde versions on the CLI and 
built WASM kernel used in the sandbox. 

# Description

# Manually testing the PR

I haven't been able to reproduce the panic 😔 So no tests, but the improvement is clear


zcabter:
Lgtm

zcabter (approved):

PR #540, opened 2024-08-13
fix(sandbox): add yield point to join task

zcabter:
# Context
Jstz node endpoint should be reachable now.

# Description

# Manually testing the PR


johnyob:
lgtm, tested locally 🚀

johnyob (approved):

PR #543, opened 2024-08-15
chore: standardise Cargo.toml files

johnyob:
# Context

# Description

This PR
- Attempts to unify and standardise the Cargo.toml files used through the project
- Order the dependencies alphabetically. 
   In the future it would be nice to use `cargo-sort` and have a CI check to keep it consistent but currently it https://github.com/DevinR528/cargo-sort/issues/54
- Hoist up dependencies to the main Cargo.toml and use the workspace dependency inheritance to align versions and provide a way to cache Cargo dependencies (as used in #542 )

In future, 
- External dependencies must be added to the root `Cargo.toml` (unless not supported as a workspace dependencies, such as _optional_ dependencies)
- Internal crates are referred to using `{ path = <crate path> }`

# Manually testing the PR

Check everything builds and passes tests :) 
```sh
make build
make test
```

### DIFF
--- a/crates/jstz_cli/Cargo.toml
+++ b/crates/jstz_cli/Cargo.toml
@@ -15,60 +15,60 @@ default-run = "jstz"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-jstz_kernel.workspace = true
-jstz_crypto.workspace = true
-jstz_proto.workspace = true
-jstz_core.workspace = true
-jstz_api.workspace = true
-jstz_rollup.workspace = true
-jstz_node.workspace = true
-octez.workspace = true
-clap = { version = "^4.4", features = ["derive"] }
-serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
-bs58 = "0.5"
-hex = "0.4"
-dirs = "3.0"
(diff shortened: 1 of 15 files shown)


zcabter:
Why do internal crates need to be referred to with relative path?

crates/jstz_cli/Cargo.toml
```diff
@@ -15,60 +15,60 @@ default-run = "jstz"
...
+dirs.workspace = true
+env_logger.workspace = true
+fs_extra.workspace = true
+futures-util.workspace = true
+futures.workspace = true
+hex.workspace = true
+http.workspace = true
+in-container.workspace = true
+indicatif.workspace = true
+jstz_api = { path = "../jstz_api" }
```
zcabter, line 39: Out of curiosity, why can't we refer to internal crates with `.workspace`?

zcabter:
Build and tested locall ✅

zcabter (approved):

crates/jstz_cli/Cargo.toml
johnyob: Since we rely on `.workspace` to differentiate between external dependencies and internal crates. 
We rely on this in #542 when building all the workspace dependencies (`cargoDeps`). If we don't have this distinction, then we'd essentially get cache misses each time, as the workspace dependencies would include all internal crates (which change in almost every PR)

PR #544, opened 2024-08-15
ci: use nix flake for environment

johnyob:
# Context

**Related Tasks**: [jstz-66](https://linear.app/tezos/issue/JSTZ-66/update-ci-to-use-nix-flake-instead-of-actions-environment)

# Description

**Dependencies**: #543 #542 

Does what it says on the tin. 

Future work includes:
- Building docs using Nix
- Linting Rust code using Nix (currently we rely on `make lint` which will pull its own dependencies)
- Using self-hosted runners for better cache locality + exposing the cache to devs (for local environments). `magic-nix-cache` isn't very good and will often drop cached dependencies in favour of space

# Manually testing the PR

Green CI 🚀

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,61 +12,34 @@ jobs:
       - uses: actions/checkout@v4
       - uses: wagoid/commitlint-github-action@v6
 
-  fmt:
-    name: Format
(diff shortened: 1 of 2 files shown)


zcabter (approved):
green ci, lgtm!

PR #548, opened 2024-08-21
feat(outbox): implement outbox queue on transaction

zcabter:
# Context
This PR omplements the lower level transactional outbox queue api

Part of [JSTZ-27](https://linear.app/tezos/issue/JSTZ-27/implement-outbox) 

# Description
## Why
The kernel is only allowed to write 100 messages per level. Luckily, the kernel sdk to expose an [OutboxQueue](https://gitlab.com/tezos/tezos/-/blob/master/src/kernel_sdk/sdk/src/outbox.rs#L108) which decouples enqueuing outbox messages and writing to the outbox, allowing a much high upper limit of outbox messages to be persisted as long as they are flushed (ideally, every level). However, the `OutboxQueue` is eager to write to persistent storage on message enqueue which makes it difficult to rollback and doesn't fit our transaction model. 

## What
Instead, we define an outbox queue for each snapshot that adheres to the transaction rules. Furthermore, an api is added to transaction to enqueue messages to the latest snapshot and to flush messages (on commit). 
Here are high level details:
* Each snapshot in the transaction carries its own outbox queue. 
* On enqueue, the outbox message is appended to the latest snapshot
* On commit:
  * The current snapshot queue is merged with its parent, respecting the enqueue order of messages. 
   * If the snapshot is at the bottom of the stack, the snapshot queue will be flushed instead.
 * On flush, the rollup outbox queue is flushed (written to outbox) first then the snapshot outbox queue until 100 messages have been flushed . The remaining messages in the snapshot outbox queue will be pushed into the rollup outbox queue. The next time flush is called, these remaining messages will be written out first followed by the snapshot and so on.

# Manually testing the PR
`cargo test -p jstz_core kv`

### DIFF
--- a/crates/jstz_core/src/kv/outbox.rs
+++ b/crates/jstz_core/src/kv/outbox.rs
@@ -1,0 +1,526 @@
+use crate::error::Result;
+use derive_more::{Display, Error, From};
+use serde::{Deserialize, Serialize};
+use tezos_smart_rollup::{
+    core_unsafe::MAX_OUTPUT_SIZE,
+    michelson::{ticket::FA2_1Ticket, MichelsonContract, MichelsonPair},
+    outbox::{
+        AtomicBatch, OutboxMessageFull, OutboxMessageTransactionBatch, OutboxQueue,
+    },
+    prelude::debug_msg,
+};
+
+use tezos_data_encoding::{enc::BinWriter, encoding::HasEncoding, nom::NomReader};
+use tezos_smart_rollup_host::{path::RefPath, runtime::Runtime};
+
+use super::Storage;
+
+const PERSISTENT_OUTBOX_QUEUE_ROOT: RefPath<'static> =
+    RefPath::assert_from(b"/outbox/persistent");
+
+const JSTZ_OUTBOX_QUEUE_META: RefPath<'static> = RefPath::assert_from(b"/outbox/meta");
+
+type NativeWithdrawalParameters = MichelsonPair<MichelsonContract, FA2_1Ticket>;
+
+type Withdrawal = OutboxMessageTransactionBatch<NativeWithdrawalParameters>;
+
+#[derive(Debug, HasEncoding, PartialEq)]
+pub enum OutboxMessage {
+    Withdrawal(Withdrawal),
+}
+
+impl AtomicBatch for OutboxMessage {}
+
+impl BinWriter for OutboxMessage {
+    fn bin_write(&self, output: &mut Vec<u8>) -> tezos_data_encoding::enc::BinResult {
+        match self {
+            OutboxMessage::Withdrawal(withdrawal) => withdrawal.bin_write(output),
+        }
+    }
+}
+
+impl<'a> NomReader<'a> for OutboxMessage {
+    fn nom_read(input: &'a [u8]) -> tezos_data_encoding::nom::NomResult<'a, Self> {
+        nom::branch::alt((nom::combinator::map(Withdrawal::nom_read, |withdrawal| {
+            OutboxMessage::Withdrawal(withdrawal)
+        }),))(input)
+    }
+}
+
+impl From<OutboxMessage> for OutboxMessageFull<OutboxMessage> {
+    fn from(message: OutboxMessage) -> Self {
+        match message {
+            OutboxMessage::Withdrawal(_) => {
+                OutboxMessageFull::AtomicTransactionBatch(message)
+            }
+        }
+    }
+}
+
+/// Represents a pending outbox queue stored as part of the
+/// trasaction's snapshot.
+#[derive(Debug, Default)]
+pub(crate) struct SnapshotOutboxQueue(Vec<OutboxMessage>);
+
+impl SnapshotOutboxQueue {
+    pub fn extend(&mut self, queue: SnapshotOutboxQueue) {
+        self.0.extend(queue.0)
+    }
+
+    pub fn queue_message(&mut self, message: OutboxMessage) {
+        self.0.push(message)
+    }
+}
+
+/// Wrapper over the kernel sdk's [OutboxQueue] which tracks the queue's
+/// current and maximum length. It is implemented as a singleton, that is,
+/// only 1 instance of [PersistentOutboxQueueInner] can exists in durable
+/// storage dring the lifetime of jstz. This struct should never be used
+/// directly outside of [PersistentOutboxQueue].
+#[derive(Debug)]
+struct PersistentOutboxQueueInner {
+    meta: OutboxQueueMeta,
+    rollup_outbox_queue: OutboxQueue<'static, RefPath<'static>>,
+}
+
+impl PersistentOutboxQueueInner {
+    /// Initializes a new [PersistentOutboxQueueInner] instance in durable
+    /// storage. Fails if an instance of the outbox queue already exists.
+    fn try_new(rt: &mut impl Runtime, max: u32) -> Result<Self> {
+        if OutboxQueueMeta::load(rt)?.is_some() {
+            Err(OutboxError::OutboxQueueMetaAlreadyExists)?
+        }
+        let meta = OutboxQueueMeta { len: 0, max };
+        meta.save(rt)?;
+        let rollup_outbox_queue = OutboxQueue::new(&PERSISTENT_OUTBOX_QUEUE_ROOT, max)?;
+        Ok(PersistentOutboxQueueInner {
+            meta,
+            rollup_outbox_queue,
+        })
+    }
+
+    fn load(rt: &mut impl Runtime) -> Result<Self> {
+        let meta =
+            OutboxQueueMeta::load(rt)?.ok_or(OutboxError::OutboxQueueMetaNotFound)?;
+        let rollup_outbox_queue =
+            OutboxQueue::new(&PERSISTENT_OUTBOX_QUEUE_ROOT, meta.max)?;
+        Ok(PersistentOutboxQueueInner {
+            meta,
+            rollup_outbox_queue,
+        })
+    }
+}
+
+/// A lazily initialized persistent outbox queue which is simply a wrapper over
+/// [PersistentOutboxQueueInner]
+#[derive(Debug, Default)]
+pub(crate) struct PersistentOutboxQueue {
+    inner: Option<PersistentOutboxQueueInner>,
+}
+
+impl PersistentOutboxQueue {
+    fn init_inner(&mut self, rt: &mut impl Runtime) -> Result<()> {
+        if self.inner.is_none() {
+            let max = u16::MAX as u32;
+            let inner = PersistentOutboxQueueInner::load(rt)
+                .or_else(|_| PersistentOutboxQueueInner::try_new(rt, max))?;
+            self.inner = Some(inner);
+        }
+        Ok(())
+    }
+
+    /// Get or initialize the [PersistentOutboxQueueInner] instance
+    fn get_or_init_inner_mut(
+        &mut self,
+        rt: &mut impl Runtime,
+    ) -> Result<&mut PersistentOutboxQueueInner> {
+        self.init_inner(rt)?;
+        let value = self
+            .inner
+            .as_mut()
+            .expect("Expecteded inner to be initialized");
+        Ok(value)
+    }
+
+    /// Flushes the outbox queue and updates the queue length.
+    /// WARN: Write heavy function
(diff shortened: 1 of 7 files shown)


crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub fn max(&self) -> u32 {
+        self.meta.max
+    }
+
+    /// Flushes the outbox queue in the order of the rollup outbox queue
+    /// then the outbox queue snapshot. The outbox has a maximum capacity
+    /// of 100 messages per level. Messages that cannot be flushed in the
+    /// current level are enqueued into the rollup outbox queue for the
+    /// next flush.
+    pub fn flush(
```
johnyob, line 152: Having read the kernel SDK implementation, I'm a bit concerned that we're writing to durable storage too much here, each `rollup_outbox_queue.queue_message` updates the `meta` in durable storage.   

What I suggest is that we leave a comment here (associated with a linear issue) for optimising durable storage writes.

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+        let mut flushed_count = self.rollup_outbox_queue.flush_queue(host);
+
+        // 2. Flush the outbox queue snapshot if there is space in the outbox
+        let mut tezos_outbox_messages =
+            snapshot.0.into_iter().map(|message| message.into());
+
+        for message in tezos_outbox_messages.by_ref() {
+            match Self::write_outbox_message(host, &message) {
+                Ok(()) => {
+                    flushed_count += 1;
```
johnyob, line 167: Why increment `flushed_count` here? `message` was never stored in `rollup_outbox_queue`?

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+                }
+                Err(crate::Error::HostError {
+                    source:
+                        tezos_smart_rollup::host::RuntimeError::HostErr(
+                            tezos_smart_rollup_host::Error::FullOutbox,
+                        ),
+                }) => {
+                    self.rollup_outbox_queue
+                        .queue_message(host, message)
+                        .unwrap();
```
johnyob, line 178: Increment `self.meta.queue_len` here

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+                }
+            }
+        }
+
+        //  3. Enqueue the remaining messages into the outbox queue
+        for message in tezos_outbox_messages {
+            self.rollup_outbox_queue
+                .queue_message(host, message)
+                .expect("Should always be able to queue message");
+        }
```
johnyob, line 198: Increment `self.meta.queue_len` here

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        // 4. Finally, update the counter outbox messages
+        self.meta.queue_len -= flushed_count as u32;
```
johnyob, line 201: Incorrect `queue_len` since we don't update `self.meta` when calling `rollup_outbox_queue.queue_message`

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+            let account =
+                PublicKeyHash::digest(format!("account{}", i).as_bytes()).unwrap();
+            messages.push(make_withdrawal(&account))
+        }
+        let mut messages = messages.into_iter();
+
+        let max = u16::MAX as u32;
+        let mut outbox_queue = OutboxQueue {
+            meta: OutboxQueueMeta {
+                queue_len: 120,
```
johnyob, line 378: Incorrect setup? Initial queue length is `60`

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+
+        for message in messages.by_ref().take(60) {
+            OUTBOX_QUEUE.queue_message(&mut host, message).unwrap();
+        }
+
+        let outbox_queue_snapshot = OutboxQueueSnapshot(messages.take(60).collect());
+
+        outbox_queue.flush(&mut host, outbox_queue_snapshot);
+
+        assert_eq!(20, outbox_queue.queue_len());
```
johnyob, line 393: Incorrect calculation: 120 - 60 (flushed) - 40 (written) = 20
Should be: 60 - 60 (flushed) + 20 (queued) = 20

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -338,6 +350,17 @@ impl Transaction {
...
     }
 
+    pub fn push_outbox_message(&mut self, message: OutboxMessage) -> Result<()> {
+        if self.outbox_queue.queue_len() + 1 > self.outbox_queue.max() {
+            Err(OutboxError::OutboxFull)?;
+        }
+
+        let current_outbox_queue = self.current_snapshot()?.outbox_queue_mut();
+        current_outbox_queue.queue_message(message);
+        self.outbox_queue.incr_queue_len();
```
johnyob, line 360: queue_message should be tied to `incr_queue_len` -- this leaks the invariant that `queue_len()` == number of messages in the queue, since it's possible to use `incr_queue_len` without queuing anything (or vice versa)

johnyob (changes requested):

crates/jstz_core/src/kv/outbox.rs
```diff
@@ -0,0 +1,461 @@
...
+/// Exposing tezos_smart_rollup::outbox::OUTBOX_QUEUE_ROOT
+const ROLLUP_OUTBOX_QUEUE_ROOT: RefPath = RefPath::assert_from(b"/__sdk/outbox");
+
+const OUTBOX_QUEUE_ROOT: RefPath = RefPath::assert_from(b"/outbox");
+
+type RollupOutboxQueue = KernelSdkOutboxQueue<'static, RefPath<'static>>;
+
+type NativeWithdrawalParameters = MichelsonPair<MichelsonContract, FA2_1Ticket>;
+
+// TODO: Might need to use OutboxMessageTransactionBatch for L1 encoding
```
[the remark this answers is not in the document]
zcabter: > How should the user define a set of outbox messages that are executed atomically?
It depends on the API expose to the user. In the PR, the outbox only exposes the native withdraw which is ok to just be a single message. Down the road, we could could make use of batch messages depending on the use case.
> How should/can the user handle errors that result from outbox messages?
Since outbox message application is a completely decoupled step, I think the way to go is for the user to have use an external setup to react to the errors

crates/jstz_core/src/kv/outbox.rs
zcabter: The point here is that instead of enqueing messages from `OutboxQueueSnapshot`s in the rollup's outbox queue only to flush it in the next line, we write as many messages to the kernel outbox ourselves. This saves us some unnecessary writes. So we take into account such writes by incrementing`flushed_out`.

crates/jstz_core/src/kv/outbox.rs
zcabter: Ah, we don't have to. `self.meta.queue_len` is the sum of `rollup_outbox_queue` length and `OutboxQueueSnapshot` length. We use `rollup_outbox_queue.queue_message` here to simply move messages from in-memory queue to the persistent queue.

crates/jstz_core/src/kv/transaction.rs
zcabter: I agree. I was running into rust troubles when doing that since `queue_message` would need to borrow `self` from`self.current_snapshot()?.outbox_queue_mut()` and `self.outbox_queue` at the same time. 

I could move the outbox implementation into transaction itself although the code organisation would not be as nice. Thoughts?

crates/jstz_core/src/kv/outbox.rs
johnyob: That seems a rather messy/error prone invariant -- since `OutboxQueueSnapshot` isn't contained within `OutboxQueue`? What I suggest is that a `Transaction` has a notion of a 'combined' outbox queue length (since it contains both the list of snapshots to write and the persistent outbox queue).

crates/jstz_core/src/kv/transaction.rs
johnyob: See above suggestion on moving the combined outbox queue length to the transaction. This should fix the borrow checking problems you encountered

crates/jstz_core/src/kv/outbox.rs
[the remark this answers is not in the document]
zcabter: Hmm we would need an oracle to watch for outbox message execution and the suspension would be for 2 weeks. But would be cool because  we could essentially revert the effects of failed application. For eg. if outbox withdrawal operation failed, then the protocol could restore the user's balance. 

Maybe an easier solution is to ensure that the withdrawal will always succeed? then the user can assume that a successful write to the outbox is as good as gold.

crates/jstz_core/src/kv/outbox.rs
zcabter: I moved the meta into `Transaction` and added RollupOutboxQueue to Transaction. `meta` tracks the combined queue length of snapshots and rollup outbox queues and ensures the queue length is limited to `max` when messages are enqueued in the snapshot. 

`flush` no longer does the queue len accounting.Iinstead, the accounting is done by `Transaction` since it holds all the information - `outbox_queue_meta`, `rollup_outbox_queue` and  `snapshop.snapshot_outbox_queue`

crates/jstz_core/src/kv/outbox.rs
zcabter: Done

crates/jstz_core/src/kv/outbox.rs
zcabter: Btw I pushed the refactoring as second commit so that you can see the changes or revert if needed. Will squash eventually.

crates/jstz_core/src/kv/transaction.rs
zcabter: See: https://github.com/jstz-dev/jstz/pull/548#discussion_r1735062412 Let's consolidate our discussion there.

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -55,6 +62,8 @@ pub struct Transaction {
     // A stack of transactional snapshots
     stack: Vec<Snapshot>,
     lookup_map: LookupMap,
+    rollup_outbox_queue: RollupOutboxQueue,
+    outbox_queue_meta: OutboxQueueMeta,
 }
```
johnyob, line 67: This still leaks the invariant slightly when using `Transaction::default` since you can get a `len` of `0` when `OUTBOX_QUEUE` might not actually have a length of `0`. I know that `Transaction::load` solves this, but perhaps we should remove `Transaction::default`. 

Alternatively, we split the meta into the non-persistent length and the persistent length and lazily load the persistent length when required and maintain it? Non-persistent length could be maintained as a field in `Transaction` and persistent length managed by `RollupOutboxQueue`?

crates/jstz_core/src/kv/outbox.rs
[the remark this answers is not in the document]
johnyob: That is definitely a nicer approach, completely forgot that we'd need to suspend it for 2 weeks 😬 😅

crates/jstz_core/src/kv/outbox.rs
johnyob: I'm still not entirely satisfied with this separation + it leaks the invariant still (via `default`). My take is that we should have two structs:
- `PersistentOutboxQueue`
- `SnapshotOutboxQueue`

The total length of the non-persistent (snapshot) outbox queues is maintained by a field in `Transaction` maintained by `Transaction::queue_outbox_message`. 

The maximum and persistent length is managed by `PersistentOutboxQueue` using a `queue_meta : Option<OutboxQueueMeta>` field. This permits us to lazily read the metadata by passing a `&rt` in getters (`PersistentOutboxQueue::queue_len(&impl Runtime)` and `PersistentOutboxQueue::max_len(&impl Runtime)`)
Flushing will decrease the in-memory representation of the persistent length. 

The transactions total outbox queue length is simply the sum of the two. `Transaction::queue_outbox_message` is modified to take an `&rt`.

zcabter: Made the changes as discussed here and on call

PR #550, opened 2024-08-29
feat(withdrawal): implement native withdrawal

zcabter:
# Context
Implements native withdrawal from external messages.

# Description

# Manually testing the PR
`cargo test -p jstz_proto -- executor::smart_function executor::withdraw`

### DIFF
--- a/crates/jstz_kernel/src/lib.rs
+++ b/crates/jstz_kernel/src/lib.rs
@@ -18,25 +18,26 @@ fn read_ticketer(rt: &impl Runtime) -> Option<ContractKt1Hash> {
     Storage::get(rt, &TICKETER).ok()?
 }
 
-fn handle_message(hrt: &mut impl Runtime, message: Message) -> Result<()> {
-    let mut tx = Transaction::default();
-    tx.begin();
-
+fn handle_message(
+    hrt: &mut impl Runtime,
+    message: Message,
+    ticketer: &ContractKt1Hash,
+    tx: &mut Transaction,
+) -> Result<()> {
     match message {
         Message::Internal(external_operation) => {
             let receipt =
-                executor::execute_external_operation(hrt, &mut tx, external_operation);
-            receipt.write(hrt, &mut tx)?
+                executor::execute_external_operation(hrt, tx, external_operation);
+            receipt.write(hrt, tx)?
         }
         Message::External(signed_operation) => {
             debug_msg!(hrt, "External operation: {signed_operation:?}\n");
-            let receipt = executor::execute_operation(hrt, &mut tx, signed_operation);
+            let receipt =
+                executor::execute_operation(hrt, tx, signed_operation, ticketer);
             debug_msg!(hrt, "Receipt: {receipt:?}\n");
-            receipt.write(hrt, &mut tx)?
+            receipt.write(hrt, tx)?
         }
     }
-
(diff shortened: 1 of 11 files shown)


crates/jstz_kernel/src/lib.rs
```diff
@@ -18,37 +18,40 @@ fn read_ticketer(rt: &impl Runtime) -> Option<ContractKt1Hash> {
...
-
-    if let Some(message) = read_message(rt, ticketer) {
-        handle_message(rt, message)
+    let mut tx = Transaction::default();
+    tx.begin();
+    if let Some(message) = read_message(rt, &ticketer) {
+        handle_message(rt, message, &ticketer, &mut tx)
             .unwrap_or_else(|err| debug_msg!(rt, "[🔴] {err:?}\n"));
     }
+    tx.commit(rt).unwrap();
```
johnyob, line 54: Prefer logging the error from `tx.commit` than panicking (its always better practice in the kernel to avoid panics (unless in testing code))

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -444,6 +444,215 @@ pub mod run {
...
+    }
+
+    pub fn execute(
+        hrt: &mut impl HostRuntime,
+        tx: &mut Transaction,
+        source: &Address,
+        run: operation::RunFunction,
+        ticketer: &ContractKt1Hash,
+    ) -> Result<receipt::RunFunction> {
+        if run.uri.host() != Some("jstz") {
```
johnyob, line 488:
```suggestion
        if run.uri.host() != Some(JSTZ_HOST) {
```

johnyob (commented):
Overall lgtm, some minor comments about reusing the `JSTZ_HOST` constant

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -444,6 +444,215 @@ pub mod run {
...
+            return Err(Error::InvalidHost);
+        }
+        match run.uri.path() {
+            WITHDRAW_PATH => {
+                // TODO: https://linear.app/tezos/issue/JSTZ-77/check-gas-limit-when-performing-native-withdraws
```
alanmarkoTrilitech, line 493: can I please get access to the linear workspace?

zcabter: I'm not an admin. @johnyob could you please add @alanmarkoTrilitech ?

crates/jstz_kernel/src/lib.rs
zcabter: Alright, done.

zcabter:
Btw, we need to remove `jstz_proto` from dev dependencies in `jstz_core` The circular reference is causing rust-analyzer to not work for tests in `jstz_core`. It looks like we only need `jstz_proto` to add balance which we can simulate without it or write new tests that tests transaction more directly. Tracked [JSTZ-107](https://linear.app/tezos/issue/JSTZ-107/remove-dev-dependency-on-jstz-proto-from-jstz-core)

crates/jstz_proto/src/executor/smart_function.rs
johnyob: Invited

johnyob (approved):
lgtm 🚢

PR #551, opened 2024-09-04
feat(cli): add bridge withdraw command

zcabter:
# Context
Adds `bridge withdraw` command to the cli

Closes [JSTZ-105](https://linear.app/tezos/issue/JSTZ-105/implement-cli-command-for-withdraw)

# Description
Underneath the hood, the cli command prepares and calls the run operation against `/jstz/withdraw`

# Manually testing the PR
1. Run the sanbox
```
cargo run -- sandbox start
```
2. Deposit funds form bootstrap1 to the logged in account 
```
cargo run -- bridge deposit --from bootstrap1 --to bob --amount 1000
```
3. Withdraw funds from the logged in account to L1 address
```
cargo run -- bridge withdraw -t tz1dbGzJfjYFSjX8umiRZ2fmsAQsk8XMH1E9 -a 123.123456
```
4. Execute the outbox using octez client
```
./scripts/execute_latest_outbox_message.sh $OCTEZ_CLIENT
```
5. Check balance
```
$OCTEZ_CLIENT -E http://127.0.0.1:18730 get balance for tz1dbGzJfjYFSjX8umiRZ2fmsAQsk8XMH1E9
```

### DIFF
--- a/crates/jstz_cli/src/bridge/withdraw.rs
+++ b/crates/jstz_cli/src/bridge/withdraw.rs
@@ -1,0 +1,39 @@
+use crate::{
+    bridge::convert_tez_to_mutez,
+    config::{Config, NetworkName},
+    error::{bail_user_error, Result},
+    run,
+    term::styles,
+    utils::AddressOrAlias,
+};
+use log::debug;
+
+pub async fn exec(
+    to: AddressOrAlias,
+    amount: f64,
+    network: Option<NetworkName>,
+) -> Result<()> {
+    let cfg = Config::load()?;
+
+    // Check network
+    if cfg.network_name(&network)? == NetworkName::Dev && cfg.sandbox.is_none() {
+        bail_user_error!(
+            "No sandbox is currently running. Please run {}.",
+            styles::command("jstz sandbox start")
+        );
+    }
+
+    let to_pkh = to.resolve_l1(&cfg, &network)?;
+    debug!("resolved `to` -> {}", &to_pkh.to_base58());
+
+    let amount = convert_tez_to_mutez(amount)?;
+    let url = "tezos://jstz/withdraw".to_string();
+    let http_method = "POST".to_string();
+    let gas_limit = 10; // TODO: set proper gas limit
+    let withdraw = jstz_proto::executor::withdraw::Withdrawal {
+        amount,
+        receiver: to_pkh,
+    };
+    let json_data = serde_json::to_string(&withdraw)?;
+    run::exec(url, http_method, gas_limit, Some(json_data), network, false).await
+}
--- a/crates/jstz_cli/src/run.rs
+++ b/crates/jstz_cli/src/run.rs
@@ -4,6 +4,7 @@ use crate::logs::{exec_trace, DEFAULT_LOG_LEVEL};
 use anyhow::bail;
 use http::{HeaderMap, Method, Uri};
 use jstz_proto::context::account::Address;
+use jstz_proto::executor::JSTZ_HOST;
 use jstz_proto::{
     operation::{Content as OperationContent, Operation, RunFunction, SignedOperation},
     receipt::Content as ReceiptContent,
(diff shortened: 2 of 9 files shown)


crates/jstz_cli/src/bridge/withdraw.rs
```diff
@@ -0,0 +1,39 @@
...
+) -> Result<()> {
+    let cfg = Config::load()?;
+
+    // Check network
+    if cfg.network_name(&network)? == NetworkName::Dev && cfg.sandbox.is_none() {
+        bail_user_error!(
+            "No sandbox is currently running. Please run {}.",
+            styles::command("jstz sandbox start")
+        );
+    }
```
johnyob, line 24: I assume this was copied from deposit, but this is actually a bug. Withdrawals are/should be supported on more than just the sandbox network

scripts/execute_latest_outbox_message.sh
johnyob, line 1: I assume the script is temporary until we upgrade to an octez version that supports the rollup node automatically executing outbox messages?

crates/jstz_cli/src/sandbox/daemon.rs
```diff
@@ -201,7 +201,7 @@ impl Sandbox {
                         }
 
                         // yield point
-                        tokio::time::sleep(Duration::from_millis(100)).await;
+                        tokio::time::sleep(Duration::from_millis(300)).await;
```
johnyob, line 204: Why the increase?

crates/jstz_cli/src/run.rs
```diff
@@ -27,6 +27,38 @@ use crate::{
 // where the FA2 transfer function was called 1000 times.
 pub const DEFAULT_GAS_LIMIT: u32 = 550000;
 
+pub const JSTZ_HOST: &str = "jstz";
```
johnyob, line 30: Reuse constant from `jstz_proto` code?

scripts/execute_latest_outbox_message.sh
zcabter: yes

crates/jstz_cli/src/sandbox/daemon.rs
zcabter: Oops, I was trying to debug why the sandbox was becoming flakey on operation injection, have reverted. 

Turns out that if the commitment period is too low, the rollup spends all its time creating commitment hashes and stops responding to RPC requests.

crates/jstz_cli/src/bridge/withdraw.rs
zcabter: Ah yeah. Isn't that true for deposit as well?

zcabter: This checks that the sandbox is running if the network is set to dev which is correct I think?

crates/jstz_cli/src/run.rs
zcabter: Ah yeah, done

crates/jstz_cli/src/bridge/withdraw.rs
johnyob: I think we want to check if a network is set, not whether the network is the sandbox. We only did it for deposit because we relied on the local `jstz_bridge` alias for depositing

zcabter:
Code cov does not look good. I'm not too sure how to test the cli without running the sandbox.

crates/jstz_cli/src/bridge/withdraw.rs
zcabter: Oh I see. I always thought it was to improve the UX since if `dev` is set, then the cli could provide a very helpful message to start the sandbox.

johnyob:
> Code cov does not look good. I'm not too sure how to test the cli without running the sandbox.

I don't think Codecov check should block the PR, especially since we don't have the infrastructure to really test CLI stuff effectively atm

crates/jstz_cli/src/bridge/withdraw.rs
johnyob: We should have a check for a network in the future w/ a helpful message, but it shouldn't only check for `dev`

crates/jstz_cli/src/sandbox/daemon.rs
johnyob: Lmao, we should probably tell the rollups team about that

crates/jstz_cli/src/bridge/withdraw.rs
johnyob: I suggest we fix this in a later PR, the issue is tracked here: https://linear.app/tezos/issue/JSTZ-74/support-bridge-commands-for-remote-networks

johnyob (approved):
lgtm :shipit:

PR #553, opened 2024-09-06
feat(jstzd): implement image

ryutamago:
# Context
Implement traits / struct to run docker container that can later be used for jstzd

The related tasks are the following but not all of them are fully covered/tested yet (e.g. mounts/hosts etc)
[image](https://linear.app/tezos/issue/JSTZ-61/implement-dockerimage)
[runnable image](https://linear.app/tezos/issue/JSTZ-57/implement-dockerrunnableimage)

# Description

Implement the docker images traits and structs

# Manually testing the PR
cargo test --package jstzd --lib

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,8 @@ urlpattern = "0.2.0"
 wasm-bindgen = "0.2.92"
 async-dropper = { version = "0.3.1", features = ["tokio", "simple"] }
 async-trait = "0.1.82"
+bollard = "0.16.1"
+
 
 [workspace.dependencies.tezos-smart-rollup]
 version = "0.2.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ base64 = "0.21.7"
 bincode = "1.3.3"
 boa_engine = { version = "0.17.0", features = ["fuzz"] }
 boa_gc = "0.17.0"
+bollard = "0.16.1"
 bs58 = "0.5"
 bytes = "1.4.0"
 chrono = { version = "0.4.34", default-features = false, features = ["std"] }
@@ -101,8 +102,6 @@ urlpattern = "0.2.0"
 wasm-bindgen = "0.2.92"
 async-dropper = { version = "0.3.1", features = ["tokio", "simple"] }
 async-trait = "0.1.82"
-bollard = "0.16.1"
-
 
 [workspace.dependencies.tezos-smart-rollup]
 version = "0.2.2"
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,9 +3,6 @@ comment:
   layout: "header, files, footer"
   hide_project_coverage: false
 
-ignore:
-  - crates/jstzd/src/main.rs
-
 # Configures coverage checks -- both patch and project-wide checks
 # can drop by 5% before failing the CI build.
 #
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,9 @@ comment:
   layout: "header, files, footer"
   hide_project_coverage: false
 
+ignore:
+  - crates/jstzd/src/main.rs
+
 # Configures coverage checks -- both patch and project-wide checks
 # can drop by 5% before failing the CI build.
 #
--- a/crates/jstzd/src/docker/runnable_image.rs
+++ b/crates/jstzd/src/docker/runnable_image.rs
@@ -1,0 +1,324 @@
+use crate::docker::Image;
+use anyhow::Result;
+use bollard::{
+    container::{Config, CreateContainerOptions},
+    secret::{ContainerCreateResponse, HostConfig, PortBinding, PortMap},
+    Docker,
+};
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    net::IpAddr,
+    sync::Arc,
+};
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum AccessMode {
+    ReadOnly,
+    ReadWrite,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum MountType {
+    Bind,
+    Volume,
+    Tmpfs,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Mount {
+    pub access_mode: AccessMode,
+    pub mount_type: MountType,
+    pub source: Option<String>,
+    pub target: Option<String>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Host {
+    Addr(IpAddr),
+    HostGateway,
+}
+
+impl ToString for Host {
+    fn to_string(&self) -> String {
+        match self {
+            Host::Addr(ip) => ip.to_string(),
+            Host::HostGateway => "host-gateway".to_string(),
+        }
+    }
+}
+
+type HostPort = u16;
+type ContainerPort = u16;
+pub struct RunnableImage<I: Image> {
+    pub image: I,
+    pub container_name: String,
+    pub overridden_cmd: Option<Vec<String>>,
+    pub network: Option<String>, //TODO: https://linear.app/tezos/issue/JSTZ-111/support-for-adding-network-to-a-container
+    pub env_vars: BTreeMap<String, String>,
+    pub hosts: BTreeMap<String, Host>,
+    pub mounts: Vec<Mount>, //TODO: https://linear.app/tezos/issue/JSTZ-110/implement-mounts-for-runnable-image
+    pub ports: BTreeMap<HostPort, ContainerPort>,
+}
+
+impl<I: Image> RunnableImage<I> {
+    pub fn new(image: I, container_name: impl Into<String>) -> Self {
+        Self {
+            image,
+            container_name: container_name.into(),
+            overridden_cmd: None,
+            network: None,
+            env_vars: BTreeMap::new(),
+            hosts: BTreeMap::new(),
+            mounts: Vec::new(),
+            ports: BTreeMap::new(),
+        }
+    }
+
+    pub fn with_overridden_cmd(mut self, cmd: Vec<String>) -> Self {
+        self.overridden_cmd = Some(cmd);
+        self
+    }
+
+    pub fn with_network(mut self, network: impl Into<String>) -> Self {
+        self.network = Some(network.into());
+        self
+    }
+
+    pub fn with_env_var(
+        mut self,
+        key: impl Into<String>,
+        value: impl Into<String>,
+    ) -> Self {
+        self.env_vars.insert(key.into(), value.into());
+        self
+    }
+
+    pub fn with_host(mut self, hostname: impl Into<String>, host: Host) -> Self {
+        self.hosts.insert(hostname.into(), host);
+        self
+    }
+
+    pub fn with_mount(mut self, mount: Mount) -> Self {
+        self.mounts.push(mount);
+        self
+    }
+
+    pub fn with_port(
+        mut self,
+        host_port: HostPort,
+        container_port: ContainerPort,
+    ) -> Self {
+        self.ports.insert(host_port, container_port);
+        self
+    }
+
+    pub async fn create_container(&self, client: Arc<Docker>) -> Result<String> {
+        self.all_container_ports_are_exposed()?;
+        let config = Config::<String> {
+            image: Some(self.image.image_uri().to_string()),
+            host_config: Some(self.host_config()),
+            entrypoint: self.entrypoint(),
+            cmd: self.overridden_cmd.clone(),
+            env: self.env(),
+            ..Config::default()
+        };
+        let options = Some(CreateContainerOptions {
+            name: self.container_name.clone(),
+            platform: None,
+        });
+        match client.create_container(options, config).await {
+            Ok(ContainerCreateResponse { id, .. }) => Ok(id),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn env(&self) -> Option<Vec<String>> {
+        match self.env_vars.is_empty() {
+            true => None,
+            false => Some(
+                self.env_vars
+                    .iter()
+                    .map(|(k, v)| format!("{}={}", k, v))
+                    .collect::<Vec<_>>(),
+            ),
+        }
+    }
+
+    fn entrypoint(&self) -> Option<Vec<String>> {
+        self.image.entrypoint().map(|e| {
+            e.split_whitespace()
+                .map(|s| s.to_string())
+                .collect::<Vec<String>>()
+        })
+    }
+
+    fn host_config(&self) -> HostConfig {
+        let port_bindings: PortMap = self.ports.iter().fold(
+            HashMap::new(),
+            |mut acc, (host_port, container_port)| {
+                let container_port = container_port.to_string();
+                let binding = PortBinding {
+                    host_ip: None,
+                    host_port: Some(host_port.to_string()),
+                };
+                acc.entry(container_port)
+                    .and_modify(|bindings| {
+                        if let Some(vec) = bindings {
+                            vec.push(binding.clone());
+                        }
+                    })
+                    .or_insert(Some(vec![binding]));
+                acc
+            },
+        );
+        let extra_hosts = self
+            .hosts
+            .iter()
+            .map(|(host, ip)| format!("{}:{}", host, ip.to_string()))
+            .collect::<Vec<_>>();
+        let mut config = HostConfig::default();
+        if !port_bindings.is_empty() {
+            config.port_bindings = Some(port_bindings);
+        }
+        if !extra_hosts.is_empty() {
+            config.extra_hosts = Some(extra_hosts);
+        }
+        config
+    }
+
+    // Check if the ports use the exposed container ports from the image if any
+    fn all_container_ports_are_exposed(&self) -> Result<()> {
+        let exposed_ports: HashSet<&u16> = self.image.exposed_ports().iter().collect();
+        if exposed_ports.is_empty() {
+            return Ok(());
+        }
+        let all_ports_are_exposed =
+            self.ports.values().all(|port| exposed_ports.contains(port));
+        if !all_ports_are_exposed {
+            return Err(anyhow::anyhow!("Invalid ports"));
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::docker::GenericImage;
+    use std::net::{IpAddr, Ipv4Addr};
+
+    #[test]
+    fn builds_runnable_image() {
+        let image = GenericImage::new("busybox");
+        let cmd = vec!["echo", "hello"]
+            .into_iter()
+            .map(String::from)
+            .collect();
+        let mount = Mount {
+            access_mode: AccessMode::ReadWrite,
+            mount_type: MountType::Bind,
+            source: Some("/source".to_string()),
+            target: Some("/target".to_string()),
+        };
+        let ip = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+        let runnable_image = RunnableImage::new(image, "busybox")
+            .with_overridden_cmd(cmd)
+            .with_network("some_network")
+            .with_env_var("KEY", "VALUE")
+            .with_host("hostname", Host::Addr(ip))
+            .with_mount(mount.clone())
+            .with_port(5000, 3000);
+
+        assert_eq!(runnable_image.container_name, "busybox");
+        assert_eq!(
+            runnable_image.overridden_cmd,
+            Some(vec!["echo".to_string(), "hello".to_string()])
+        );
+        assert_eq!(runnable_image.network, Some("some_network".to_string()));
+        assert_eq!(
+            runnable_image.env_vars,
+            vec![("KEY".to_string(), "VALUE".to_string())]
+                .into_iter()
+                .collect()
+        );
+        assert_eq!(
+            runnable_image.hosts,
+            vec![("hostname".to_string(), Host::Addr(ip))]
+                .into_iter()
+                .collect()
+        );
+        assert_eq!(runnable_image.mounts, vec![mount]);
+        assert_eq!(
+            runnable_image.ports,
+            vec![(5000, 3000)].into_iter().collect()
+        );
+    }
+
+    #[test]
+    fn parses_env() {
+        let image = GenericImage::new("busybox");
+        let runnable_image =
+            RunnableImage::new(image, "busybox_test").with_env_var("KEY", "VALUE");
+        let env: Option<Vec<String>> = runnable_image.env();
+        assert_eq!(env, Some(vec!["KEY=VALUE".to_string()]));
+    }
+
+    #[test]
+    fn parses_entrypoint() {
+        let image = GenericImage::new("busybox").with_entrypoint("echo hello");
+        let runnable_image = RunnableImage::new(image, "busybox_test");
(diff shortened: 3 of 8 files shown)


crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn image_name(&self) -> &str {
+        self.image_name
+    }
+
+    fn entrypoint(&self) -> Option<&str> {
+        self.entrypoint
+    }
+
+    fn image_tag(&self) -> &str {
+        self.image_tag.unwrap_or("latest")
```
zcabter, line 50:
```suggestion
        self.image_tag.unwrap_or(Self::LATEST_TAG)
```

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+use bollard::{
+    image::{CreateImageOptions, ListImagesOptions},
+    Docker,
+};
+
+#[async_trait]
+pub trait Image {
+    fn image_name(&self) -> &str;
+    fn image_tag(&self) -> &str {
+        "latest"
```
zcabter, line 14: Extract into local `const LATEST_TAG` then
```suggestion
        Self::LATEST_TAG
```

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+async fn create_image<I>(image: &I, client: Arc<Docker>) -> anyhow::Result<()>
+where
+    I: Image + ?Sized,
+{
+    let options = Some(CreateImageOptions {
+        from_image: image.image_name(),
+        tag: image.image_tag(),
+        ..Default::default()
```
zcabter, line 128:
```suggestion
        .. CreateImageOptions::default()
```

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+    });
+    let mut stream = client.create_image(options, None, None);
+    while let Some(create_info) = stream.next().await {
+        match create_info {
+            Ok(info) => {
+                if let Some(status) = info.status {
+                    println!("{:?}", status)
```
zcabter, line 135:
```suggestion
                    println!("{}", status)
```

zcabter:
The manual test fails with `Error: Docker responded with status code 404: No such image: busybox:latest`. Seems like the `pull_image` step is missing.

zcabter:
Commit messages are missing scope. Prefer to include it if possible. `feat(jstzd)` is a good one for this PR.

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    pub async fn start(&self) -> anyhow::Result<()> {
+        self.0.start().await
+    }
+
+    pub async fn stop(&self) -> anyhow::Result<()> {
+        self.0.stop().await
+    }
+}
```
johnyob, line 107: Suggest using `#[derive(AsyncDrop)]` on `ContainerInner` (to be renamed to `Container`) to avoid this wrapper boilerplate

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    fn image_tag(&self) -> &str {
+        self.image_tag.unwrap_or("latest")
+    }
+
+    fn exposed_ports(&self) -> &[u16] {
+        self.exposed_ports.unwrap_or(&[])
+    }
+}
```
johnyob, line 56: Suggest alternative name: `GenericImage`, externally this would read as `docker::GenericImage`. 

Additionally remove the lifetimes (all data here can easily be owned)

crates/jstzd/src/runnable_image/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+            platform: None,
+        });
+        match client.create_container(options, config).await {
+            Ok(container) => Ok(container.id),
+            Err(e) => Err(e.into()),
+        }
+    }
+}
+
+/////// helpers ///////
```
zcabter, line 91:
```suggestion
```

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+
+pub struct DockerImageBuilder<'a> {
+    image_name: &'a str,
+    image_tag: Option<&'a str>,
+    entrypoint: Option<&'a str>,
+    exposed_ports: Option<&'a [u16]>,
+}
```
johnyob, line 63: This builder is identical to `DockerImage`. You can reuse the same type. Separating the builder type with the main types makes sense when the types must differ

crates/jstzd/src/runnable_image/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+    if !port_bindings.is_empty() {
+        config.port_bindings = Some(port_bindings);
+    }
+    if !extra_hosts.is_empty() {
+        config.extra_hosts = Some(extra_hosts);
+    }
+    Some(config)
+}
+
+// check if the ports use the exposed container ports from the image if any
```
zcabter, line 147:
```suggestion
// Check if the ports use the exposed container ports from the image if any
```

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+            Ok(info) => {
+                if let Some(status) = info.status {
+                    println!("{:?}", status)
+                }
+            }
+            Err(e) => return Err(e.into()),
+        }
+    }
+    Ok(())
+}
```
johnyob, line 142: Suggest that these are methods on `RunnableImage`

crates/jstzd/src/network.rs
```diff
@@ -0,0 +1,60 @@
+use async_dropper::{AsyncDrop, AsyncDropper};
+use async_trait::async_trait;
+use bollard::{network::CreateNetworkOptions, Docker};
+use std::sync::Arc;
+
+#[derive(Default)]
+struct NetworkInner {
+    name: String,
+    client: Option<Arc<Docker>>,
```
johnyob, line 9: Ditto, why is this optional? Suggest to make it non-optional

crates/jstzd/src/runnable_image/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+pub struct RunnableImage<I: Image> {
+    pub image: I,
+    pub container_name: String,
+    pub overridden_cmd: Option<Vec<String>>,
+    pub network: Option<String>, //todo
+    pub env_vars: BTreeMap<String, String>,
+    pub hosts: BTreeMap<String, Host>,
+    pub mounts: Vec<Mount>, //todo
+    pub ports: BTreeMap<HostPort, ContainerPort>,
+}
```
johnyob, line 64: Similarly, the builder and runnable image are the same type. Recommend using a single type (and file) for these

crates/jstzd/src/network.rs
johnyob: Also remove boilerplate using derive for AsyncDrop

crates/jstzd/src/runnable_image/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+
+fn entrypoint<I: Image>(runnable_image: &RunnableImage<I>) -> Option<Vec<String>> {
+    runnable_image.image.entrypoint().map(|e| {
+        e.split_whitespace()
+            .map(|s| s.to_string())
+            .collect::<Vec<String>>()
+    })
+}
+
+fn host_config<I: Image>(runnable_image: &RunnableImage<I>) -> Option<HostConfig> {
```
johnyob, line 113: All of these helpers should be private methods

zcabter, line 113: `HostConfig` doesn't have a None case

```suggestion
fn host_config<I: Image>(runnable_image: &RunnableImage<I>) -> HostConfig> {
```

crates/jstzd/src/runnable_image/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+type HostPort = u16;
+type ContainerPort = u16;
+pub struct RunnableImage<I: Image> {
+    pub image: I,
+    pub container_name: String,
+    pub overridden_cmd: Option<Vec<String>>,
+    pub network: Option<String>, //todo
+    pub env_vars: BTreeMap<String, String>,
+    pub hosts: BTreeMap<String, Host>,
+    pub mounts: Vec<Mount>, //todo
```
zcabter, line 62: Create an issue in linear to track the todo

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
...
+    _private: (),
+}
+
+pub struct Container(AsyncDropper<ContainerInner>);
+
+impl ContainerInner {
+    /// Creates a new container with running `id`
+    pub fn new(
+        client: Arc<Docker>,
+        id: ContainerId,
```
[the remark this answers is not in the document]
johnyob: The original design was that this `new` method shouldn't be exposed outside of `RunnableImage::start`

johnyob:
Additionally in future, I suggest we stick to 1 linear task per a PR. This single PR partially closes 4 (some support is missing here and there). This will not only make task tracking easier (and fits more nicely into linear's workflow), but will make reviews easier

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use crate::network::Network;
+
+pub type ContainerId = String;
+#[derive(Default)]
+struct ContainerInner {
+    pub id: ContainerId,
+    pub name: String,
+    client: Option<Arc<Docker>>,
+    // TODO: network: Option<Network>,
```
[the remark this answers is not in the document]
ryutamago: that task is done so, i've created a differnt task!

crates/jstzd/src/runnable_image/mod.rs
ryutamago: didn't know that i coudl unit tests private methods! thanks !

crates/jstzd/src/image.rs
ryutamago: i'm not sure if we should instantiate a `RunnableImage` with a container name inside `pull_image`. do you mean associated function rather than a method?

crates/jstzd/src/image.rs
ryutamago: yea i just wanted to explicitly make a builder pattern but that makes sense, thanks!

crates/jstzd/src/network.rs
ryutamago: > Ditto, why is this optional? Suggest to make it non-optional

I had to make this `optional` as the AsyncDropper required the `Default` trait. However now it's disabled so can remove it thanks for the suggestion! @zcabter

ryutamago: > Also remove boilerplate using derive for AsyncDrop

is there any example for this? doesn't seem like i can derive using `async_dropper_simple`

crates/jstzd/src/image.rs
```diff
@@ -0,0 +1,160 @@
...
+        if image_exists(self, client.clone()).await {
+            println!("Image: {:?} already exists ", self.image_name());
+            return Ok(());
+        }
+        create_image(self, client.clone()).await
+    }
+}
+
+pub struct DockerImage<'a> {
+    image_name: &'a str,
```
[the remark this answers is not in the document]
ryutamago: i agree it's easier to manage with the struct owning the string if we are using dynamically create strings but i imagine we will be using static string to specify the names/tags in jstzd so I stick to this. let me know what you think :)

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
...
+use bollard::{container::StartContainerOptions, Docker};
+use std::sync::Arc;
+
+use crate::network::Network;
+
+pub type ContainerId = String;
+#[derive(Default)]
+struct ContainerInner {
+    pub id: ContainerId,
+    pub name: String,
```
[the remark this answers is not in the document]
ryutamago: in order to stat a container, the api requires us to pass in a container name.
```
   pub async fn start_container<T>(
        &self,
        container_name: &str,
        options: Option<StartContainerOptions<T>>,
    ) -> Result<(), Error>
```

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub async fn start(&self) -> anyhow::Result<()> {
+        let client = self.client()?;
+        if let Some(state) = client.inspect_container(&self.name, None).await?.state {
+            if let Some(true) = state.running {
+                println!("Container is already running");
+                return Ok(());
+            }
+        }
+        client
+            .start_container(&self.name, None::<StartContainerOptions<String>>)
```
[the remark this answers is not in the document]
ryutamago: The container name should be unique similar to the id, since the `name` is required to create a container (in RunnableImage::create_container), if we just use the `name` we can remove the extra field `id` from the container as well? wdyt

crates/jstzd/src/container.rs
ryutamago: can't use the `derive` feature together w the `simple` feature that enables us to remove the Default trait. I will stick to the simple feature for now

crates/jstzd/src/docker/runnable_image.rs
```diff
@@ -0,0 +1,371 @@
...
+        container_port: ContainerPort,
+    ) -> Self {
+        self.ports.insert(host_port, container_port);
+        self
+    }
+
+    pub async fn create_container(&self, client: Arc<Docker>) -> Result<String> {
+        self.all_container_ports_are_exposed()?;
+        let config = Config::<String> {
+            image: Some(self.image.image_name().to_string()),
```
zcabter, line 121: `image` should refer to the full uri like `<name>:<tag>` otherwise it will always default tag to latest.
Suggest extending `trait Image` with `fn image_uri() -> String`

crates/jstzd/src/container.rs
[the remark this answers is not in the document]
johnyob: `name` is often used to refer to a human readable name that is supplied when creating the container. We want to solely rely on `id` (which is always returned from the docker daemon): https://docs.rs/bollard/latest/bollard/models/struct.ContainerCreateResponse.html#structfield.id

crates/jstzd/src/container.rs
```diff
@@ -0,0 +1,114 @@
+use async_dropper::{AsyncDrop, AsyncDropper};
+use async_trait::async_trait;
+use bollard::{container::StartContainerOptions, Docker};
+use std::sync::Arc;
+
+use crate::network::Network;
+
+pub type ContainerId = String;
+#[derive(Default)]
```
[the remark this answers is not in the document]
johnyob: Re-opening since in my opinion is easier & neater to come up with a cheap `Default` constraint as opposed to a singleton wrapper on `AsyncDropper`. This was we can exploit the existing `derive` functionality

crates/jstzd/src/container.rs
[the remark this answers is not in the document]
johnyob: See discussion above about `id`. Still recommend using `id` over `name`

crates/jstzd/src/container.rs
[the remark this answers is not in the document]
ryutamago: the derive feature rquires us to implement `Eq` and `PartialEq` trait on Docker, which deosn't make much sense so we could stick to the `simple` feature
```
error[E0277]: the trait bound `Docker: std::cmp::Eq` is not satisfied
  --> crates/jstzd/src/docker/network.rs:10:5
   |
7  | #[derive(Debug, Default, PartialEq, Eq, AsyncDrop)]
   |                                     -- in this derive macro expansion
...
10 |     client: Option<Arc<Docker>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::cmp::Eq` is not implemented for `Docker`, which is required by `Option<Arc<Docker>>: std::cmp::Eq`
   |
   = help: the trait `std::cmp::Eq` is implemented for `Option<T>`
   = note: required for `Arc<Docker>` to implement `std::cmp::Eq`
   = note: 1 redundant requirement hidden
   = note: required for `Option<Arc<Docker>>` to implement `std::cmp::Eq`
note: required by a bound in `AssertParamIsEq`
  --> /private/tmp/rust-20240518-7986-rmyaj5/rustc-1.78.0-src/library/core/src/cmp.rs:360:1
   = note: this error originates in the derive macro `Eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

crates/jstzd/src/container.rs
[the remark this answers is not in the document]
ryutamago: done! just stored the id instead :)

crates/jstzd/src/image.rs
ryutamago: put them as a default method on `Image` trait

crates/jstzd/src/docker/runnable_image.rs
ryutamago: done!

zcabter:
> What was the reason/advantage for using rust to generate docker image rather thank just using a Dockerfile and docker-compose?
This work doesn't generate docker images but automates pulling and running them in containers @jasbir-dhillon

jasbir-dhillon:
> > What was the reason/advantage for using rust to generate docker image rather thank just using a Dockerfile and docker-compose?
> 
> This work doesn't generate docker images but automates pulling and running them in containers @jasbir-dhillon. These APIs will ultimately be used to orchestrate the jstzd sandbox environment.

@zcabter Could that have been done by just using a docker-compose file. You would of needed less lines of code and it would have been more readable?

crates/jstzd/src/docker/container.rs
```diff
@@ -1,79 +1,79 @@
...
-        self.0.inner_mut().async_drop().await;
+impl Drop for Container {
+    fn drop(&mut self) {
+        if !self.dropped {
+            // Prevent the copy `this` to drop again
+            self.dropped = true;
+            let mut this = std::mem::take(self);
+            // Prevent the original `self` to drop again
+            self.dropped = true;
+            TokioScope::scope_and_block(|s| {
```
huancheng-trili, line 72: Just a note about `scope_and_block`: this only works with a multi-threaded runtime. Otherwise, the code will panic here. Sample error message:
```
thread '...' panicked at ...:                                                                                                                                                              
can call blocking only when running on the multi-threaded runtime

**Comment by zcabter** [comment]
> > > What was the reason/advantage for using rust to generate docker image rather thank just using a Dockerfile and docker-compose?
> > 
> > 
> > This work doesn't generate docker images but automates pulling and running them in containers @jasbir-dhillon. These APIs will ultimately be used to orchestrate the jstzd sandbox environment.
> 
> @zcabter Could that have been done by just using a docker-compose file. You would of needed less lines of code and it would have been more readable?

It could but we'd still need to write an api over it as Docker is just one of `Task`s target backends. The jstzd orchestrator would still need to work with tasks in a backend agnostic way.

**Comment by zcabter** [comment]
The testing coverage is a bit low partly because integration tests are missing. I know you put a lot of effort into this but I think this PR needs to be broken down into 4 PRs, 1 per issue. Where we are running docker commands, it needs to be 
integration tested. This will make it easier to merge the parts of your code that is ready.

Additionally, I suggest removing the async dropper and custom drop logic for now since it will be re-thought (per https://github.com/jstz-dev/jstz/pull/555#discussion_r1754813401) but keeping the clean up logic.

**Comment by ryutamago** [comment]
> The testing coverage is a bit low partly because integration tests are missing. I know you put a lot of effort into this but I think this PR needs to be broken down into 4 PRs, 1 per issue. Where we are running docker commands, it needs to be integration tested. This will make it easier to merge the parts of your code that is ready.
> 
> Additionally, I suggest removing the async dropper and custom drop logic for now since it will be re-thought (per [#555 (comment)](https://github.com/jstz-dev/jstz/pull/555#discussion_r1754813401)) but keeping the clean up logic.

I've split into three PRs, the first one doesn't require integration test so it could be merged.  The 2 and 3 pr is tested manually and it only does the basic stuff so I personally don't think there's much problem merging it so we can see something is working. I've also removed the drop for now in the 3rd pr. 

1. image / runnable image
2. [pull-image](https://github.com/jstz-dev/jstz/pull/558)
3. [container](https://github.com/jstz-dev/jstz/pull/559)

@johnyob @zcabter

File: crates/jstzd/src/docker/container.rs (hunk 22 above)
Comment (huancheng-trili) [reply crates/jstzd/src/docker/container.rs]: I realised that this is not just for `scope_and_block` but for blocking async in general, e.g. `tokio::task::block_in_place`.

I guess we can keep that assumption for now. For unit tests this might fail since single-threaded runtime is used by default. This can be resolved with `#[tokio::test(flavor = "multi_thread")]`.

File: crates/jstzd/src/docker/runnable_image.rs [hunk 23]
```diff
@@ -0,0 +1,324 @@
...
+    pub fn with_port(
+        mut self,
+        host_port: HostPort,
+        container_port: ContainerPort,
+    ) -> Self {
+        self.ports.insert(host_port, container_port);
+        self
+    }
+
+    pub async fn create_container(&self, client: Arc<Docker>) -> Result<String> {
```
johnyob, line 115: I would move all asssociated logic for `create_container` under a different PR (will help with your testing). 
The task is: https://linear.app/tezos/issue/JSTZ-62/implement-start-for-runnableimage

Cargo.toml
```diff
@@ -12,7 +12,8 @@ members = [
   "crates/jstz_rollup",
   "crates/jstz_sdk",
   "crates/jstz_wpt",
-  "crates/octez"
+  "crates/octez",
+  "crates/jstzd"
```
johnyob, line 16: I don't believe this is necessary anymore if you rebase

Cargo.toml
```diff
@@ -98,6 +99,10 @@ tokio-util = "0.7.10"
 url = "2.4.1"
 urlpattern = "0.2.0"
 wasm-bindgen = "0.2.92"
+bollard = "0.16.1"
+# async-dropper = { version = "0.3.1", features = ["tokio", "simple"] }
+# async-trait = "0.1.82"
+# async-scoped = "0.9.0"
```
johnyob, line 105: Avoid comments in the `Cargo.toml` file

Cargo.toml
```diff
@@ -98,6 +99,10 @@ tokio-util = "0.7.10"
 url = "2.4.1"
 urlpattern = "0.2.0"
 wasm-bindgen = "0.2.92"
+bollard = "0.16.1"
```
johnyob, line 102: nit: dependencies are supposed to be sorted alphabetically

codecov.yml
```diff
@@ -3,6 +3,9 @@ comment:
   layout: "header, files, footer"
   hide_project_coverage: false
 
+ignore:
+  - crates/jstzd/src/main.rs
+
```
johnyob, line 8: Not necessary anymore

ryutamago: done

Cargo.toml
ryutamago: done

Cargo.toml
ryutamago: done

Cargo.toml
ryutamago: done

crates/jstzd/src/docker/runnable_image.rs
ryutamago: done

johnyob: No, this should be a separate PR -- it is a different task to `pull_image`

johnyob (approved):
lgtm 🚀

PR #555, opened 2024-09-10
feat(jstzd): define Task trait and type JoinHandle

huancheng-trili:
# Context

Define the `Task` trait and implement the future wrapper type `JoinHandle`. `Task` represents any task to be run with backends, e.g. docker engine, shell, jstz node.

[JSTZ-53](https://linear.app/tezos/issue/JSTZ-53/implement-the-task-trait)

# Description

Changes in `Cargo.toml` will be updated accordingly after #553 is merged.

# Manually testing the PR

`cargo test`

```sh
$ cd $(git rev-parse --show-toplevel)/crates/jstzd && cargo test
...

running 6 tests
test task::joinhandle::test_joinhandle::wait_err ... ok
test task::joinhandle::test_joinhandle::drop_test ... ok
test task::joinhandle::test_joinhandle::abort_test ... ok
test task::joinhandle::test_joinhandle::wait_ok ... ok
test task::joinhandle::test_joinhandle::future_implementation ... ok
test task::joinhandle::test_joinhandle::signal_send ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.04s

   Doc-tests jstzd

running 1 test
test crates/jstzd/src/task/joinhandle.rs - task::joinhandle::JoinHandle::new (line 77) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.37s

```


crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+        task_group: &Self::TaskSet,
+    ) -> anyhow::Result<()>;
+
+    /// Spawns the task
+    async fn spawn_task(self) -> JoinHandle;
+}
+
+pub struct JoinHandle {
+    task: OnceCell<tokioJoinHandle<()>>,
+    allowed_signal: SignalKind,
```
johnyob, line 33: I think you've misunderstood the purpose of signals here. A signal is used to notify the task of some external action (commonly terminating). 

But instead of abruptly aborting the Tokio task (which doesn't leave any room for graceful termination / cleaning up of resources), its the implementator of the `Task` trait responsibility to ensure that when sent a termination signal, `run_task` returns. 

The intended use would be that the signal would be queued in the `mpsc::Sender`, with the implementator using the `mpsc::Recieve` to handle signals (including termination).

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+        signals: mpsc::Receiver<Signal>,
+        task_group: &Self::TaskSet,
+    ) -> anyhow::Result<()>;
+
+    /// Spawns the task
+    async fn spawn_task(self) -> JoinHandle;
+}
+
+pub struct JoinHandle {
+    task: OnceCell<tokioJoinHandle<()>>,
```
johnyob, line 32: Would `Option` work instead? I don't see anywhere that we're relying on `OnceCell::set`?

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+            ..Default::default()
+        }
+    }
+
+    /// This function will block until the task has exited, returning a success or an error
+    async fn wait(&mut self) -> anyhow::Result<()> {
+        return match self.task.take() {
+            Some(handle) => {
+                let result = handle.await;
+                if result.is_err() {
```
johnyob, line 79: Suggest using an `if let` here to avoid the `unwrap_err`

crates/jstzd/src/task/task.rs
johnyob: Perhaps we should have an `abort` on a configurable timeout?

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+            Some(handle) => {
+                if handle.is_finished() {
+                    Poll::Ready(())
+                } else {
+                    let waker = cx.waker().clone();
+                    tokio::spawn(async move {
+                        tokio::time::sleep(std::time::Duration::from_secs_f32(0.3)).await;
+                        waker.wake();
+                    });
+                    Poll::Pending
```
johnyob, line 126: Instead of creating a new thread to wakeup the waker, recommend using the `.poll` method of the `handle` (it also implements `Future`). Something like the following should work:
```rust
  let pin = Pin::new(handle);
  match pin.poll(cx) {
      Poll::Ready(_) => Poll::Ready(())
      Poll::Pending => Poll::Pending,
  }
  ```

File: crates/jstzd/src/task/task.rs [hunk 5]
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    async fn shutdown(&mut self) {
+        let _ = self.signal(SignalKind::terminate());
+        let _ = self.wait().await;
+    }
+}
+
+impl Future for JoinHandle {
+    type Output = ();
```
johnyob, line 114: Shouldn't we return an `anyhow::Result<()>`? Since `.await` is essentially `.wait`?

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// # tokio_test::block_on(async {
+    /// let handle = JoinHandle::new(
+    ///     tokio::spawn(async move {
+    ///         tokio::time::sleep(std::time::Duration::from_secs(10)).await;
+    ///     }),
+    ///     SignalKind::terminate(),
+    /// );
+    /// # });
+    /// ```
+    pub fn new(task: tokioJoinHandle<()>, signal: SignalKind) -> Self {
```
johnyob, line 66: Not really intended to be part of the API, recommend guarding with `#[cfg(test)]`

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+impl Drop for JoinHandle {
+    fn drop(&mut self) {
+        if !self.dropped {
+            let mut this = JoinHandle::default();
+            std::mem::swap(&mut this, self);
+            this.dropped = true;
+            tokio::spawn(async move { this.async_drop().await });
+        }
+    }
+}
```
johnyob, line 144: I assume this is rather similar to the `AsyncDropper` wrapper from `async_dropper_simple`. I'm more leaning towards using the `derive` feature from `async_drop` instead of the `simple` version in which case we could use `#[derive(AsyncDrop)]`. What do you think? cc @ryutamago @zcabter

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// This is a non-blocking operation
+    fn signal(&self, signal: SignalKind) -> anyhow::Result<()> {
+        if signal != self.allowed_signal && signal != SignalKind::terminate() {
+            return Err(anyhow::anyhow!(
+                "unacceptable signal '{:?}'",
+                signal.as_raw_value()
+            ));
+        }
+        if let Some(handle) = self.task.get() {
+            handle.abort();
```
johnyob, line 102: See comment above re: signals

crates/jstzd/src/task/task.rs
zcabter: Yeah initially I was leaning towards `async_dropper_simple` since it removed the Default requirement as it gets more difficult to define default for large structs and there is a tendency to simply derive non-meaningful defaults makes the code harder to use but it adds invasive boilerplate.

If we use derive, our structs are rather simple (✅ ) and so we just have to be careful to define a meaningful defaults where we can. tldr; derive is fine with me, define meaningful defaults where possible.

crates/jstzd/src/task/task.rs
huancheng-trili: As discussed offline, calling poll here requires a mutable reference to `handle` and that makes things quite complicated here. I made a few attempts but still could not make it work.

crates/jstzd/src/task/task.rs
huancheng-trili: Yes. This is now updated.

crates/jstzd/src/task/task.rs
zcabter: Alternatively, we don't need to implement drop if the tasks don't panic. Instead, have the jstzd orchestrator registers shutdown/clean up hooks for each task it registers. The orchestrator could control shutdown of tasks based on the state of other tasks. If orchestrator itself panics, it calls shutdown in its async dropper which isolates the async dropper boilerplate. This makes the shutdowns more explicit rather than implicitly cascading, its easier to test and easier to read. wdyt? 

@johnyob

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,297 @@
...
+            Err(_) => Poll::Pending,
+        };
+    }
+}
+
+// On drop, send shutdown signal and wait for the task to terminate
+impl Drop for JoinHandle {
+    fn drop(&mut self) {
+        if !self.dropped {
+            let mut this = JoinHandle::default();
```
ryutamago, line 149: shouldn't we set `self.dropped = true` before and after the swap to prevent double drop?

```
if !self.dropped {
   self.dropped = true
   ... 
   std::mem::swap(&mut this, self);

   self.dropped = true;
```

crates/jstzd/src/task/task.rs
```diff
@@ -0,0 +1,297 @@
...
+            std::mem::swap(&mut this, self);
+            this.dropped = true;
+            tokio::spawn(async move { this.async_drop().await });
```
ryutamago, line 152: if the main thread exits before the async_drop completes, the drop will not get executed. the async-dropper-simple library uses the [scope_and_block](https://docs.rs/async-scoped/latest/async_scoped/struct.Scope.html#method.scope_and_block) function instead to  block the current thread until the task ends. perhaps it is a better option ? cc @johnyob  @zcabter

zcabter: I think calling `tokio::spawn` here causes race conditions. `Drop` is synchronous so we should only allow synchronous calls. I suggest something like `tokio::block_in_place`?

huancheng-trili:
I've managed to fix everything. This implementation is a bit annoying, but I think we need it to be like this. Here is the explanation.

We want to enable the following things:

* Users should be able to wait for the task and retrieve the result in the end.
* Users should be able to implement signal handlers inside tasks and send signals to tasks to gracefully shutdown tasks.
* There must be some mechanism for aborting tasks forcefully.
* When a handle is being dropped, the running task should be cancelled.

To achieve these, some new members are added to the `JoinHandle` struct. Now it has the following members:

* `handle`: the tokio task handle of the actually running task in a wrapper struct.
* `signal_tx`: the channel sender that allows users to send signals to the running task. Its receiver counterpart should be injected somewhere in the task defined by users.
* `execution_result`: a mutex that keeps the result of the running task.

Signalling is now achieved through channels. Task creators are supposed to somehow include the receiver in the task definition (and implement signal handling) and pass the sender to the handle. Calling `.signal` basically calls the underlying signal sender to send the given signals.

One main challenge here is moving handle instances. A possible use case of the handles is as follows:
```rs
let handle = JoinHandle::new(...);

// Do something and send signals to the task
tokio::spawn(async move {
    ...
    handle.signal(...);
});

// Abort the task should anything fatal happen
tokio::spawn(async move {
    ...
    handle.abort().await;
});

let result = handle.wait().await;
```

`handle` then needs to be moved, unless copies are passed into the futures. This means the handle needs to implement `Clone` and each copy should hold a reference to the running task handle. On the other hand, we want to cancel the underlying task when a handle instance is being destroyed but not when the copies are destroyed. This means this cancellation needs to be implemented for the underlying handle struct, which is why there is a new wrapper struct `JoinHandleWrapper` that wraps around the actual tokio handles. The main point for this wrapper is that it terminates the running task in its `Drop` implementation. Therefore, when the last copy of our `JoinHandle` struct is destroyed, the underlying task will surely be cancelled but not prematurely when other copies of the handle are dropped.

PR #560, opened 2024-09-12
feat(withdrawal): implement fa withdrawal

zcabter:
# Context
Adds support for direct fa withdraw from external operation `RunFunction` and through smart function call. 

Because this is implemented as smart function call, also closes https://linear.app/tezos/issue/JSTZ-112/fa-withdrawal-smart-function-call for free.

# Description
This design is based on TZIP-029 except that the user initiates the withdrawal by calling the proxy smart function instead of the protocol.

This design has a few benefits over TZIP-029
1. Simpler and symmetric UX - Just like how users don't need to know the ticket details of a deposit, users should not need to know ticket details to withdraw. This information is correctly exposed in proxy smart function only.
2. Simplifies implementation in the protocol
3. Removes 1 level of indirection. It is obvious now that the proxy smart function needs to implement a withdraw bridge to withdraw their tokens

Withdraw and FA withdraw share similarities in how they are validated. The only difference between their request objects is the body, so the `validate_withdraw_request` was refactored with a generic `<T: Deserialize>`.

Doc string comments were added to `TicketTable::add/sub` as the return type is an ambiguous `Amount`

# Manually testing the PR
```
cargo test -p jstz_proto fa_withdraw
```

### DIFF
--- a/crates/jstz_proto/src/executor/fa_withdraw.rs
+++ b/crates/jstz_proto/src/executor/fa_withdraw.rs
@@ -1,0 +1,348 @@
+use crate::context::{
+    account::{Address, Amount},
+    ticket_table::TicketTable,
+};
+
+use crate::{Error, Result};
+use derive_more::{Display, Error, From};
+use jstz_api::http::body::HttpBody;
+use jstz_core::{
+    host::HostRuntime,
+    kv::{outbox::OutboxMessage, Transaction},
+};
+use jstz_crypto::public_key_hash::PublicKeyHash;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use tezos_crypto_rs::hash::ContractKt1Hash;
+use tezos_smart_rollup::{
+    michelson::{
+        ticket::{FA2_1Ticket, TicketHash},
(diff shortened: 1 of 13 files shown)


crates/jstz_proto/src/executor/fa_withdraw.rs
```diff
@@ -0,0 +1,346 @@
...
+    let ticket = ticket_info.to_ticket(amount)?;
+    let outbox_message_id =
+        withdraw_from_ticket_owner(rt, tx, source, &routing_info, amount, ticket)?;
+    Ok(FaWithdrawalReceiptContent {
+        source: source.clone(),
+        outbox_message_id,
+    })
+}
+
+/// Execute the [FaWithdrawal] request atomically. See [fa_withdraw].
```
johnyob, line 165: stylistic nit: But wouldn't it make more sense to have this in an `impl` for `FaWithdraw` 
e.g. `FaWithdraw::execute(rt, tx, source, gas_limit)`

johnyob (approved):
Overall lgtm 🚀 

Small nit on Rust style

PR #563, opened 2024-09-13
feat(jstzd): define Task trait

huancheng-trili:
# Context

Define the Task trait. Task represents any task to be run with backends, e.g. octez node, jstz node, etc.

[JSTZ-115](https://linear.app/tezos/issue/JSTZ-115/define-the-task-trait)

# Description

Define the basic `Task` trait.

The methods:
* `spawn`: The initialiser that takes in the necessary config for this task and runs the task.
* `kill`: Abort the running task.
* `health_check`: Check if the task is still alive.

# Manually testing the PR

`cargo build`

```sh
$ cd $(git rev-parse --show-toplevel)/crates/jstzd && cargo build
   Compiling jstzd v0.1.0-alpha.0 (/Users/huanchengchang/code/jstz/crates/jstzd)
    Finished dev [unoptimized + debuginfo] target(s) in 0.64s
```

PR #564, opened 2024-09-17
feat(jstzd): Implement OctezNode

huancheng-trili:
# Context

# Description

Created struct `OctezNode` that implements `Trait`. Each instance of this struct represents an octez node process.

Note:
* There is no config validation in `OctezNode`. It simply attempts to call the node binary with the given config values and returns errors if there is any. It does not check if there are existing files or directories, e.g. previously created data directory. It is therefore possible that the created process does not use the given config values because it reads the config file from the previously created data directory. Validation is left to the caller. I thought about implementing these checks or removing existing files, but then I thought it might actually lead to unexpected results, so I decided to leave it as is and leave the checks to the caller.
* This implementation leverages the node implementation in crate `octez`. Methods there are all synchronous. An async version is created and referenced in #565.
* Health check is excluded from this PR so that the PR does not get too big. #566 covers the health check.

I used mocks for the `OctezNode` struct and `Child` in unit tests to make test cases simple enough, though side effects are still inevitable, e.g. creating temporary files and directories.

Also re-ordered dependencies in `Cargo.toml`. This was missed out in #563.

# Manually testing the PR

* Unit test: new test cases
```sh
$ cargo test task::octez_node
running 2 tests
test task::octez_node::tests::config_builder ... ok
test task::octez_node::tests::config_builder_default ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s
```
* Manual integration test: added one case
```sh
$ cargo test --test octez_node_test
running 1 test
test octez_node_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 20.59s
```

### DIFF
--- a/crates/jstzd/src/task/mod.rs
+++ b/crates/jstzd/src/task/mod.rs
@@ -1,3 +1,5 @@
+pub mod octez_node;
+
 use anyhow::Result;
 use async_trait::async_trait;
 
--- a/crates/jstzd/src/task/octez_node.rs
+++ b/crates/jstzd/src/task/octez_node.rs
@@ -1,0 +1,225 @@
+use super::Task;
+use anyhow::Result;
+use async_dropper_simple::{AsyncDrop, AsyncDropper};
+use async_trait::async_trait;
+use std::{fs::File, path::PathBuf, sync::Arc};
+use tokio::sync::RwLock;
+
+use octez::OctezNode as InnerOctezNode;
+use std::process::Child;
+
+const DEFAULT_RPC_ENDPOINT: &str = "localhost:8732";
+const DEFAULT_NETWORK: &str = "sandbox";
+const DEFAULT_BINARY_PATH: &str = "octez-node";
+
+#[derive(Clone)]
+pub struct OctezNodeConfig {
+    /// Path to the octez node binary.
+    binary_path: PathBuf,
+    /// Path to the directory where the node keeps data.
+    data_dir: PathBuf,
+    /// Name of the tezos network that the node instance runs on.
+    network: String,
+    /// HTTP endpoint of the node RPC interface, e.g. 'localhost:8732'
+    rpc_endpoint: String,
+    /// Path to the file that keeps octez node logs.
+    log_file: PathBuf,
+    /// Run options for octez node.
+    options: Vec<String>,
+}
+
+#[derive(Default)]
+pub struct OctezNodeConfigBuilder {
+    binary_path: Option<PathBuf>,
+    data_dir: Option<PathBuf>,
+    network: Option<String>,
+    rpc_endpoint: Option<String>,
+    log_file: Option<PathBuf>,
+    options: Option<Vec<String>>,
+}
+
+impl OctezNodeConfigBuilder {
+    pub fn new() -> Self {
+        OctezNodeConfigBuilder::default()
+    }
+
+    /// Sets the path to the octez node binary.
+    pub fn set_binary_path(&mut self, path: &str) -> &mut Self {
+        self.binary_path = Some(PathBuf::from(path));
+        self
+    }
+
+    /// Sets the path to the directory where the node keeps data.
+    pub fn set_data_dir(&mut self, path: &str) -> &mut Self {
+        self.data_dir = Some(PathBuf::from(path));
+        self
+    }
+
+    /// Sets the name of the tezos network that the node instance runs on.
+    pub fn set_network(&mut self, network: &str) -> &mut Self {
+        self.network = Some(network.to_owned());
+        self
+    }
+
+    /// Sets the HTTP(S) endpoint of the node RPC interface, e.g. 'http://localhost:8732'
+    pub fn set_rpc_endpoint(&mut self, endpoint: &str) -> &mut Self {
+        self.rpc_endpoint = Some(endpoint.to_owned());
+        self
+    }
+
+    /// Sets the path to the file that keeps octez node logs.
+    pub fn set_log_file(&mut self, path: &str) -> &mut Self {
+        self.log_file = Some(PathBuf::from(path));
+        self
+    }
+
+    /// Sets run options for octez node.
+    pub fn set_run_options(&mut self, options: &[&str]) -> &mut Self {
+        self.options = Some(
+            options
+                .iter()
+                .map(|v| (*v).to_owned())
+                .collect::<Vec<String>>(),
+        );
+        self
+    }
+
+    /// Builds a config set based on values collected.
+    pub fn build(&mut self) -> Result<OctezNodeConfig> {
+        Ok(OctezNodeConfig {
+            binary_path: self
+                .binary_path
+                .take()
+                .unwrap_or(PathBuf::from(DEFAULT_BINARY_PATH)),
+            data_dir: self
+                .data_dir
+                .take()
+                .unwrap_or(PathBuf::from(tempfile::TempDir::new().unwrap().path())),
+            network: self.network.take().unwrap_or(DEFAULT_NETWORK.to_owned()),
+            rpc_endpoint: self
+                .rpc_endpoint
+                .take()
+                .unwrap_or(DEFAULT_RPC_ENDPOINT.to_owned()),
+            log_file: self.log_file.take().unwrap_or(PathBuf::from(
+                tempfile::NamedTempFile::new().unwrap().path(),
+            )),
+            options: self.options.take().unwrap_or_default(),
+        })
+    }
+}
+
+#[derive(Default)]
+struct ChildWrapper {
+    inner: Option<Child>,
+}
+
+impl ChildWrapper {
+    pub async fn kill(&mut self) -> anyhow::Result<()> {
+        if let Some(mut v) = self.inner.take() {
+            return Ok(v.kill()?);
+        }
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl AsyncDrop for ChildWrapper {
+    async fn async_drop(&mut self) {
+        let _ = self.kill().await;
+    }
+}
+
+#[derive(Default, Clone)]
+pub struct OctezNode {
+    inner: Arc<RwLock<AsyncDropper<ChildWrapper>>>,
+}
+
+#[async_trait]
+impl Task for OctezNode {
+    type Config = OctezNodeConfig;
+
+    /// Spins up the task with the given config.
+    async fn spawn(config: Self::Config) -> Result<Self> {
+        let node = InnerOctezNode {
+            octez_node_bin: Some(config.binary_path),
+            octez_node_dir: config.data_dir,
+        };
+
+        // localhost:8731 refers to the peer http endpoint. This will be removed
+        // when we switch to the async node implementation where this option is removed
+        node.config_init(&config.network, "localhost:8731", &config.rpc_endpoint, 0)?;
+        node.generate_identity()?;
+        Ok(OctezNode {
+            inner: Arc::new(RwLock::new(AsyncDropper::new(ChildWrapper {
+                inner: Some(
+                    node.run(
+                        &File::create(&config.log_file)?,
+                        config
+                            .options
+                            .iter()
+                            .map(|s| s as &str)
+                            .collect::<Vec<&str>>()
+                            .as_slice(),
+                    )?,
+                ),
+            }))),
+        })
+    }
+
+    /// Aborts the running task.
+    async fn kill(&mut self) -> Result<()> {
+        let mut inner = self.inner.write().await;
+        Ok(inner.inner_mut().kill().await?)
+    }
+
+    /// Conducts a health check on the running task.
+    async fn health_check(&self) -> Result<bool> {
+        todo!()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use crate::task::octez_node::{
+        DEFAULT_BINARY_PATH, DEFAULT_NETWORK, DEFAULT_RPC_ENDPOINT,
+    };
+
+    use super::OctezNodeConfigBuilder;
+
+    #[test]
+    fn config_builder() {
+        let config = OctezNodeConfigBuilder::new()
+            .set_binary_path("/tmp/binary")
+            .set_data_dir("/tmp/something")
+            .set_network("network")
+            .set_rpc_endpoint("my_endpoint")
+            .set_log_file("/log_file")
+            .set_run_options(&["foo", "bar"])
+            .build()
+            .unwrap();
+        assert_eq!(config.binary_path, PathBuf::from("/tmp/binary"));
+        assert_eq!(config.data_dir, PathBuf::from("/tmp/something"));
+        assert_eq!(config.network, "network".to_owned());
+        assert_eq!(config.rpc_endpoint, "my_endpoint".to_owned());
+        assert_eq!(config.log_file, PathBuf::from("/log_file"));
+        assert_eq!(
+            config.options,
+            Vec::from(["foo".to_owned(), "bar".to_owned()])
+        );
+    }
+
+    #[test]
+    fn config_builder_default() {
+        let config = OctezNodeConfigBuilder::new().build().unwrap();
+        assert_eq!(config.binary_path, PathBuf::from(DEFAULT_BINARY_PATH));
+        // Checks if the default path is a valid one that actually can exist in the file system
+        std::fs::create_dir(config.data_dir).unwrap();
+        assert_eq!(config.network, DEFAULT_NETWORK.to_owned());
+        assert_eq!(config.rpc_endpoint, DEFAULT_RPC_ENDPOINT.to_owned());
+        // Checks if the default path is a valid one that actually can exist in the file system
+        std::fs::File::create(config.log_file).unwrap();
+        assert_eq!(config.options, Vec::<String>::default());
+    }
+}
--- a/crates/jstzd/tests/octez_node_test.rs
+++ b/crates/jstzd/tests/octez_node_test.rs
@@ -1,0 +1,70 @@
+use jstzd::task::{octez_node, Task};
+
+async fn retry<'a, F>(retries: u16, interval_ms: u64, f: impl Fn() -> F) -> bool
+where
+    F: std::future::Future<Output = anyhow::Result<bool>> + Send + 'a,
+{
+    let duration = tokio::time::Duration::from_millis(interval_ms);
+    for _ in 0..retries {
+        tokio::time::sleep(duration).await;
+        if let Ok(v) = f().await {
+            if v {
+                return true;
+            }
+        }
+    }
+    false
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn octez_node_test() {
+    let data_dir = tempfile::tempdir().unwrap();
+    let log_file = tempfile::NamedTempFile::new().unwrap();
+    let port = std::net::TcpListener::bind("127.0.0.1:0")
+        .unwrap()
+        .local_addr()
+        .unwrap()
+        .port();
+    let rpc_endpoint = format!("localhost:{}", port);
+
+    let mut config_builer = octez_node::OctezNodeConfigBuilder::new();
+    config_builer
+        .set_binary_path("octez-node")
+        .set_data_dir(data_dir.path().to_str().unwrap())
+        .set_network("sandbox")
+        .set_rpc_endpoint(&rpc_endpoint)
+        .set_log_file(log_file.path().to_str().unwrap())
+        .set_run_options(&[]);
+    let mut f = octez_node::OctezNode::spawn(config_builer.build().unwrap())
+        .await
+        .unwrap();
+
+    let health_check_endpoint = format!("http://{}/health/ready", rpc_endpoint);
+    // Should be able to hit the endpoint since the node should have been launched
+    let node_ready = retry(10, 1000, || async {
+        let res = reqwest::get(&health_check_endpoint).await;
+        if let Ok(raw_body) = res {
+            let body = raw_body
+                .json::<std::collections::HashMap<String, bool>>()
+                .await
+                .unwrap();
+            return body.get("ready").cloned().ok_or(anyhow::anyhow!(""));
+        }
+        Err(anyhow::anyhow!(""))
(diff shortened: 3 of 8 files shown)


crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
...
+#[cfg(not(test))]
+use octez::OctezNode as InnerOctezNode;
+#[cfg(not(test))]
+use std::process::Child;
+
+pub struct OctezNodeConfig {
+    binary_path: PathBuf,
+    data_dir: PathBuf,
+    network: String,
+    http_endpoint: String,
```
zcabter, line 20: In #565, `http_endpoint` is being used as the p2p endpoint but we shouldn't need to set that. This field can be removed.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
...
+    rpc_endpoint: String,
+    num_connections: u32,
```
zcabter, line 22: For the purposes of IT and sandbox, I don't forsee us needing to set the `num_connections` to anything other than 0.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
...
+    log_file: PathBuf,
+    options: Vec<String>,
+}
```
johnyob, line 25: Some of these options can have sensible defaults. Additionally I think it might be clearer to use a builder pattern to construct the config, as `OctezNodeConfig::new` (without an LSP i.e. in GitHub review) is rather difficult to read / parse

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
+use super::Task;
+use anyhow::Result;
+use async_dropper::AsyncDrop;
+use async_trait::async_trait;
+use std::{fs::File, path::PathBuf, sync::Arc};
+use tokio::{runtime::Handle, sync::RwLock};
+
+#[cfg(test)]
+use tests::{Child, InnerOctezNode};
```
johnyob, line 9: I think it might be nicer to use `mockall` to Mock these types using `mockall_double` to handle the namespacing issues

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
...
+            options: options
+                .iter()
+                .map(|v| (*v).to_owned())
+                .collect::<Vec<String>>(),
+        }
+    }
+}
+
+#[derive(Default)]
+struct ChildWrapper {
```
johnyob, line 56: Isn't this identical to the `AsyncDropper<T>` from `async_dropper_simple`?

crates/jstzd/src/task/octez_node.rs
johnyob: Additionally I'm not convinced that we want to be mocking everything here. Some things are useful interfaces to mock, but the CI environment should be able to start/stop an octez-node and this should be tested as part of this PR

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,240 @@
...
+#[async_trait]
+impl AsyncDrop for ChildWrapper {
+    async fn async_drop(&mut self) {
+        let _ = self.kill().await;
+    }
+}
+
+#[derive(Default)]
+pub struct OctezNode {
+    inner: Arc<RwLock<AsyncDropper<ChildWrapper>>>,
```
zcabter, line 75: Hmm the inner child instance is never shared outside of OctezNode. Why does it need to be wrapped in `Arc<..>`?

zcabter: I'm also confused as to why it needs to be wrapped in the `RwLock<>` if the instance is never shared?

crates/jstzd/src/task/octez_node.rs
huancheng-trili: I do think that these are necessary since we definitely don't want each test to spin up a node instance. I've added one integration test to ensure that everything works as expected. This should be sufficient after #567 is merged.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Yes. I've replaced this with `AsyncDropper`.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: At this moment it's indeed not very useful. I'm thinking about the case where one might want to clone the node, e.g.
```rs
let node = octez_node::OctezNode::spawn(...).await;
let copy = node.clone();
let trap = tokio::spawn(async {
    ...
    if (...) {
        node.kill().await;
    }
});
// continue with other logic
let _ = node.health_check().await;
```

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Removed

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Removed

crates/jstzd/src/task/octez_node.rs
zcabter: Ah ok

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,264 @@
...
+
+pub struct OctezNodeConfig {
+    binary_path: PathBuf,
+    data_dir: PathBuf,
+    network: String,
+    http_endpoint: String,
+    rpc_endpoint: String,
+    num_connections: u32,
+    log_file: PathBuf,
+    options: Vec<String>,
```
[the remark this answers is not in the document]
huancheng-trili: I'll do this after #565 is merged so that I don't need to convert options to strings so that they can be passed into the current octez node implementation.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: I've added one config builder. `options` is left as a vector for now. That will be addressed in a separate PR along with updating the underlying async octez node interface so that now I don't need to convert options into strings.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,294 @@
...
+                .unwrap_or(PathBuf::from("/tmp/.octez-node")),
+            network: self.network.take().unwrap_or("sandbox".to_owned()),
+            rpc_endpoint: self
+                .rpc_endpoint
+                .take()
+                .unwrap_or("localhost:8732".to_owned()),
+            log_file: self
+                .log_file
+                .take()
+                .unwrap_or(PathBuf::from("/tmp/octez-node.log")),
```
zcabter, line 94: Perhaps we should use (tempfile crate)[https://docs.rs/tempfile/latest/tempfile/] for creating unique temps and cleaning them up on drop.

huancheng-trili: Yup sure. I think we don't need to clean them up for now as the OS will do this at some point. I'm using `NamedTempFile` here so that we can keep track of the path.

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,53 @@
...
+        .set_log_file(log_path.to_str().unwrap())
+        .set_run_options(&[]);
+    let mut f = octez_node::OctezNode::spawn(config_builer.build().unwrap())
+        .await
+        .unwrap();
+
+    let health_check_endpoint = format!("http://{}/health/ready", rpc_endpoint);
+    // Should be able to hit the endpoint since the node should have been launched
+    tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
+    let res = reqwest::get(&health_check_endpoint).await;
```
zcabter, line 39: Maybe instead we just poll with 10 sec timeout until health check is returns success?

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,53 @@
+#![cfg(not(target_os = "macos"))]
+// This test is disabled for macos because octez binary is not available
```
zcabter, line 2: could we default to look for binaries on `PATH`? Then at least its runnable locally for us.

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,53 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn octez_node_test() {
+    let tmp_dir = std::env::temp_dir();
+    let ts = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
+    let data_dir_path = tmp_dir.join(format!("jstzd-octez-node-test-{}", ts));
+    std::fs::create_dir(&data_dir_path).unwrap();
+    let log_path = tmp_dir.join(format!("jstzd-octez-node-test-{}-log.txt", ts));
```
zcabter, line 18: Use `tempfile` here as well

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,53 @@
...
+    let rpc_endpoint = format!(
+        "localhost:{}",
+        10000 + rand::thread_rng().gen_range(0..10000)
+    );
```
zcabter, line 22: There's a tiny chance you might land on a port in use. Could we use TcpListener to grab an os allocated  port instead? 
```
let port = TcpListener::bind("127.0.0.1:0")
        .unwrap()
        .local_addr()
        .unwrap()
        .port();
```

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,53 @@
+#![cfg(not(target_os = "macos"))]
+// This test is disabled for macos because octez binary is not available
+
+use jstzd::task::{octez_node, Task};
+use rand::Rng;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+#[ignore = "ignore for now until octez binaries in CI are set up properly"]
+#[tokio::test(flavor = "multi_thread")]
+async fn octez_node_test() {
```
zcabter, line 10: things like creating tmp dir with prefix and generating random ports will likely be reused. I think it would make sense to factor into a common module.

crates/jstzd/tests/octez_node_test.rs
huancheng-trili: Updated

crates/jstzd/src/task/octez_node.rs
zcabter: I tried to depend on mockall but it is only supported for rust v1.74 and above despite what the doc says, at least the latest few releases of the lib.

crates/jstzd/tests/octez_node_test.rs
huancheng-trili: This line is removed since the test is ignored now anyway.

crates/jstzd/tests/octez_node_test.rs
huancheng-trili: Indeed, but let's do this later on when we are reusing these pieces of code.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,293 @@
...
+            }
+            File::create(self.file_path.as_ref().unwrap()).unwrap();
+            Ok(())
+        }
+    }
+
+    pub struct InnerOctezNode {
+        pub octez_node_bin: Option<PathBuf>,
+        pub octez_node_dir: PathBuf,
+    }
```
zcabter, line 195: Here, you are mocking Child to test that kill command succeeds/fails when octez node is (not)running. But I think you could test this with no overhead as part of `octez_node_test` in the integration tests  I'm not a fan of mocking the Child and InnerOctezNode in this way.

zcabter: Additionally, you could also expose is_running on `ChildWrapper` to check if the process is running and bubble that to the top. 

```
pub async fn is_running(&mut self) -> anyhow::Result<bool> {
        self.inner
            .as_mut()
            .map(|child| {
                let exit_status = child.try_wait()?;
                Ok(exit_status.is_none())
            })
            .unwrap_or(Ok(false))
    }
```

huancheng-trili: > But I think you could test this with no overhead as part of octez_node_test in the integration tests I'm not a fan of mocking the Child and InnerOctezNode in this way.

Did you miss out a few words?

I feel that having too many external dependencies, e.g. child processes, makes unit tests cumbersome and possibly more flakey. `kill` is covered by the integration test as well. I agree that the mocks are not really meaningful here. We can surely remove them if testing with them don't really make sense, but then there is a check for test coverage, unless we decide to skip it for this. Setting up things like signals to kill processes in unit tests to mimic different scenarios and touch the branches sounds too much to me.

That `is_running` method helps remove the "signal file" thing in these test cases. I've added that and now these tests completely run in memory.

zcabter: I'm in favour of removing the mocks. Integration tests will contribute to the test coverage

huancheng-trili: Okay. I've removed the mocks and unit test cases. I've added new cases to cover the config builder.

crates/jstzd/tests/octez_node_test.rs
zcabter: I don't see any changes. Did you forget to push the change?

huancheng-trili: No I have pushed it

zcabter: Oops was looking at outdated!

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,63 @@
...
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+        let res = reqwest::get(&health_check_endpoint).await;
+        // Should get an error since the node should have been terminated
+        if let Err(e) = res {
+            node_destroyed = e.to_string().contains("Connection refused");
+            if node_destroyed {
+                break;
+            }
+        }
+    }
```
zcabter, line 61: The polling can be factored out - DRY

huancheng-trili: This will be replaced by the health check in #566.

crates/jstzd/tests/octez_node_test.rs
zcabter: Ok, tracked here: https://linear.app/tezos/issue/JSTZ-151/factor-out-common-utility-functions

crates/jstzd/tests/octez_node_test.rs
zcabter: Ah yes, sure. Then let's just factor out the retries? 
`pub fn retry(f: Fn() -> Option<T>, retries: u16, interval_ms: u32) -> Result<T> `

huancheng-trili: Sorry I forgot that the retry part will not be replaced. I don't think this is that helpful here at this moment, but I've moved the retry logic out anyway.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,215 @@
...
+                .unwrap_or(PathBuf::from("octez-node")),
+            data_dir: self
+                .data_dir
+                .take()
+                .unwrap_or(PathBuf::from(tempfile::TempDir::new().unwrap().path())),
+            network: self.network.take().unwrap_or("sandbox".to_owned()),
+            rpc_endpoint: self
+                .rpc_endpoint
+                .take()
+                .unwrap_or("localhost:8732".to_owned()),
```
zcabter, line 98: Could you extract "localhost:8732" into a `const`?

crates/jstzd/src/task/octez_node.rs
zcabter: Current implementation looks good to me. Resolving.

zcabter:
PR lgtm barring minor nit and commits that need squashing.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Done. I also moved `sandbox` and `octez-node` into constants.

crates/jstzd/src/task/octez_node.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /// Builds a config set based on values collected.
+    pub fn build(&mut self) -> Result<OctezNodeConfig> {
+        Ok(OctezNodeConfig {
+            binary_path: self
+                .binary_path
+                .take()
+                .unwrap_or(PathBuf::from(DEFAULT_BINARY_PATH)),
+            data_dir: self
```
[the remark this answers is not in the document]
huancheng-trili: I left that to the binary. If anything is invalid, the subprocess will return an error and the node implementation simply surface that. I don't think it makes sense to repeat too much validation in jstzd when the binaries can handle that for us.

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,77 @@
+use jstzd::task::{octez_node, Task};
+
+async fn retry<'a>(
+    retries: u16,
+    interval_ms: u64,
+    f: &(dyn Fn() -> std::pin::Pin<
+        Box<dyn std::future::Future<Output = anyhow::Result<bool>> + Send + 'a>,
+    > + Sync),
+) -> bool {
```
zcabter, line 9:
```suggestion
async fn retry<'a, F>(retries: u16, interval_ms: u64, f: impl Fn() -> F) -> bool
where
    F: Future<Output = anyhow::Result<bool>> + Send + 'a,
{
```

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,77 @@
...
+        Box::pin(async {
+            let res = reqwest::get(&health_check_endpoint).await;
+            if let Ok(raw_body) = res {
+                let body = raw_body
+                    .json::<std::collections::HashMap<String, bool>>()
+                    .await
+                    .unwrap();
+                return body.get("ready").cloned().ok_or(anyhow::anyhow!(""));
+            }
+            Err(anyhow::anyhow!(""))
```
zcabter, line 57:
```suggestion
     let node_ready = retry(10, 1000, || async {
        let res = reqwest::get(&health_check_endpoint).await;
        if let Ok(raw_body) = res {
            let body = raw_body
                .json::<std::collections::HashMap<String, bool>>()
                .await
                .unwrap();
            return body.get("ready").cloned().ok_or(anyhow::anyhow!(""));
        }
        Err(anyhow::anyhow!(""))
```

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,77 @@
...
+        })
+    })
+    .await;
+    assert!(node_ready);
+
+    let _ = f.kill().await;
+    // Wait for the process to shutdown entirely
+    let node_destroyed = retry(10, 1000, &|| {
+        Box::pin(async {
```
zcabter, line 66:
```suggestion
    let node_destroyed = retry(10, 1000, || {
        async {
```

crates/jstzd/tests/octez_node_test.rs
zcabter: Prefer impl Trait over dyn Trait for compile time polymorphism.

crates/jstzd/tests/octez_node_test.rs
```diff
@@ -0,0 +1,77 @@
...
+            let res = reqwest::get(&health_check_endpoint).await;
+            // Should get an error since the node should have been terminated
+            if let Err(e) = res {
+                return Ok(e.to_string().contains("Connection refused"));
+            }
+            Err(anyhow::anyhow!(""))
+        })
+    })
+    .await;
```
zcabter, line 75: With the change to ref taking an `impl Fn` closure, there shouldn't be a need to pass the closure as `&` or pin the async func. Should be able to do the following I think: 
```
retry(10, 1000, || async {
 ... 
});
```

crates/jstzd/tests/octez_node_test.rs
huancheng-trili: Updated

crates/jstzd/src/task/octez_node.rs
zcabter: Changes LGTM

zcabter (approved):
LGTM

crates/jstzd/src/task/mod.rs
johnyob, line 1: Is `task` the best module to place Octez node (and all the other octez process interfaces?)
I was thinking would probably be best suited in the `octez` crate?

huancheng-trili: Correct me if I'm wrong, but from what I understand now, the `octez` crate contains code that actually interact with binaries. The `OctezNode` here will be one of the components in jstzd and it happens to sort of control the node, but since it's still not a binary wrapper and will probably not be shared with other crates, I think it's okay to keep it here.

On the other hand, the `AsyncOctezNode` implementation in #565 should perhaps be moved to the `octez` crate as it will replace the current `OctezNode` implementation in the `octez` crate at some point.

johnyob: I don't think it should just be a wrapper. It should have more complex behaviour (such as containing some parts of the configuration builder there). 

I think a good rule of thumb would be that the logic for octez in `jstzd` should only relate to the steps necessary for setting up / the creation of the sandbox, keeping as much as possible in `octez`. Additionally I think the module struct of jstzd should match that of our linear milestones / components e.g. `src/octez/{mod.rs, node.rs, baker.rs, rollup_node.rs, etc}`, `src/jstz/{mod.rs, kernel.rs, bridge.rs, node.rs}`, src/task.rs`, etc

huancheng-trili: I see. I'll move the pieces that are not really specifically for jstzd to `octez` and restructure the module.

PR #565, opened 2024-09-17
feat(jstzd): Use async octez node

huancheng-trili:
# Context

Part of JSTZ-116.

# Description

Create an async version of `OctezNode` from crate `octez`.

This piece of code will replace the current `OctezNode` struct later on. For now it's kept in a separate module in order not to affect other pieces of code that are referencing `OctezNode`.

# Manually testing the PR

* Integration test: ran the code in the file `crate/jstzd/tests/octez_node_test.rs` and everything worked as expected.

### DIFF
--- a/crates/octez/src/async_node.rs
+++ b/crates/octez/src/async_node.rs
@@ -1,0 +1,73 @@
+use std::{fs::File, path::PathBuf, process::Stdio};
+
+use tokio::process::{Child, Command};
+
+use super::path_or_default;
+use anyhow::Result;
+
+pub struct AsyncOctezNode {
+    /// Path to the octez-node binary
+    /// If None, the binary will inside PATH will be used
+    pub octez_node_bin: Option<PathBuf>,
+    /// Path to the octez-node directory
+    pub octez_node_dir: PathBuf,
+}
+
+impl AsyncOctezNode {
+    fn command(&self) -> Command {
+        Command::new(path_or_default(self.octez_node_bin.as_ref(), "octez-node"))
+    }
+
+    pub async fn config_init(
+        &self,
+        network: &str,
+        rpc_endpoint: &str,
+        num_connections: u32,
+    ) -> Result<Child> {
+        Ok(self
+            .command()
+            .args([
+                "config",
+                "init",
+                "--network",
+                network,
+                "--data-dir",
+                self.octez_node_dir.to_str().expect("Invalid path"),
+                "--rpc-addr",
+                rpc_endpoint,
+                "--connections",
+                num_connections.to_string().as_str(),
+            ])
+            .spawn()?)
+    }
+
+    pub async fn generate_identity(&self) -> Result<Child> {
+        Ok(self
+            .command()
+            .args([
+                "identity",
+                "generate",
+                "0",
+                "--data-dir",
+                self.octez_node_dir.to_str().expect("Invalid path"),
+            ])
+            .spawn()?)
+    }
+
+    pub async fn run(&self, log_file: &File, options: &[&str]) -> Result<Child> {
+        let mut command = self.command();
+
+        command
+            .args([
+                "run",
+                "--data-dir",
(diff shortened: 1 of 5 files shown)


crates/jstzd/src/task/inner_octez_node.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> Result<()> {
+        run_command(self.command().args([
+            "config",
+            "init",
+            "--network",
+            network,
+            "--data-dir",
+            self.octez_node_dir.to_str().expect("Invalid path"),
+            "--net-addr",
+            http_endpoint,
```
zcabter, line 54: `--net-addr` controls the p2p connection point, not the http endpoint. Can be removed in favour of the default.

crates/jstzd/src/task/inner_octez_node.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub async fn config_init(
+        &self,
+        network: &str,
+        http_endpoint: &str,
+        rpc_endpoint: &str,
+        num_connections: u32,
+    ) -> Result<()> {
+        run_command(self.command().args([
```
zcabter, line 46: The init is missing `--synchronisation-threshold 0` and `--no-bootstrap-peers`

crates/jstzd/src/task/inner_octez_node.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    pub async fn run(&self, log_file: &File, options: &[&str]) -> Result<Child> {
+        let mut command = self.command();
+
+        command
+            .args([
+                "run",
+                "--data-dir",
+                self.octez_node_dir.to_str().expect("Invalid path"),
+                "--singleprocess",
```
zcabter, line 71: I think we should just run validation in a separate process to avoid unresponsive node hiccups.

huancheng-trili:
Removed unit test mocks as in #564.

crates/octez/src/async_node.rs
```diff
@@ -0,0 +1,83 @@
...
+    if !output.status.success() {
+        return Err(anyhow!(
+            "Command {:?} failed:\n {}",
+            command,
+            String::from_utf8_lossy(&output.stderr)
+        ));
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
huancheng-trili: I've just removed this since it doesn't do much if it does not wait for commands. We can collect all `command.wait()` later on and refactor.

crates/octez/src/async_node.rs
```diff
@@ -0,0 +1,73 @@
...
+                "run",
+                "--data-dir",
+                self.octez_node_dir.to_str().expect("Invalid path"),
+                "--singleprocess",
+            ])
+            .args(options)
+            .stdout(Stdio::from(log_file.try_clone()?))
+            .stderr(Stdio::from(log_file.try_clone()?));
+
+        Ok(command.spawn()?)
```
ryutamago, line 71: i think we should check the status of the process, unleess exit code is 0, throw 
```
    let status = command.spawn()?.wait().await?;
        match status.code() {
            Some(0) => Ok(()),
            _ => bail!("failed to generate keys for {}", alias),
        }
```

ryutamago (approved):
LGTM, just the error handling of the process and good to go !

crates/octez/src/async_node.rs
huancheng-trili: I don't think this is needed here because `run` returns the only the child process. Whoever calls this can implement what you described. Also I think it's better to keep this behaviour the same as what is done in the synchronous node implementation now as eventually this async node will replace it and I don't want to add more logic to all references to accommodate this behaviour change when that takes place. We can do that afterwards if it's necessary.

ryutamago: oh i meant to comment on the `generate identity` / `config init` as it could fail to init the config file if it already etc

PR #566, opened 2024-09-17
feat(jstzd): Implement health check for OctezNode

huancheng-trili:
# Context

# Description

Implement health check for OctezNode.

# Manually testing the PR

* Unit test: more test cases
```sh
$ cargo test task::octez_node
running 6 tests
test task::octez_node::tests::health_check_server_error ... ok
test task::octez_node::tests::kill_should_fail ... ok
test task::octez_node::tests::health_check_json_error ... ok
test task::octez_node::tests::health_check_body_error ... ok
test task::octez_node::tests::kill_should_succeed ... ok
test task::octez_node::tests::health_check_ready ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.02s

     Running unittests src/main.rs (/code/target/debug/deps/jstzd-9e3af1ab7d28b9a1)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```
* Manual integration test: ran the following code locally
```rs
use jstzd::task::{octez_node, Task};

#[tokio::main]
async fn main() {
    let mut f = octez_node::OctezNode::spawn(octez_node::OctezNodeConfig::new(
        "/usr/local/bin/octez-node",
        "/tmp/something",
        "sandbox",
        "localhost:8733",
        "localhost:8732",
        0,
        "/tmp/log",
        &[],
    ))
    .await
    .unwrap();
    tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
    let r = f.health_check().await.unwrap();
    println!("{r}"); // returned `true`
    let _ = f.kill().await;
    tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
}
```
and confirmed that the node process was created after `spawn` and before `kill` was called.


crates/jstzd/src/task/octez_node.rs
```diff
@@ -174,7 +176,22 @@ impl Task for OctezNode {
...
+        // Returns whether or not the node is ready to answer to requests.
+        // https://gitlab.com/tezos/tezos/-/raw/2e84c439c25c4d9b363127a6685868e223877034/docs/api/rpc-openapi.json
+        let res =
+            reqwest::get(format!("http://{}/health/ready", &self.config.rpc_endpoint))
+                .await;
+        if res.is_err() {
+            return Ok(false);
+        }
+        let body = res
+            .unwrap()
```
ryutamago, line 188: perhaps beter to return error with `?`

ryutamago (approved):

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Here I don't think we should return the error. The cases I imagined here are like before the RPC server is even launched. These requests would then fail with something like connection refused errors, and these should be considered as "the node not being ready yet" rather than severe errors. The `json::` case, on the other hand, is where errors should be returned since if there is any error in de-serialisation, it might indicate that the implementation here is outdated, e.g. we did not catch the changes in RPC, and we would need to fix that.

PR #570, opened 2024-09-23
chore(nix): add octez distribution to nix shell and integration tests

johnyob:
# Context

We will require octez for our integration tests for jstzd

# Description

This PR simply adds `octez-v21` using the Nix flake from https://gitlab.com/tezos/tezos/-/blob/master/flake.nix. 
Unfortunately, this will take along time to build, but the dependency will rarely change + once we have a Mac CI runner, we can use that as a cache. 

This PR includes two patches to the `tezos/tezos` flake:
- The first patches the `/bin/bash` issue and the `/usr/bin/env bash` shebang issue for `build.sh` in `src/rust_deps`
- The second patches the cargo dependencies for `src/rust_deps`, prefetching them using `importCargoLock`. Without this, the build fails as the `build.sh` script invokes `cargo` which tries to download the dependencies from `crates.io`. 

# Manually testing the PR

To execute the integration test that uses octez in a Nix environment, run:
```sh
nix build -L -v --print-build-logs .#checks.aarch64-darwin.cargo-test-int
```

### DIFF
--- a/crates/jstzd/tests/dummy.rs
+++ b/crates/jstzd/tests/dummy.rs
@@ -1,6 +1,17 @@
 use jstzd::main;
+use tokio::process::Command;
 
 #[tokio::test]
 async fn test_main() {
     main().await.unwrap();
+}
+
+#[tokio::test]
+async fn test_octez_client() {
+    let output = Command::new("octez-client")
+        .arg("--version")
+        .output()
+        .await
+        .unwrap();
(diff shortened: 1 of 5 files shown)


zcabter:
lgtm

zcabter (approved):

zcabter:
Fixup needs squashing

PR #571, opened 2024-09-24
feat(jstzd): Implement Endpoint

huancheng-trili:
# Context

[JSTZ-130](https://linear.app/tezos/issue/JSTZ-130/implement-endpoint)

# Description

Implemented struct `Endpoint` that represents an octez node RPC endpoint.

Note that the line `#[allow(dead_code)]` is added because the struct fields are not referenced anywhere else for now.

# Manually testing the PR

* Unit test: added test cases

### DIFF
--- a/crates/jstzd/src/task/endpoint.rs
+++ b/crates/jstzd/src/task/endpoint.rs
@@ -1,0 +1,86 @@
+#![allow(dead_code)]
+use http::{uri::Scheme, Uri};
+
+#[derive(Debug, Clone)]
+pub struct Endpoint {
+    scheme: String,
+    host: String,
+    port: u16,
+}
+
+impl Endpoint {
+    pub fn localhost(port: u16) -> Self {
+        Endpoint {
+            scheme: "http".to_owned(),
+            host: "localhost".to_owned(),
+            port,
+        }
+    }
+}
+
+impl TryFrom<Uri> for Endpoint {
+    type Error = anyhow::Error;
+    fn try_from(value: Uri) -> Result<Self, Self::Error> {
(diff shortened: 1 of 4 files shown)


crates/jstzd/src/task/endpoint.rs
```diff
@@ -0,0 +1,86 @@
+#![allow(dead_code)]
+use http::{uri::Scheme, Uri};
+
+#[derive(Debug, Clone)]
```
ryutamago, line 4: nit:
if we implement `Eq, PartialEq` we can easily assert_eq in the test

crates/jstzd/src/task/endpoint.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct Endpoint {
+    scheme: String,
+    host: String,
+    port: u16,
+}
+
+impl Endpoint {
```
ryutamago, line 11: i think we can use this struct for OctezNode too?

huancheng-trili: Yes, but I don't think it's that useful for now since the endpoint eventually gets transformed into a string anyway. We might update `AsyncOctezNode::config_init` so that it consumes `Endpoint` instances, but I would say we do this after other pieces of code are changed to use this async node implementation. If it's really necessary, we do it then.

zcabter:
This PR seems independent from the OctezNode work? Perhaps it could target `main` instead. The PR looks quite small so we can get this in quickly to unblock https://github.com/jstz-dev/jstz/pull/572

zcabter (approved):

PR #572, opened 2024-09-24
feat(jstzd): add octez client

ryutamago:
# Context

[Octez client](https://linear.app/tezos/issue/JSTZ-126/implement-octezclient)

# Description

Add constructor for octez client
Note: the actual octez-client binary sets `base-dir` to `$HOME/.tezos-client` if no values are provided. i followed the spec and ignored this pattern and instead create a tempdir if the base-dir isn't specified.  

# Manually testing the PR

```
6 unit tests:
cargo test --package jstzd --lib -- task::octez_client::test
```

### DIFF
--- a/crates/jstzd/src/task/octez_client.rs
+++ b/crates/jstzd/src/task/octez_client.rs
@@ -1,10 +1,11 @@
-use super::{endpoint::Endpoint, octez_node::DEFAULT_RPC_ENDPOINT};
+use super::{directory::Directory, endpoint::Endpoint, octez_node::DEFAULT_RPC_ENDPOINT};
 use anyhow::{bail, Result};
 use http::Uri;
 use std::{path::PathBuf, str::FromStr};
-use tempfile::{tempdir, TempDir};
+use tempfile::tempdir;
 
 const DEFAULT_BINARY_PATH: &str = "octez-client";
+
 #[derive(Default)]
 pub struct OctezClientBuilder {
     // if None, use the binary in $PATH
@@ -46,13 +47,13 @@ impl OctezClientBuilder {
 
     pub fn build(self) -> Result<OctezClient> {
         self.validate_binary_path()?;
-        self.validate_base_dir()?;
         let node_default_endpoint = format!("http://{}", DEFAULT_RPC_ENDPOINT);
         Ok(OctezClient {
             binary_path: self.binary_path.unwrap_or(DEFAULT_BINARY_PATH.into()),
-            base_dir: self
-                .base_dir
-                .map_or(Directory::TempDir(tempdir()?), Directory::Path),
+            base_dir: match self.base_dir {
+                Some(path_buf) => Directory::try_from(path_buf)?,
+                None => Directory::TempDir(tempdir()?),
+            },
             endpoint: self
                 .endpoint
                 .unwrap_or(Endpoint::try_from(Uri::from_str(&node_default_endpoint)?)?),
@@ -71,26 +72,10 @@ impl OctezClientBuilder {
         }
         Ok(())
     }
-
-    fn validate_base_dir(&self) -> Result<()> {
-        if let Some(base_dir) = &self.base_dir {
-            if !base_dir.exists() {
-                bail!("Base directory does not exist");
-            }
-            if !base_dir.is_dir() {
-                bail!("Base directory is not a directory");
-            }
-        }
-        Ok(())
-    }
-}
-
-enum Directory {
-    TempDir(TempDir),
-    Path(PathBuf),
 }
 
 #[allow(dead_code)]
+#[derive(Debug)]
 pub struct OctezClient {
     binary_path: PathBuf,
     base_dir: Directory,
@@ -100,7 +85,7 @@ pub struct OctezClient {
 
 #[cfg(test)]
 mod test {
-    use tempfile::NamedTempFile;
+    use tempfile::{NamedTempFile, TempDir};
 
     use super::*;
     #[test]
@@ -162,8 +147,7 @@ mod test {
         let octez_client = OctezClientBuilder::new()
             .set_base_dir(dir_path.clone())
             .build();
-        assert!(octez_client
-            .is_err_and(|e| &e.to_string() == "Base directory does not exist"));
+        octez_client.expect_err("Should fail when base dir does not exist ");
     }
 
     #[test]
@@ -173,8 +157,8 @@ mod test {
         let octez_client = OctezClientBuilder::new()
             .set_base_dir(invalid_dir_path.clone())
             .build();
-        assert!(octez_client
-            .is_err_and(|e| &e.to_string() == "Base directory is not a directory"));
+
+        octez_client.expect_err("Should fail when base dir is not a directory");
     }
 
     #[test]
@@ -190,6 +174,7 @@ mod test {
             octez_client.is_err_and(|e| &e.to_string() == "Binary path does not exist")
         );
     }
+
     #[test]
     fn validates_binary_path_is_file() {
         let temp_dir = TempDir::new().unwrap();
--- a/crates/jstzd/src/task/octez_client.rs
+++ b/crates/jstzd/src/task/octez_client.rs
@@ -1,0 +1,204 @@
+use super::{endpoint::Endpoint, octez_node::DEFAULT_RPC_ENDPOINT};
+use anyhow::{bail, Result};
+use http::Uri;
+use std::{path::PathBuf, str::FromStr};
+use tempfile::{tempdir, TempDir};
+
+const DEFAULT_BINARY_PATH: &str = "octez-client";
+#[derive(Default)]
+pub struct OctezClientBuilder {
+    // if None, use the binary in $PATH
+    binary_path: Option<PathBuf>,
+    // if None, use temp directory
+    base_dir: Option<PathBuf>,
+    // if None, use localhost:8732 (the default endpoint for octez-node)
+    endpoint: Option<Endpoint>,
+    disable_unsafe_disclaimer: bool,
+}
+
+impl OctezClientBuilder {
+    pub fn new() -> Self {
+        OctezClientBuilder::default()
+    }
+
+    pub fn set_binary_path(mut self, binary_path: PathBuf) -> Self {
+        self.binary_path = Some(binary_path);
+        self
+    }
+
+    pub fn set_base_dir(mut self, base_dir: PathBuf) -> Self {
+        self.base_dir = Some(base_dir);
+        self
+    }
+
+    pub fn set_endpoint(mut self, endpoint: Endpoint) -> Self {
+        self.endpoint = Some(endpoint);
+        self
+    }
+
+    pub fn set_disable_unsafe_disclaimer(
+        mut self,
+        disable_unsafe_disclaimer: bool,
+    ) -> Self {
+        self.disable_unsafe_disclaimer = disable_unsafe_disclaimer;
+        self
+    }
+
+    pub fn build(self) -> Result<OctezClient> {
+        self.validate_binary_path()?;
+        self.validate_base_dir()?;
+        let node_default_endpoint = format!("http://{}", DEFAULT_RPC_ENDPOINT);
+        Ok(OctezClient {
+            binary_path: self.binary_path.unwrap_or(DEFAULT_BINARY_PATH.into()),
+            base_dir: self
+                .base_dir
+                .map_or(Directory::TempDir(tempdir()?), Directory::Path),
+            endpoint: self
+                .endpoint
+                .unwrap_or(Endpoint::try_from(Uri::from_str(&node_default_endpoint)?)?),
+            disable_unsafe_disclaimer: self.disable_unsafe_disclaimer,
+        })
+    }
+
+    fn validate_binary_path(&self) -> Result<()> {
+        if let Some(binary_path) = &self.binary_path {
+            if !binary_path.exists() {
+                bail!("Binary path does not exist");
+            }
+            if !binary_path.is_file() {
+                bail!("Binary path is not a file");
+            }
+        }
+        Ok(())
+    }
+
+    fn validate_base_dir(&self) -> Result<()> {
+        if let Some(base_dir) = &self.base_dir {
+            if !base_dir.exists() {
+                bail!("Base directory does not exist");
+            }
+            if !base_dir.is_dir() {
+                bail!("Base directory is not a directory");
+            }
+        }
+        Ok(())
+    }
+}
+
+enum Directory {
+    TempDir(TempDir),
+    Path(PathBuf),
+}
+
+#[allow(dead_code)]
+pub struct OctezClient {
+    binary_path: PathBuf,
+    base_dir: Directory,
+    endpoint: Endpoint,
+    disable_unsafe_disclaimer: bool,
+}
+
+#[cfg(test)]
+mod test {
+    use tempfile::NamedTempFile;
+
+    use super::*;
+    #[test]
+    fn builds_default_octez_client() {
+        let octez_client = OctezClientBuilder::new().build().unwrap();
+        assert_eq!(
+            octez_client.binary_path.to_str().unwrap(),
+            DEFAULT_BINARY_PATH
+        );
+        assert!(matches!(octez_client.base_dir, Directory::TempDir(_)));
+        assert!(!octez_client.disable_unsafe_disclaimer);
+        assert_eq!(octez_client.endpoint, Endpoint::localhost(8732))
+    }
+
+    #[test]
+    fn temp_dir_is_created_by_default() {
+        let octez_client = OctezClientBuilder::new().build().unwrap();
+        assert!(
+            matches!(octez_client.base_dir, Directory::TempDir(temp_dir) if temp_dir.path().exists())
+        );
+    }
+
+    #[test]
+    fn temp_dir_is_removed_on_drop() {
+        let octez_client = OctezClientBuilder::new().build().unwrap();
+        match &octez_client.base_dir {
+            Directory::TempDir(temp_dir) => {
+                let temp_dir_path = temp_dir.path().to_path_buf();
+                drop(octez_client);
+                assert!(!temp_dir_path.exists());
+            }
+            _ => panic!("Expected TempDir"),
+        };
+    }
+
+    #[test]
+    fn sets_custom_binary_and_dir_path() {
+        let temp_file = NamedTempFile::new().unwrap();
+        let binary_path = temp_file.path().to_path_buf();
+        let temp_dir = TempDir::new().unwrap();
+        let dir_path = temp_dir.path().to_path_buf();
+        let octez_client = OctezClientBuilder::new()
+            .set_binary_path(binary_path.clone())
+            .set_base_dir(dir_path.clone())
+            .build()
+            .unwrap();
+        assert_eq!(octez_client.binary_path, binary_path);
+        assert!(
+            matches!(octez_client.base_dir, Directory::Path(path) if path == dir_path)
+        );
+    }
+
+    #[test]
+    fn validates_base_dir_exists() {
+        let temp_dir = TempDir::new().unwrap();
+        let dir_path = temp_dir.path().to_path_buf();
+        let _ = std::fs::remove_file(&dir_path);
+        let _ = std::fs::remove_dir_all(&dir_path);
+        let octez_client = OctezClientBuilder::new()
+            .set_base_dir(dir_path.clone())
+            .build();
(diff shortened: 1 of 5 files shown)


crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn default() -> Self {
+        OctezClientDirectory::TempDir(tempdir().expect("Failed to create tempdir"))
+    }
+}
+
+// TODO: remove later
+#[allow(dead_code)]
+pub struct OctezClient {
+    // if None, use the binary in $PATH
+    binary_path: Option<PathBuf>,
```
huancheng-trili, line 86: I don't think this should be `Option`. The default path should have been set in `build` and passed to the client.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+    pub fn set_disable_unsafe_disclaimer(
+        mut self,
+        disable_unsafe_disclaimer: bool,
+    ) -> Self {
+        self.disable_unsafe_disclaimer = disable_unsafe_disclaimer;
+        self
+    }
+    pub fn build(mut self) -> OctezClient {
+        let base_dir = self
```
huancheng-trili, line 42: Just curious, why not use
```rs
self.base_dir.unwrap_or( <call tempdir and return path> )
```
but create another enum `OctezClientDirectory`?

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,152 @@
...
+            .base_dir
+            .take()
+            .map_or(OctezClientDirectory::default(), OctezClientDirectory::Path);
+        OctezClient {
```
huancheng-trili, line 46: Just my opinion, `config.build` generating a client instead of a built config seems less natural to me compared with `Client::new(config_builder.build())`. I'm fine if others are okay with this implementation.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,152 @@
...
+pub struct OctezClientConfig {
+    binary_path: Option<PathBuf>,
+    // if None, use temp directory
+    base_dir: Option<PathBuf>,
+    endpoint: OctezNodeEndpoint,
+    disable_unsafe_disclaimer: bool,
+}
+
+impl OctezClientConfig {
+    pub fn new(endpoint: OctezNodeEndpoint) -> Self {
```
huancheng-trili, line 14: Why must `new` take `endpoint`? If it's because of having to a default value for `endpoint`, the client binary by default sets the RPC endpoint to localhost:8732. We can probably follow this. What do you think?

crates/jstzd/src/task/octez_client.rs
ryutamago: yes i was looking into the design doc and thought about it. however i felt that it would be better to distinguish between a path to the binary of type `PathBuf` vs a command "octez-client" explicitly like this. wdyt @zcabter

crates/jstzd/src/task/octez_client.rs
ryutamago: i've followed the spec here but i agree with you. what about just renaming the `OctezClientConfig` to `OctezClientBuilder`? cc @zcabter

crates/jstzd/src/task/octez_client.rs
huancheng-trili: It's fine that the directory gets removed. The client binary will create one if the path does not point to an existing directory.

crates/jstzd/src/task/octez_client.rs
huancheng-trili: Do you mean asking users to call something like `OctezClientConfig::new(OctezNodeEndpoint::RpcEndpoint("localhost:8732"))` anyway?

crates/jstzd/src/task/octez_client.rs
ryutamago: oh but don't we still want the tmp file to be deleted after the octez client goes out of scope or is that handled by the binary as well?

crates/jstzd/src/task/octez_client.rs
ryutamago: nono, i meant implemenitng the default for OctezNodeEndpoint that will return `localhost:8732`. the client don't need to specify anything

huancheng-trili: Yes that sounds good

crates/jstzd/src/task/octez_client.rs
ryutamago:
```
./octez-client --base-dir ~/Desktop/oc gen keys leo            1 ↵ ──(Wed,Sep25)─┘
```
Error:
  Specified --base-dir does not exist. Please create the directory and try again.
  
i get error for gen keys still with non existent path

huancheng-trili: I see. I think it's okay to leave the directory there because it's created in `/tmp` and the operating system will clean it up eventually. It's fine if you want to do it here. I didn't do that with `OctezNode` Maybe I should implement that later on.

huancheng-trili: Oh I didn't know that it's different in `octez-client`. In octez-node it's fine with a non-existent `data-dir` parameter. I guess here we need to hold that directory.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+        mut self,
+        disable_unsafe_disclaimer: bool,
+    ) -> Self {
+        self.disable_unsafe_disclaimer = disable_unsafe_disclaimer;
+        self
+    }
+    pub fn build(mut self) -> OctezClient {
+        let base_dir = self
+            .base_dir
+            .map_or(OctezClientDirectory::default(), OctezClientDirectory::Path);
```
zcabter, line 41: `OctezClientDirectory::default()` is misleading because the default directory of octez-client is `~/.tezos-client`. Could you instead inline the temp dir creation?

zcabter, line 41: Second arg should be checked that it exists and that it is directory.

crates/jstzd/src/task/async_octez_node.rs
```diff
@@ -8,7 +8,7 @@ use tokio::process::{Child, Command};
 
 use anyhow::{anyhow, Result};
 
-fn path_or_default<'a>(path: Option<&'a PathBuf>, default: &'a str) -> &'a str {
+pub fn path_or_default<'a>(path: Option<&'a PathBuf>, default: &'a str) -> &'a str {
```
ryutamago, line 11: need this for the following pr

crates/jstzd/src/task/octez_client.rs
zcabter: Agree with Huan that `binary_path` shouldn't be optional. 

No need to distinguish between custom binary path and "octez-client" as the shell will do right thing.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+        OctezNodeEndpoint::RpcEndpoint(endpoint)
+    }
+}
+impl From<OctezNode> for OctezNodeEndpoint {
+    fn from(node: OctezNode) -> Self {
+        OctezNodeEndpoint::Node(node)
+    }
+}
+
+pub enum OctezClientDirectory {
```
zcabter, line 73:
```suggestion
pub enum Directory {
```

zcabter, line 73: Should implement `impl TryFrom<Path> for Directory` that validates Path is a directory or something similar along those lines.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+            base_dir,
+            endpoint: self.endpoint.unwrap_or_default(),
+            disable_unsafe_disclaimer: self.disable_unsafe_disclaimer,
+        }
+    }
+}
+
+pub enum OctezNodeEndpoint {
+    RpcEndpoint(Endpoint),
+    Node(OctezNode),
```
zcabter, line 53: I see what you are trying to do but it doesn't make sense to hold `OctezNode` as an endpoint type because `OctezNode` simply represents a running process - there are no apis defined on it

crates/jstzd/src/task/async_octez_node.rs
zcabter: I suggest you add it in the following PR

crates/jstzd/src/task/octez_client.rs
zcabter: Good idea, I'm in favour of `OctezClientBuilder` over `OctezClientConfig`

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn builds_default_octez_client() {
+        let octez_client = OctezClientBuilder::new().build();
+        assert!(octez_client.binary_path.is_none());
+        assert!(matches!(
+            octez_client.base_dir,
+            OctezClientDirectory::TempDir(_)
+        ));
+        assert!(!octez_client.disable_unsafe_disclaimer);
+        match octez_client.endpoint {
+            OctezNodeEndpoint::RpcEndpoint(endpoint) => {
```
zcabter, line 107: I suggest to implement derive `PartialEq` on OctezNodeEndpoint, then you could do
`assert_eq!(octez_client.endpoint, OctezNodeEndpoint::RpcEndpoint(Endpoint::localhost(8732)))`

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    #[test]
+    fn temp_dir_is_created_by_default() {
+        let octez_client = OctezClientBuilder::new().build();
+
+        let temp_dir = match octez_client.base_dir {
+            OctezClientDirectory::TempDir(temp_dir) => temp_dir,
+            _ => panic!("Expected TempDir"),
+        };
+        assert!(temp_dir.path().exists());
```
zcabter, line 122:
```suggestion
        match octez_client.base_dir {
            OctezClientDirectory::TempDir(temp_dir) => assert!(temp_dir.path().exists()),
            _ => panic!("Expected TempDir"),
        };
        
```

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    #[test]
+    fn temp_dir_is_removed_on_drop() {
+        let octez_client = OctezClientBuilder::new().build();
+
+        let temp_dir_path = match &octez_client.base_dir {
+            OctezClientDirectory::TempDir(temp_dir) => temp_dir.path().to_path_buf(),
```
zcabter, line 130: Inline the assert like above for consistency

zcabter, line 130: Inline the drop and assert (like above) for consistency

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn sets_binary_custom_path() {
+        let binary_path = PathBuf::from("/path/to/binary");
+        let dir_path = PathBuf::from("/path/to/directory");
+        let octez_client = OctezClientBuilder::new()
+            .set_path(binary_path)
+            .set_base_dir(dir_path)
+            .build();
+        assert_eq!(
+            octez_client.binary_path.unwrap(),
+            PathBuf::from("/path/to/binary")
```
zcabter, line 147:
```suggestion
           binary_path.clone()
```

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,156 @@
...
+        );
+        match octez_client.base_dir {
+            OctezClientDirectory::Path(path) => {
+                assert_eq!(path, PathBuf::from("/path/to/directory"))
```
zcabter, line 151:
```suggestion
                assert_eq!(path, dir_path.clone())
```

crates/jstzd/src/task/octez_client.rs
ryutamago: i've followed the design doc here as well, i will just use `Endpoint` in that case

crates/jstzd/src/task/octez_client.rs
ryutamago: i already added the validation logic as part of `OctezClientBuilder::build` where we validates the binary path and the directory. I think renaming it to `Directory` and adding validation for it would be too generic to be in this file and out of scope of this PR.
We can leave the implementation in the build for now and create a more generic "File" and "Directory" struct that validates the path in a later PR. That way we can make OctezNode and other Nodes use these structs too. wdyt?

ryutamago: done

crates/jstzd/src/task/octez_client.rs
ryutamago: done

crates/jstzd/src/task/octez_client.rs
ryutamago: done

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+                drop(octez_client);
+                assert!(!temp_dir_path.exists());
+            }
+            _ => panic!("Expected TempDir"),
+        };
+    }
+
+    #[test]
+    fn sets_custom_binary_and_dir_path() {
+        let binary_path = PathBuf::from("/tmp/binary1");
```
zcabter, line 136: I think you should use `NamedTempFile` then get its path with `NamedTempFile::path`.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+        let dir_path = PathBuf::from("/tmp/dir1");
+        let _ = std::fs::remove_dir(&binary_path);
+        let _ = std::fs::remove_file(&binary_path);
+        let _ = std::fs::remove_dir(&dir_path);
+        let _ = std::fs::remove_file(&dir_path);
+        std::fs::File::create(&binary_path).unwrap();
+        std::fs::create_dir(&dir_path).unwrap();
```
zcabter, line 143: To avoid managing the temp files, use `NamedTempFile` then get its path with `NamedTempFile::path`. The same can be done `TempDir::new` and `TempDir::path`

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+        let octez_client = OctezClientBuilder::new()
+            .set_path(binary_path.clone())
+            .set_base_dir(dir_path.clone())
+            .build()
+            .unwrap();
+        assert_eq!(octez_client.binary_path, binary_path);
+        match octez_client.base_dir {
+            Directory::Path(path) => {
+                assert_eq!(path, dir_path)
+            }
```
zcabter, line 153: I think its possible to match on the pattern and add a value guard:
```
assert!(matches!(octez_clientl.base_dir, Directory::Path(path) if path == dir_path)) 
```

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+            _ => panic!("Expected Path"),
+        }
+    }
+
+    #[test]
+    fn validates_base_dir() {
+        let dir_path = PathBuf::from("/tmp/dir2");
+        let _ = std::fs::remove_dir_all(&dir_path);
+        let _ = std::fs::remove_file(&dir_path);
```
zcabter, line 162: Use temp dir like above.

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+        let octez_client = OctezClientBuilder::new()
+            .set_base_dir(dir_path.clone())
+            .build();
+        assert!(octez_client.is_err());
```
zcabter, line 166: Redundant line. Assertion checked by line below

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+        assert!(octez_client
+            .is_err_and(|e| &e.to_string() == "Base directory does not exist"));
+        std::fs::File::create(&dir_path).unwrap();
+        let octez_client = OctezClientBuilder::new()
+            .set_base_dir(dir_path.clone())
+            .build();
+        assert!(octez_client.is_err());
```
zcabter, line 173: Redundant line. Assertion checked by line below

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    #[test]
+    fn validates_binary_path() {
+        let binary_path = PathBuf::from("/tmp/binary3");
+        let _ = std::fs::remove_dir_all(&binary_path);
+        let _ = std::fs::remove_file(&binary_path);
+        let octez_client = OctezClientBuilder::new()
+            .set_path(binary_path.clone())
+            .build();
+        assert!(octez_client.is_err());
```
zcabter, line 186: Redundant assertion

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+    // if None, use localhost:8732 (the default endpoint for octez-node)
+    endpoint: Option<Endpoint>,
+    disable_unsafe_disclaimer: bool,
+}
+
+impl OctezClientBuilder {
+    pub fn new() -> Self {
+        OctezClientBuilder::default()
+    }
+    pub fn set_path(mut self, path: PathBuf) -> Self {
```
zcabter, line 23: Nit: add empty line in between `fn` definitions.
Note: This should be part of the formatter. Tracked: https://linear.app/tezos/issue/JSTZ-154/update-formatter-to-add-empty-line-in-between-fn-definitions-in-impl

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+            if !base_dir.is_dir() {
+                bail!("Base directory is not a directory");
+            }
+        }
+        Ok(())
+    }
+}
+
+// TODO: remove later
+#[allow(dead_code)]
```
zcabter, line 82: This oculd be removed

crates/jstzd/src/task/octez_client.rs
```diff
@@ -0,0 +1,199 @@
...
+        let _ = std::fs::remove_dir_all(&binary_path);
+        let _ = std::fs::remove_file(&binary_path);
+        let octez_client = OctezClientBuilder::new()
+            .set_path(binary_path.clone())
+            .build();
+        assert!(octez_client.is_err());
+        assert!(
+            octez_client.is_err_and(|e| &e.to_string() == "Binary path does not exist")
+        );
+        std::fs::create_dir(&binary_path).unwrap();
```
zcabter, line 190: Use temp dir

crates/jstzd/src/task/octez_client.rs
zcabter: I would imagine that Directory could be re-used by other components so having default validation implemented for it would be ideal

crates/jstzd/src/task/octez_client.rs
ryutamago: much cleaner 🧹

crates/jstzd/src/task/octez_client.rs
ryutamago: neat!

crates/jstzd/src/task/octez_client.rs
ryutamago: done

crates/jstzd/src/task/octez_client.rs
ryutamago: my ide complains about it so thought we needed it

crates/jstzd/src/task/octez_client.rs
ryutamago: i think we could work on that once it's needed. probably better to define it outside of otctez_client in that case

zcabter: Hmm I'm not satisfied that you are able to create a `Directory` that isn't a directory. It leaks the semantic.

Secondly, the author who ultimately makes the change may not know about your code changes and miss refactoring your code.  And I agree that this should be outside `octez_client.rs`, perhaps in `mod.rs`, sorry I missed that.

zcabter (approved):

PR #574, opened 2024-09-26
feat(jstzd): implement ProtocolParameterBuilder

huancheng-trili:
# Context

Part of JSTZ-149.

# Description

Implement struct `ProtocolParameterBuilder` that builds full protocol parameter sets. The implementation in this PR adds bootstrap accounts to the base protocol parameter set.

Base parameters will be stored as json files in directory `resources` and used as templates to build the final parameter set for the sandbox. The parameters are to be determined and I just put placeholders there to show that the tests work. 

# Manually testing the PR

Added unit test cases.

```sh
$ cargo test --lib protocol
running 3 tests
test protocol::tests::parameter_builder ... ok
test protocol::tests::parameter_builder_default ... ok
test protocol::tests::build_parameters_from_given_file ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 21 filtered out; finished in 0.00s
```


crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                .parameter_file(self.constants)
+                .to_str()
+                .unwrap(),
+        )
+        .unwrap();
+        let mut raw_json: Value = serde_json::from_slice(&f.data)?;
+        let json = raw_json
+            .as_object_mut()
+            .ok_or(anyhow::anyhow!("Failed to convert json file"))?;
+        drop(f);
```
zcabter, line 105: Why can't we just let rust drop it at the end of the function?

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.constants = constants;
+        self
+    }
+
+    pub fn set_path(&mut self, path: &str) -> &mut Self {
+        self.path = Some(PathBuf::from(path));
+        self
+    }
+
+    pub async fn build(self) -> anyhow::Result<PathBuf> {
```
zcabter, line 93: Does this function need to be `async`?

zcabter:
I ran the test command but got a different output. I don't think this PR adds boostrap accounts? 
```
running 3 tests
test protocol::tests::parameter_builder ... ok
test protocol::tests::parameter_builder_default ... ok
test protocol::tests::build_protocol_parameter ... ok
```

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        let f = ProtocolParameterFile::get(
+            self.protocol
+                .parameter_file(self.constants)
+                .to_str()
+                .unwrap(),
+        )
+        .unwrap();
```
zcabter, line 100: We probably want `?` here

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        let mut raw_json: Value = serde_json::from_slice(&f.data)?;
+        let json = raw_json
+            .as_object_mut()
+            .ok_or(anyhow::anyhow!("Failed to convert json file"))?;
+        drop(f);
+        let path = self
+            .path
+            .unwrap_or(tempfile::NamedTempFile::new().unwrap().path().to_path_buf());
```
zcabter, line 108: I think we should always write the new parameters file to a temp file

crates/jstzd/src/protocol/mod.rs
zcabter, line 93: I think maybe there's a misunderstanding what this function does. If the caller supplies a parameters file then it should use that one, otherwise we will default to our sandbox params file. Then, any values set on the builder should take precedence over the values set in the parameters file.

crates/jstzd/src/protocol/mod.rs
huancheng-trili: We can. I think I did something previously that required dropping the file early. I've removed this line.

crates/jstzd/src/protocol/mod.rs
huancheng-trili: No. Updated.

crates/jstzd/src/protocol/mod.rs
huancheng-trili: I've replaced unnecessary `unwrap`s with `?`.

crates/jstzd/src/protocol/mod.rs
zcabter: Agreed, the `set_protocol` and `set_constants `would mainly be used for our internal framework since we won't be passing in specific params files but will be using the base params files provided. Perhaps it should be an error if set_protocol differs from the supplied parameters file?

huancheng-trili: But there's no way of knowing which protocol version the provided parameter file is for. octez node does reject parameters that are invalid for a given protocol version, but if all of them are valid, there is no way of determining if the intended protocol version is different from the one updated by `set_protocol`.

huancheng-trili:
Yup. Initially bootstrap accounts were included in this PR. I moved the changes into #578 but forgot to update the test result here.

crates/jstzd/src/protocol/mod.rs
zcabter: Ah yes, my bad, thought the protocol was in the parameters file. 

No-op is fine.

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    pub fn build(&mut self) -> anyhow::Result<PathBuf> {
+        let mut raw_json = self.load_parameter_json()?;
+        let json = raw_json.as_object_mut().ok_or(anyhow::anyhow!(
+            "Failed to convert loaded json file into a json object"
+        ))?;
+
+        let output_path = tempfile::NamedTempFile::new().unwrap().path().to_path_buf();
+        let file = std::fs::File::create(&output_path)?;
```
zcabter, line 101: Hmm `tempfile::NamedTempFile::new()` should be enough to create the file but I guess we don't want to drop the temp file when the function exits. Instead, I suggest adding the following trait marker 
```
pub trait ReadWritable: Read + Write {}

impl ReadWritable for tempfile::NamedTempFile {}
```

then this function will return a Result<impl ReadWritable> and this part can become: 
```
let file = tempfile::NamedTempFile::new().unwrap();
serde_json::to_writer(file.as_file(), &json)?;
Ok(file)
```

crates/jstzd/src/protocol/mod.rs
huancheng-trili: This is updated as well

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+        let source_path = format!(
+            "{}/resources/test/protocol_parameters/sandbox/test-hash",
+            env!("CARGO_MANIFEST_DIR")
+        );
+        builder.set_source_path(&source_path);
+        let output_path = builder.build().unwrap();
+        let file = std::fs::File::open(output_path).unwrap();
+        let json: serde_json::Value = serde_json::from_reader(file).unwrap();
+
+        // The given parameter file `test-hash` has only one kv-pair "foo:bar"
```
zcabter, line 183: Nit: I would have preferred to write the contents to a temp file since the contents of the file is tiny and we can easily see what to expect instead of commenting here.

huancheng-trili: It is written to a temp file. `json` is loaded from the path to the temp file. I'll update the comment to avoid confusion.

crates/jstzd/src/protocol/mod.rs
huancheng-trili: Does it matter if we drop the temp file? Temp files are different from `TempDir`: they don't get removed when the instance is dropped.

crates/jstzd/src/protocol/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+use serde_json::Value;
+use std::fmt::Display;
+use std::io::Read;
+use std::path::{Path, PathBuf};
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum ProtocolConstants {
+    Sandbox,
+    #[cfg(test)]
+    TestConstants,
```
zcabter, line 11: Looking at the tests, if we just set protocol to alpha and constant to sandbox, it will automatically load base parameters from the `resource/protocol_parameters/sandbox/ProtoAlpha...` which is a sufficient test. `TestConstants` and `TestVersion` could be removed.

crates/jstzd/src/protocol/mod.rs
zcabter: `NamedTempFile` owns `TempPath` and if it drops then `TempPath` gets drop which will delete the file.

 I think what's happening here is that when you call `to_path_buf()`, the path is copied and because there is no reference to `NamedTempFile`, it gets deleted. The file then gets created again on the next line.

huancheng-trili: I see. Yes I realised I was wrong about that when I tried to use a temp source file in the last test case. I've updated this.

crates/jstzd/src/protocol/mod.rs
huancheng-trili: I added `TestConstants` and `TestVersion` to check that the values were actually updated with the setters. The default values for these fields in the builder are `Sandbox` and `Alpha`, so if we set these values but the underlying logic somehow is wrong, we won't be able to detect it by looking for `Sandbox` and `Alpha`. We wouldn't need this if there are more than one possible enum values, but we don't at this moment. If you still think this is not really necessary, I can remove it.

zcabter: Oh I see, `new` calls `default`. Hmm should `new` create an empty Builder with optional fields? That's more conventional as a builder.

Either way, I don't think the test enums are necessary.

huancheng-trili: I was thinking of setting the default values easily so didn't use optional fields. I've changed the logic here so that we can still kind of test default values with optional fields and without the test enums.

PR #575, opened 2024-09-26
chore(contributing): update PR guidelines

zcabter:
# Context
As the jstz team grows, it is expected that PR management becomes inconsistent between deveopers. This PR adds PR guidelines to `CONTRIBUTING.md` that all jstz devs should adhere to where possible.

Closes [JSTZ-153](https://linear.app/tezos/issue/JSTZ-153/update-pr-guidelines-in-contributingmd)

# Description
Updates the Creating a pull request section of `CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -186,18 +186,50 @@ To edit documentation:
 
 ## Creating a pull request 📩
 
-Please write a meaningful description for your pull request. If your pull request references an issue or Asana task, please mention it in the description. The format for pull request titles is `component/kind: description`.
+Please adhere to the following PR guidelines
 
-For 'stacked' pull requests, please ensure your pull request lists all dependencies and uses the predecessor branch as the target.
+### PR preparation
 
-To make sure your pull request is easy to review:
+- Have a title of the form `type(component): subject`. See [commitlint](https://github.com/conventional-changelog/commitlint/#what-is-commitlint) for possible types
+- Give meaning context and description about the issue solved/feature added
+  - Context - Why is this change required? What problem does it solve?
+  - Description - Describe your changes in detail. Add anything else to highlight to the reviewer.
+- Anticipate questions: explain anything which may look surprising either in PR description or in code comments if it will be useful for future readers
+- Include [magic word + link/issue ID](https://linear.app/docs/github?tabs=206cad22125a#link-using-pull-requests) in the description for project tracking
+- 1 Linear issue per PR
+- Clean commit history (jstz uses a semi-linear commit history)
+- [Meaningful commit messages](https://cbea.ms/git-commit/)
+- Have tests / testing instructions
+- Aim for above 80% patch coverage
+- Favour small PRs
+- Author's are responsible to have assignees / reviewers allocated
+- Stacked PRs should target their predecessor
 
-- **Use `git rebase`**. We maintain a semi-linear git history. This means that your branch should be a direct suffix of `main` (or the target branch). Addtionally, it should not contain any merge commits.
-- **Don't push fixup commits\***. When your reviewer asks for changes, they will want you to rewrite your branch history so that the commit history is clean.
+### PR management
 
-  If you branch history is dirty (containing fixup commits, etc) then we will squash-merge\*. However, this is undesirable as we lose the information that individual commits provide.
+- Use assignee field to indicate action is required from that particular person
+- 1 assignee at a time to avoid responsibility dilution
+- Commenter responsible for resolving conversations except in code suggestions which automatically close a conversation when applied
+- Assignee assigns the next person to take action Eg. Reviewer assigns author after providing feedback. Author assigns Reviewer after applying feedback.
+- Mark as Draft to indicate PR isn’t ready to be reviewed/merged
+- Maintain a logical commit history if possible. Squash commit history if not (not ideal as lose information that the commit provides).
+- Force push to relevant commits when applying simple changes
+- Use Fixup commits when applying complicated changes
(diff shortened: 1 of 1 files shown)


huancheng-trili:
I'll just take this opportunity and ask some questions here about the guidelines. Why not just squash merge by default?

Since we expect PRs to be small, each merged PR should be small enough to fit in one single commit when they are merged, then why not just squash merge instead of asking authors to squash commits manually? Then there will be no need to keep rebasing or force-pushing either as merge commits or tiny fix commits in branches won't matter as long as PRs can only be merged when they have merged the latest commit in the target branch. I personally find manually squashing pretty meaningless when this can be done automatically on merge. Having multiple "clean" commits in one PR will not be the case when PRs are expected to be small. Even if a PR is inevitably large, having multiple clean commits won't help either since eventually reviewers only care about whether or not follow-up fix commits address the issues/comments and it will still be the author's duty to manually figure out how to split changes into clean commits, which is not worth the time in my opinion.

zcabter:
> Since we expect PRs to be small, each merged PR should be small enough to fit in one single commit when they are merged

The guideline is to prefer small PRs but there will be some features that even their smallest component is chunky and breaking it into sub-prs would cause `main` to be in a partially incomplete state. In this case, it is much easier to understand and navigate the work via its commits if the author ensured a logical and singly- responsible commit history.  Delineating the exact logical boundary of the commits requires judgement on the devs part. In cases where 

> having multiple clean commits won't help either since eventually reviewers only care about whether or not follow-up fix commits address the issues/comments

One tends to lose meaningfulness of commit messages if the scope of the commit is too large which makes `main` commit history less useful, which makes understanding recent changes to `main` more difficult. There's also a case where the work needs to be taken over by someone else (e.g author went on holiday) and in that case, the existing commits give attribution to the dev who did that work.

Re: Reviewers don't care - I personally take a look at commit history first to get an idea of what work was done esp. for large PR or areas of codebase that I'm not familiar with.

> which is not worth the time in my opinion.

Keeping a clean history can be time consuming but I've found that using tools like [stg](https://stacked-git.github.io/) makes it a lot easier. 

> And rebasing simply screws up stacked PRs.

I've been using a custom script to help manage these for when I have many concurrent stacked PRs. But I've found that just manually rebasing when the stack is short is quite manageable. 

I'm of the opinion squashing into single commit is a pragmatic trade-off for higher velocity considering the size and experience of the team.

huancheng-trili:
> The guideline is to prefer small PRs but there will be some features that even their smallest component is chunky and breaking it into sub-prs would cause main to be in a partially incomplete state.

I think this depends on what "a partially incomplete state" means. To me, as long as every PR to be merged into main builds and passes all checks, the main branch will never be in an incomplete state. Large PRs can then be split into multiple smaller stacked ones as long as the author makes all of them pass the checks. This would be the same as reviewing a series of clean commits in one large PR, but reviewing can be easier in that case because each PR can give more details -- notes can be added to PR descriptions, comments can be added separately rather than all flying around in one single PR. It's also easier for authors to re-scope each individual commit by moving changes between branchs than having to undo commits, especially when there are more commits on top of the one to be updated. The outcome will be the same in the main branch. It's just that clean commits are achieved by several smaller squashed merged PRs.

> One tends to lose meaningfulness of commit messages if the scope of the commit is too large which makes main commit history less useful, which makes understanding recent changes to main more difficult.

PRs are always there to provide more information. After finding out when something was merged into the main branch, I always check the corresponding PR to read the PR description, the original ticket, discussions, comments, etc. Commit messages never provide that many details.

> There's also a case where the work needs to be taken over by someone else (e.g author went on holiday) and in that case, the existing commits give attribution to the dev who did that work.

I would say that can be settled by those involved. Creating smaller PRs mitigates this too.

zcabter:
> I think this depends on what "a partially incomplete state" means. To me, as long as every PR to be merged into main builds and passes all checks, the main branch will never be in an incomplete state. Large PRs can then be split into multiple smaller stacked ones as long as the author makes all of them pass the checks. 

Yes if the author can get all the checks passing then that's usually fine. I think it can be a bit cumbersome to write tests for scaffolding code that was put in just to get the code to compile/ maintain the same behaviour while the component is in transition but if the dev is willing to write it I think that's ok.

Ultimately, the goal of commit history is to help reviewers understand the code. Whether that is delivered in multiple PRs or not is fine. However, I don't think single commit PRs should be the priority when scoping work. Instead, we should prioritize a logically related change set (which usually makes it easier to review) and then a small scope.  If they conflict, logical change takes precedence.

I think that if its possible to scope work into a single commit PR, then that's ideal, but if not then there's that flexibility to have multiple "setup" commits before the main changes. In some cases, for example cross component sensitive changes, it is often helpful for reviewers to see how exactly these changes are going to affect the final solution to evaluate its suitability. If the change occurs across multiple PRs, this can be very difficult to review - The solution to this with single commit PRs is to have 1 big PR at the end that shows the final changes. I think that's fine as it solves the problem of providing context but I think it will fail over at some point. The other solution would be to scope the work across slightly larger multi commit PRs; combining sensitive changes with the subset of downstream changes.  I think the guidelines are flexible enough to allow the dev to apply judgement on what's best for PR scope.

> PRs are always there to provide more information. After finding out when something was merged into the main branch, I always check the corresponding PR to read the PR description, the original ticket, discussions, comments, etc. Commit messages never provide that many details.

I think they serve different purposes; PRs/issues provides a rich context while a commit provide simple locallized context. I find that working with commits to understand changes is a lot quicker if its enough.

huancheng-trili (approved):

johnyob (approved):

PR #576, opened 2024-09-27
feat(jstzd): add bootstrap contracts to ProtocolParameterBuilder

huancheng-trili:
# Context

Part of JSTZ-149.

# Description

Adds bootstrap contracts to the base protocol parameter set.

A `BootstrapContract` instance has 3 fields:
```rs
pub struct BootstrapContract {
    script: serde_json::Value,
    amount_tez: u64,
    hash: Option<tezos_crypto_rs::hash::ContractKt1Hash>,
}
```

`script` is the contract code in micheline JSON format. `hash` is the hash of this contract to be deployed. Note that it's optional because the node will just generate one if it's not given, but it can actually be set to any valid contract hash, which might be handy when we need to know the address of a contract in advance.

# Manually testing the PR

* Generated one parameter file with the sample contract code attached and tested the file in a sandbox.
* Unit test: add new test cases
* Integration test: waiting for other octez client implementation


crates/octez/src/async/protocol.rs
```diff
@@ -183,6 +199,25 @@ impl ProtocolParameterBuilder {
...
+        if let Some(value) = json.get("bootstrap_contracts") {
+            if let Some(existing_contracts) = value.as_array() {
+                for contract in existing_contracts {
+                    contracts.add_contract(BootstrapContract::try_from(contract)?);
+                }
+            }
+        }
+        for contract in self.bootstrap_contracts.contracts() {
+            contracts.add_contract(contract.clone());
+        }
```
zcabter, line 217: `impl BootstrapContracts` should define `fn merge(&mut self, bootstrap_contracts: BootstrapContracts)`

crates/octez/src/async/protocol.rs
```diff
@@ -183,6 +199,25 @@ impl ProtocolParameterBuilder {
...
+        json: &mut serde_json::Map<String, Value>,
+    ) -> anyhow::Result<()> {
+        let mut contracts = BootstrapContracts::default();
+        if let Some(value) = json.get("bootstrap_contracts") {
+            if let Some(existing_contracts) = value.as_array() {
+                for contract in existing_contracts {
+                    contracts.add_contract(BootstrapContract::try_from(contract)?);
+                }
+            }
+        }
```
zcabter, line 214: If Deserializer is implemented for `BootstrapContracts`, then you could do `let bootstrap_contracts: BoostramContract = serde_json::from_value(value)` which is much cleaner.

zcabter (changes requested):

crates/octez/src/async/bootstrap.rs
```diff
@@ -91,16 +92,127 @@ impl From<&BootstrapAccounts> for Value {
...
+                "amount".to_owned(),
+                Value::String(value.amount_mutez.to_string()),
+            );
+            if let Some(v) = &value.hash {
+                map.insert("hash".to_owned(), Value::String(v.to_string()));
+            }
+            map
+        })
+    }
+}
```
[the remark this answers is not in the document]
huancheng-trili: Updated

crates/octez/src/async/bootstrap.rs
```diff
@@ -91,16 +92,127 @@ impl From<&BootstrapAccounts> for Value {
...
+
+impl BootstrapContract {
+    pub fn new(
+        script: Value,
```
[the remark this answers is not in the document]
huancheng-trili: I tried but didn't find it that helpful. My editor did not pick up the alias or the doc attached to the alias. I ended up adding a comment to the field, which seems to be more helpful.

crates/octez/src/async/bootstrap.rs
```diff
@@ -91,16 +92,127 @@ impl From<&BootstrapAccounts> for Value {
...
+pub struct BootstrapContracts {
+    keys: HashSet<String>,
+    contracts: Vec<BootstrapContract>,
+}
+
+impl From<&BootstrapContracts> for Value {
+    fn from(value: &BootstrapContracts) -> Self {
+        Value::Array(value.contracts.iter().map(Value::from).collect())
+    }
+}
```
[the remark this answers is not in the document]
huancheng-trili: Updated

crates/octez/src/async/protocol.rs
```diff
@@ -1,5 +1,6 @@
-pub use super::bootstrap::BootstrapAccount;
-use super::bootstrap::BootstrapAccounts;
+pub use super::bootstrap::{BootstrapAccount, BootstrapContract};
```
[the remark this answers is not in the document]
huancheng-trili: I don't think so. Users only interact with `ProtocolParamsBuilder` and its method `add_contract` that involves `BootstrapContract`. The container types are mostly for managing internal logic like conflicts in contract addresses.

crates/octez/src/async/protocol.rs
huancheng-trili: Updated

crates/octez/src/async/protocol.rs
huancheng-trili: Updated

crates/octez/src/async/bootstrap.rs
```diff
@@ -91,16 +95,127 @@ impl From<&BootstrapAccounts> for Value {
     }
 }
 
+#[serde_with::serde_as]
+#[serde_with::skip_serializing_none]
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
+pub struct BootstrapContract {
+    /// Smart contract code in micheline format as a `serde_json::Value` instance.
```
zcabter, line 102: Thanks, perhaps it would be helpful to be explicit that the code does not get validated.

crates/octez/src/async/bootstrap.rs
```diff
@@ -91,16 +95,127 @@ impl From<&BootstrapAccounts> for Value {
...
+    script: Value,
+    #[serde(rename = "amount")]
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    amount_mutez: u64,
+    hash: Option<ContractKt1Hash>,
```
zcabter, line 107: could you add a doc string comment to `hash` like how its explained in the Pr desc?

crates/octez/src/async/protocol.rs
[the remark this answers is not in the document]
zcabter: Yes, that's what I meant. We could indeed remove the `pub` here and merge the imports into line 2.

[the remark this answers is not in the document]
huancheng-trili: Sorry I misunderstood this a bit. I'm re-exporting `BootstrapAccount` and `BootstrapContract` here because I think it makes sense to make them available through this module. These types are both used to construct protocol parameters. Module `bootstrap` is only visible to the crate and thus `BootstrapAccount` and `BootstrapContract` won't be available through that module. I think it's okay to keep it private.

crates/octez/src/async/bootstrap.rs
huancheng-trili: Updated

crates/octez/src/async/bootstrap.rs
huancheng-trili: Updated

crates/octez/src/async/protocol.rs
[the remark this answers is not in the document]
zcabter: Ah ok

zcabter (approved):
LGTM

PR #579, opened 2024-09-27
feat(jstzd): add bootstrap smart rollups to ProtocolParameterBuilder

huancheng-trili:
# Context

Completes JSTZ-149.

# Description

Adds bootstrap smart rollups to the base protocol parameter set.

Each `BootstrapSmartRollup` instance has 4 fields:
```rs
pub struct BootstrapSmartRollup {
    /// Rollup address.
    address: SmartRollupHash,
    /// Type of Proof-generating Virtual Machine (PVM) that interprets the kernel.
    pvm_kind: SmartRollupPvmKind,
    /// Smart rollup kernel in hex string.
    kernel: String,
    /// Michelson type of the rollup entrypoint in JSON format.
    parameters_ty: Value,
}
```

To add a bootstrap rollup, the simplest way imo is to create a sandbox and actually deploy the rollup. The operation where the rollup is originated contains all the information needed to instantiate `BootstrapSmartRollup`. Added one example file that contains all necessary information of a toy rollup. Note that `address` can be any valid smart rollup address.

# Manually testing the PR

* Generated one parameter file with the sample smart rollup code attached and tested the file in a sandbox.
* Unit test: add new test cases
* Integration test: waiting for other octez-client implementation

### DIFF
--- a/crates/octez/src/async/bootstrap.rs
+++ b/crates/octez/src/async/bootstrap.rs
@@ -5,7 +5,7 @@ use serde::de::{Error, Visitor};
 use serde::ser::SerializeSeq;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use tezos_crypto_rs::hash::ContractKt1Hash;
+use tezos_crypto_rs::hash::{ContractKt1Hash, SmartRollupHash};
 
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct BootstrapAccount {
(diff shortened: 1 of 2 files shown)


crates/octez/src/async/bootstrap.rs
```diff
@@ -209,18 +209,140 @@ impl BootstrapContracts {
     }
 }
 
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
+pub enum SmartRollupPvmKind {
+    #[serde(rename = "wasm_2_0_0")]
+    Wasm,
+    #[serde(rename = "arith")]
+    Arith,
```
zcabter, line 217: This is not a real PVM. It was a reference implementation when the pvm code first came into being. Can be removed

crates/octez/src/async/protocol.rs
```diff
@@ -216,6 +233,23 @@ impl ProtocolParameterBuilder {
...
+        let mut rollups = BootstrapSmartRollups::default();
+        if let Some(value) = json.get("bootstrap_smart_rollups") {
+            let existing_rollups = serde_json::from_value(value.clone())?;
+            rollups.merge(&existing_rollups);
+        }
+        rollups.merge(&self.bootstrap_smart_rollups);
+        json.insert(
+            "bootstrap_smart_rollups".to_owned(),
+            serde_json::to_value(rollups)?,
+        );
```
zcabter, line 250: Might be nice to generalize the merge of bootstrap containers in another PR. I think they just need to implement a `trait Mergeable<T> { fn merge(&self) -> Result<T> }` and this fn could be generalized to:
```
fn merge<T: Mergeable + Serialize + Default>(value: T, json: &mut serde_json::Map<String, Value>, key: &str) -> Result<()> 
```

zcabter (approved):

PR #587, opened 2024-10-02
jstzd: client import secret key

ryutamago:
# Context

[import secret key](https://linear.app/tezos/issue/JSTZ-134/implement-octezclientimport-secret-key)

# Description

add import secret key for octez client. 
I've replaced the `spawn_command` with `spawn_and_wait_command` as octez client won't have any long running processs 

# Manually testing the PR

added two integration tests that tests for importing the key
```
cargo test --package jstzd --test octez_client_test 
```


crates/jstzd/src/task/octez_client.rs
```diff
@@ -123,38 +123,49 @@ impl OctezClient {
         Ok(command)
     }
 
-    fn spawn_command<S: AsRef<OsStr>, I: IntoIterator<Item = S>>(
+    async fn spawn_and_wait_command<S: AsRef<OsStr>, I: IntoIterator<Item = S>>(
         &self,
         args: I,
-    ) -> Result<Child> {
+    ) -> Result<()> {
```
zcabter, line 129: I think this should return `Result<Output>`

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -69,5 +72,40 @@ async fn generates_keys_throws() {
     let alias = "test_alias".to_string();
     let _ = octez_client.gen_keys(&alias, None).await;
     let res = octez_client.gen_keys(&alias, None).await;
-    assert!(res.is_err_and(|e| e.to_string().contains("failed to generate key")));
+    assert!(res.is_err_and(|e| { e.to_string().contains("\"gen\" \"keys\"") }));
+}
+
```
zcabter, line 77: Maybe add 1 more test where the gen keys is called with the signature field set. Probably should have been covered in the gen keys PR

crates/jstzd/src/task/octez_client.rs
ryutamago: i don't think we need it now unless we have a method that needs the stdout. The stdout/stderr gets printed out in the cli and i think that's sufficient. do you have anything in mind that would require the stdout from the client?

zcabter: Yes, stdout will be parsed to get output for commands later. Making the change now means that your commands in this PR will not need to be refactored later.

ryutamago: ok i i will add this for now and test it in the following PR for show address

crates/jstzd/tests/octez_client_test.rs
ryutamago: done

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -58,6 +61,27 @@ async fn generates_keys() {
...
+    let res = octez_client
+        .gen_keys(&alias, Some(jstzd::task::octez_client::Signature::BLS))
+        .await;
+    assert!(res.is_ok());
+    let hashes = first_item(read_file(&base_dir.join("public_key_hashs")));
+    let pub_keys = first_item(read_file(&base_dir.join("public_keys")));
+    let secret_keys = first_item(read_file(&base_dir.join("secret_keys")));
+    assert_eq!(hashes["name"], alias);
+    assert_eq!(pub_keys["name"], alias);
+    assert_eq!(secret_keys["name"], alias);
```
zcabter, line 82: Maybe you could check that the prefixes are correct.

zcabter:
Left some nit commit, lgtm otherwise.

crates/jstzd/tests/octez_client_test.rs
ryutamago: done

PR #588, opened 2024-10-02
feat(jstzd): implement node run option

huancheng-trili:
# Context

Completes JSTZ-116.

# Description

Replaces run options in string with a proper struct in order to be more selective compared with accepting random strings.

Also moved `OctezNodeConfig` to crate octez since it's actually not just for jstzd.

# Manually testing the PR

Added test cases for unit testing in crate octez.

### DIFF
--- a/crates/octez/src/node_config.rs
+++ b/crates/octez/src/node_config.rs
@@ -1,0 +1,292 @@
+use crate::unused_port;
+use anyhow::Result;
+use std::path::PathBuf;
+
+const DEFAULT_NETWORK: &str = "sandbox";
+const DEFAULT_BINARY_PATH: &str = "octez-node";
+const LOCALHOST: &str = "localhost";
+const LOCAL_ADDRESS: &str = "127.0.0.1";
+
+#[derive(Clone, PartialEq, Debug)]
+pub enum OctezNodeHistoryMode {
+    Archive,
+    // The numerical value represents additional cycles preserved. 0 is acceptable.
+    Full(u8),
+    Rolling(u8),
+}
+
+impl ToString for OctezNodeHistoryMode {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Archive => "archive".to_owned(),
+            Self::Full(v) => format!("full:{v}"),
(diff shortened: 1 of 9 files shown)


crates/octez/src/node_config.rs
```diff
@@ -0,0 +1,278 @@
...
+#[derive(Default, Clone)]
+pub struct OctezNodeConfig {
+    /// Path to the octez node binary.
+    pub binary_path: PathBuf,
+    /// Path to the directory where the node keeps data.
+    pub data_dir: PathBuf,
+    /// Name of the tezos network that the node instance runs on.
+    pub network: String,
+    /// HTTP endpoint of the node RPC interface, e.g. 'localhost:8732'
+    pub rpc_endpoint: String,
```
ryutamago, line 98: both the client and baker uses the `Endpoint` struct, probabbly better to use that?

crates/octez/src/node_config.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn set_network(&mut self, network: &str) -> &mut Self {
+        self.network = Some(network.to_owned());
+        self
+    }
+
+    /// Sets the HTTP(S) endpoint of the node RPC interface, e.g. 'http://localhost:8732'
+    pub fn set_rpc_endpoint(&mut self, endpoint: &str) -> &mut Self {
+        self.rpc_endpoint = Some(endpoint.to_owned());
+        self
+    }
```
ryutamago, line 142: isn't this `locahost:8732` instead of `http://localhost:8732`? imo using the `Endpoint` struct would be better to avoid such mistakes

huancheng-trili: This line was copied from octez_node.rs and that original line is updated in #589. I've corrected this line.

crates/octez/src/node_config.rs
huancheng-trili: I agree, but I would keep it as is for now and do this in a separate PR. Changing this requires moving the endpoint struct to octez crate and updating the octez node implementation, which are beyond the scope of this PR.

ryutamago (approved):
neater 💪

PR #589, opened 2024-10-04
jstzd: tezos-client activate protocol

ryutamago:
# Context

[activate protocol](https://linear.app/tezos/issue/JSTZ-135/implement-octezclientactivate-protocol)

# Description

```
  pub async fn activate_protocol(
        &self,
        protocol: &str,
        fitness: &str,
        key: &str,
        parameters_file: &Path,
    ) -> Result<()> 
```

# Manually testing the PR

Added integration test that tests the protocol and the block level before and after the activation


crates/jstzd/src/jstzd.rs
```diff
@@ -0,0 +1 @@
+pub const DEFAULT_RPC_ENDPOINT: &str = "localhost:8732";
```
huancheng-trili, line 1: This doesn't seem to be referenced anywhere?

nix/crates.nix
```diff
@@ -122,7 +122,7 @@ in {
 
     cargo-test-int = craneLib.cargoNextest (commonWorkspace
       // {
-        buildInputs = commonWorkspace.buildInputs ++ [octez pkgs.cacert];
+        buildInputs = commonWorkspace.buildInputs ++ [pkgs.iana-etc octez pkgs.cacert];
```
huancheng-trili, line 125: What's this for?

result-1
```diff
@@ -0,0 +1 @@
+/nix/store/ps7mj6p92ckwc2ypzb2ra6zmw28mg2gp-jstz
```
huancheng-trili, line 1: What does this do?

Cargo.toml
```diff
@@ -81,6 +81,7 @@ regex = "1"
 reqwest = { version = "0.11.24", features = ["json"] }
 reqwest-eventsource = "0.5.0"
 rustyline = "14.0.0"
+serial_test = "3.1.1"
```
huancheng-trili, line 84: Is this in use?

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -138,3 +144,77 @@ async fn imports_secret_key_throws() {
...
+            "ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK",
+            "0",
+            &activator,
+            &params_file,
+        )
+        .await;
+    assert!(protocol_activated.is_ok());
+    // 5. check if the protocol is activated and the block is baked.
+    let response = get_response_text(&blocks_head_endpoint).await;
+    assert!(response.contains(
```
huancheng-trili, line 189: I think it's better to look for `"next_protocol":"ProtoALphaALphaALphaALphaALphaALphaALphaALphaDdp3zK"` because you still get `"protocol":"ProtoGenesisGenesisGenesisGenesisGenesisGenesk612im"` when alpha is actually not activated but block 1 is baked.

nix/crates.nix
ryutamago: nix thing for the test to work, @johnyob added

crates/jstzd/src/jstzd.rs
ryutamago: used as a default value for the node endpoint when building the octez_client

result-1
zcabter: this file should be removed

nix/crates.nix
zcabter: Allows the query for `/etc/services` to work within the nix sandbox. This is required by cohttp lib, that octez client depends on.

PR #603, opened 2024-10-17
feat(jstzd): use Endpoint in octez node config

huancheng-trili:
# Context

Closes JSTZ-116.

# Description

Implement the suggestion in https://github.com/jstz-dev/jstz/pull/588#discussion_r1803429789. Also moved `Endpoint` to crate octez.

# Manually testing the PR

All existing tests should pass.

### DIFF
--- a/crates/octez/src/async_node.rs
+++ b/crates/octez/src/async_node.rs
@@ -2,7 +2,7 @@ use std::{fs::File, path::PathBuf, process::Stdio};
 
 use tokio::process::{Child, Command};
 
-use crate::OctezNodeRunOptions;
+use crate::{Endpoint, OctezNodeRunOptions};
 
 use super::path_or_default;
 use anyhow::Result;
@@ -23,8 +23,8 @@ impl AsyncOctezNode {
     pub async fn config_init(
         &self,
         network: &str,
-        rpc_endpoint: &str,
-        p2p_endpoint: &str,
+        rpc_endpoint: &Endpoint,
+        p2p_endpoint: &Endpoint,
         num_connections: u32,
     ) -> Result<Child> {
         Ok(self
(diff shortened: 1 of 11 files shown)


crates/octez/src/async_node.rs
```diff
@@ -36,12 +36,13 @@ impl AsyncOctezNode {
                 network,
                 "--data-dir",
                 self.octez_node_dir.to_str().expect("Invalid path"),
+                // the node executable does not accept the scheme part
                 "--rpc-addr",
-                rpc_endpoint,
+                &rpc_endpoint.to_string().split("://").last().unwrap(),
```
ryutamago, line 41: we can make a method that will create a string only for the host and port? would prob be useful

crates/octez/src/endpoint.rs
```diff
@@ -18,6 +18,12 @@ impl Endpoint {
     }
 }
 
+impl Default for Endpoint {
+    fn default() -> Self {
+        Self::localhost(80)
```
ryutamago, line 23: q: is 80 the default port for web server? was wondering if this is relevant to octez binaries

crates/octez/src/async_node.rs
huancheng-trili: Sure. Naming is the main problem. If something like `to_string_without_scheme` sounds okay, I can implement that.

crates/octez/src/endpoint.rs
huancheng-trili: This has nothing to do with octez binaries. 80 is the default port for http.

crates/octez/src/async_node.rs
ryutamago: perhaps authority?
https://developer.mozilla.org/en-US/docs/Web/URI/Authority

huancheng-trili: That's not necessarily enough since one can set up a node behind something such that it's accessible at a certain path

ryutamago: right, could go without_scheme then!

huancheng-trili: Sorry I was just overthinking. Authority should be enough.

crates/octez/src/endpoint.rs
```diff
@@ -16,6 +16,16 @@ impl Endpoint {
             port,
         }
     }
+
+    pub fn get_authority(&self) -> String {
```
ryutamago, line 20: nit:
`get_` isn't used in rust convention IIRC, what about `to_authority_string`?

ryutamago (approved):
just one nit, LGTM 💪

PR #607, opened 2024-10-18
fix(sandbox): activate parisC in sandbox

zcabter:
# Context
Bake for command is available only on latest protocols of a 'version' of the protocol. Since we upgraded to the latest octez release, ParisC is now the latest protocol

# Description

# Manually testing the PR
`cargo run -- sandbox start`


zcabter:
Code coverage is pretty much dead as we don't want to spend resources testing the old sandbox.

huancheng-trili (approved):

PR #610, opened 2024-10-21
feat(jstzd): originate contract

huancheng-trili:
# Context

Closes JSTZ-140.

[JSTZ-140](https://linear.app/tezos/issue/JSTZ-140/implement-octezclientoriginate-contract)

# Description

Implements `octez-client originate contract`. If the execution is successful, the contract address and the hash of the operation where the origination took place are returned.

# Manually testing the PR

* Unit test: new test cases
```sh
$ cargo test --lib task::octez_client::test::parse_

running 6 tests
test task::octez_client::test::parse_operation_hash_pattern_mismatch ... ok
test task::octez_client::test::parse_contract_address_pattern_mismatch ... ok
test task::octez_client::test::parse_operation_hash ... ok
test task::octez_client::test::parse_contract_address_bad_hash ... ok
test task::octez_client::test::parse_operation_hash_bad_hash ... ok
test task::octez_client::test::parse_contract_address ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.01s
```
* Integration test: one new test case

### DIFF
--- a/crates/octez/src/async/client.rs
+++ b/crates/octez/src/async/client.rs
@@ -11,6 +11,7 @@ use std::{
     str::FromStr,
 };
 use tempfile::tempdir;
+use tezos_crypto_rs::hash::{ContractKt1Hash, OperationHash};
 use tokio::process::Command;
 
 use super::{directory::Directory, endpoint::Endpoint, node::DEFAULT_RPC_ENDPOINT};
@@ -331,6 +332,80 @@ impl OctezClient {
         self.spawn_and_wait_command(args).await?;
         Ok(())
     }
+
+    pub async fn originate_contract(
+        &self,
+        contract_alias: &str,
+        funding_account_src: &str,
+        fund_tez: f64,
+        contract_path: &Path,
+        init_data: Option<&str>,
+        burn_cap: Option<f64>,
+    ) -> Result<(ContractKt1Hash, OperationHash)> {
+        let fund_str = fund_tez.to_string();
+        let burn_cap_str = burn_cap.map(|v| v.to_string());
+        let mut args = vec![
+            "originate",
+            "contract",
+            contract_alias,
+            "transferring",
+            &fund_str,
+            "from",
+            funding_account_src,
+            "running",
+            contract_path
+                .to_str()
+                .ok_or(anyhow::anyhow!("failed to convert contract path to string"))?,
+        ];
+        if let Some(data) = init_data {
+            args.append(&mut vec!["--init", data]);
+        }
+        if let Some(v) = &burn_cap_str {
+            args.append(&mut vec!["--burn-cap", v]);
+        }
+        let output = self.spawn_and_wait_command(args).await?;
+
+        let operation_hash = parse_operation_hash(&output);
+        if let Err(e) = operation_hash {
+            anyhow::bail!(
+                "failed to parse operation hash from execution output: {:?}",
+                e
+            );
+        }
+        let contract_address = parse_contract_address(&output);
+        if let Err(e) = contract_address {
+            anyhow::bail!(
+                "failed to parse contract address from execution output: {:?}",
+                e
+            );
+        }
+        Ok((contract_address.unwrap(), operation_hash.unwrap()))
+    }
+}
+
+fn parse_regex(pattern_str: &str, output: &str) -> Result<String> {
+    let pattern = regex::Regex::new(pattern_str)?;
+    Ok(pattern
+        .captures(output)
+        .ok_or(anyhow::anyhow!("input string does not match the pattern"))?
+        .get(1)
+        .ok_or(anyhow::anyhow!("cannot find the first match group"))?
+        .as_str()
+        .to_owned())
+}
+
+fn parse_operation_hash(output: &str) -> Result<OperationHash> {
+    let raw_operation_hash =
+        parse_regex("Operation hash is '(o[1-9A-HJ-NP-Za-km-z]{50})'", output)?;
+    Ok(OperationHash::from_base58_check(&raw_operation_hash)?)
+}
+
+fn parse_contract_address(output: &str) -> Result<ContractKt1Hash> {
+    let raw_contract_hash = parse_regex(
+        "New contract (KT1[1-9A-HJ-NP-Za-km-z]{33}) originated.",
+        output,
+    )?;
+    Ok(ContractKt1Hash::from_base58_check(&raw_contract_hash)?)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,81 @@ impl OctezClient {
...
+                .to_str()
+                .ok_or(anyhow::anyhow!("failed to convert contract path to string"))?,
+        ];
+        if let Some(data) = init_data {
+            args.append(&mut vec!["--init", data]);
+        }
+        if !burn_cap_str.is_empty() {
+            args.append(&mut vec!["--burn-cap", &burn_cap_str]);
+        }
+        let output = self.spawn_and_wait_command(args).await?;
```
zcabter, line 369: `burn_cap` is provided as an Optional arg, there is no need to transform it to a string on line 346 then test its presence with `is_empty`? Please could you use `if let Some` here instead and remove line 346-348.

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,81 @@ impl OctezClient {
...
+
+        let operation_hash = parse_operation_hash(&output);
+        if operation_hash.is_none() {
+            return Err(anyhow::anyhow!(
+                "failed to parse operation hash from execution output"
+            ));
```
zcabter, line 375: Please use `bail!`

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,81 @@ impl OctezClient {
...
+        }
+        let contract_address = parse_contract_address(&output);
+        if contract_address.is_none() {
+            return Err(anyhow::anyhow!(
+                "failed to parse contract address from execution output"
+            ));
```
zcabter, line 381: please use `bail!`

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,81 @@ impl OctezClient {
...
+    if let Some(capture) = operation_hash_pattern.captures(output) {
+        let raw_operation_hash = capture.get(1).unwrap().as_str();
+        if let Ok(v) = OperationHash::from_base58_check(raw_operation_hash) {
+            return Some(v);
+        }
+    }
+    None
+}
+
+fn parse_contract_address(output: &str) -> Option<ContractKt1Hash> {
```
zcabter, line 399: Seems like a lot of repetition between `parse_contract_address` and `parse_operation_hash`. You can factorize it into:
 1. Extract regex
 2. Parse as hash

crates/octez/src/async/client.rs
```diff
@@ -516,4 +592,58 @@ mod test {
         let res = OctezClient::extract_digits(input);
         assert!(res.is_ok_and(|balance| balance == 15.23453));
     }
+
+    #[test]
+    fn parse_contract_address() {
+        let contract_hash = "KT1F3MuqvT9Yz57TgCS3EkDcKNZe9HpiavUJ";
+        assert_eq!(
+            super::parse_contract_address(&format!(
```
zcabter, line 600: Isn't `parse_contract_address` already imported in `use::super::*`? There shouldn't be a need to 
use the full qualified name. 

Applicable to the other tests below

crates/octez/src/async/client.rs
huancheng-trili: It's like this because rustc complains that the option string being pushed does not live long enough if I use `if let Some`.

crates/octez/src/async/client.rs
huancheng-trili: Updated

crates/octez/src/async/client.rs
huancheng-trili: Updated

crates/octez/src/async/client.rs
huancheng-trili: I wouldn't call this "a lot of repetition" and extracting the common parts does not make things much more readable, but I've updated it anyway.

crates/octez/src/async/client.rs
huancheng-trili: The test case itself is named `parse_contract_address` and thus collides with the imported method. I've renamed those test cases even though I don't think prefixing `super::` matters that much here.

crates/octez/src/async/client.rs
zcabter: Ah :(

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,83 @@ impl OctezClient {
...
+        Ok((contract_address.unwrap(), operation_hash.unwrap()))
+    }
+}
+
+fn parse_regex(pattern_str: &str, output: &str) -> Option<String> {
+    let pattern = regex::Regex::new(pattern_str).unwrap();
+    if let Some(capture) = pattern.captures(output) {
+        return Some(capture.get(1).unwrap().as_str().to_owned());
+    }
+    None
```
zcabter, line 388: Nit: Probably could use `?` instead of explicitly matching the Some, so could so something like:
`pattern.captures(output)?.get(1)?.as_str().to_string()`

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,83 @@ impl OctezClient {
...
+fn parse_contract_address(output: &str) -> Option<ContractKt1Hash> {
+    if let Some(raw_contract_hash) = parse_regex(
+        "New contract (KT1[1-9A-HJ-NP-Za-km-z]{33}) originated.",
+        output,
+    ) {
+        if let Ok(v) = ContractKt1Hash::from_base58_check(&raw_contract_hash) {
+            return Some(v);
+        }
+    }
+    None
```
zcabter, line 411:
```suggestion
    let raw_operation_hash =
        parse_regex("Operation hash is '(o[1-9A-HJ-NP-Za-km-z]{50})'", output)?;
    OperationHash::from_base58_check(&raw_operation_hash).ok()
```

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,83 @@ impl OctezClient {
...
+        let contract_address = parse_contract_address(&output);
+        if contract_address.is_none() {
+            anyhow::bail!("failed to parse contract address from execution output");
+        }
+        Ok((contract_address.unwrap(), operation_hash.unwrap()))
+    }
+}
+
+fn parse_regex(pattern_str: &str, output: &str) -> Option<String> {
+    let pattern = regex::Regex::new(pattern_str).unwrap();
```
zcabter, line 384:
```suggestion
    let pattern = regex::Regex::new(pattern_str).ok()?;
```

crates/octez/src/async/client.rs
ryutamago: the client also has a logic that parses the tez balance printed out by the client as well(e.g. 100 ꜩ)
 What about we create a struct for parsing client stdout that contains all these logic ?

crates/octez/src/async/client.rs
huancheng-trili: Updated

crates/octez/src/async/client.rs
huancheng-trili: Updated. Sorry I'm not sure why I kept the awkward `if let` like that

crates/octez/src/async/client.rs
huancheng-trili: Updated

crates/octez/src/async/client.rs
huancheng-trili: Yup. That would be in [JSTZ-132](https://linear.app/tezos/issue/JSTZ-132/implement-ascommand-trait-helpers) and [JSTZ-133](https://linear.app/tezos/issue/JSTZ-133/implement-ascommand-for-octezclient).

crates/octez/src/async/client.rs
zcabter: Hmm in that case, would it be useful to return Result instead of Option because the caller of `parse_` expects the output to contain the regex and it should be a failure otherwise.

crates/octez/src/async/client.rs
```diff
@@ -331,6 +332,71 @@ impl OctezClient {
         self.spawn_and_wait_command(args).await?;
         Ok(())
     }
+
+    pub async fn originate_contract(
+        &self,
+        contract_alias: &str,
+        funding_account_src: &str, // can be an address or an alias
```
zcabter, line 339:
```suggestion
        funding_account_src: &str,
```

crates/octez/src/async/client.rs
huancheng-trili: Okay I've changed this back to Result. I thought not finding anything should not be an error from the point of view of `parse_regex` itself, but I suppose in this case specifically we can consider that an error.

crates/octez/src/async/client.rs
huancheng-trili: Updated

zcabter (approved):

PR #611, opened 2024-10-22
feat(jstzd): implement transfer

huancheng-trili:
# Context

Closes JSTZ-138.

[JSTZ-138](huanchengchang-jstz-138)

# Description

Implement `octez-client transfer`.

# Manually testing the PR

* Integration test: one new test case

### DIFF
--- a/crates/jstzd/src/task/octez_client.rs
+++ b/crates/jstzd/src/task/octez_client.rs
@@ -384,6 +384,24 @@ impl OctezClient {
         }
         Ok((contract_address.unwrap(), operation_hash.unwrap()))
     }
+
+    pub async fn transfer(
+        &self,
+        src: &str,
+        dst: &str,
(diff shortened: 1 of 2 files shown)


crates/jstzd/src/task/octez_client.rs
```diff
@@ -380,6 +386,24 @@ impl OctezClient {
         }
         Ok((contract_address.unwrap(), operation_hash.unwrap()))
     }
+
+    pub async fn transfer(
```
ryutamago, line 390: FYI i'm using the transfer command in call contract too, wondering if we should just have one general function to transfer / call contract?

huancheng-trili: sure

ryutamago:
closing this as there are duplication w contract call

PR #612, opened 2024-10-22
fix(jstzd): hold tempdir in octez node

huancheng-trili:
# Context

Completes JSTZ-116.

# Description

Similar to what was implemented in #572, `OctezNode` will hold the created temporary directory and file instances so that they don't get removed too early but do get cleaned up after the node is dropped. 

# Manually testing the PR

* Unit test: added test cases
* Integration test: existing test cases still work. If the temporary directories were dropped, the tests would also fail because of inconsistent node states.


crates/octez/src/async/baker.rs
```diff
@@ -69,18 +66,6 @@ impl OctezBakerConfigBuilder {
         self
     }
 
-    pub fn with_node_and_client(
```
ryutamago, line 72: what's the intention behind removing this? i think the motivation was to have it for convenicince. What if we pass 
```
with_node_and_client(node: OctezNode, client: OctezClient) 
```

crates/jstzd/tests/octez_baker_test.rs
```diff
@@ -55,24 +61,22 @@ async fn test_baker() {
     let _ = octez_node.kill().await;
 }
 
-async fn spawn_octez_node() -> (octez_node::OctezNode, TempDir) {
-    let temp_dir = TempDir::new().unwrap();
-    let data_dir = temp_dir.path();
+async fn spawn_octez_node() -> jstzd::task::octez_node::OctezNode {
```
ryutamago, line 64: nit import `jstzd::task::octez_node::OctezNode`

crates/jstzd/src/task/file.rs
```diff
@@ -0,0 +1,104 @@
+use std::fs::File;
+use std::path::PathBuf;
+
+use anyhow::Result;
+use tempfile::NamedTempFile;
+
+#[derive(Debug)]
+pub enum FileWrapper {
```
ryutamago, line 8: nit we could just name it as `File` similar to `Directory`?

crates/jstzd/src/task/octez_node.rs
```diff
@@ -1,25 +1,31 @@
...
+use std::sync::Arc;
 
 #[derive(Default, Clone)]
 pub struct OctezNode {
     inner: SharedChildWrapper,
     config: OctezNodeConfig,
+    // holds the TempDir instance so that the directory does not get deleted too soon
+    _data_dir: Arc<Directory>,
+    // holds the NamedTempFile instance so that the file does not get deleted too soon
+    _log_file: Arc<FileWrapper>,
```
ryutamago, line 19: what about having the `OctezNodeConfig` own the File and Directory? that way we can avoid having these fields and the baker can take both the `OctezNode` and `OctezClient` to initialize

crates/jstzd/tests/octez_baker_test.rs
huancheng-trili: There's too much potential confusion now with OctezNode in crate jstzd and crate octez. I prefer to keep the fully-qualified name.

crates/jstzd/src/task/file.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl TryFrom<PathBuf> for FileWrapper {
+    type Error = anyhow::Error;
+
+    fn try_from(path: PathBuf) -> Result<Self> {
+        Ok(FileWrapper::File((
+            File::options()
+                .read(true)
+                .write(true)
+                .truncate(true)
```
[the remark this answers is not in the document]
huancheng-trili: This is basically equivalent to what it used to be, i.e. `File::create`, creating the log file regardless of its existence. I'm setting these options manually just to make the file instance readable. `File::create` returns a file instance that is only writable.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: I've tried all these yesterday. I don't think `octez::r#async::node:OctezNode` should own the file or the directory since it's simply a wrapper for the executable, and neither should the config because the config carrying file instances sounds way too much responsibility for config, which imo should not be more than a k-v mapping, not to mention that the carried file needs to be in an arc to make the config clonable. `jstzd::task::octez_node::OctezNode`, on the other hand, calls the node executable and thus should manage everything around that, including cleaning up the files or so, and therefore should own the file and the directory. `OctezBaker::with_node_and_client` with the node is better, but it requires access to `octez::r#async::node:OctezNode` instances, which means they need to be somehow exposed when they are not directly accessible, e.g. when there are only `jstzd::task::octez_node::OctezNode` and the node config. Keeping the three setters is sufficient to me without entangling these implicit dependencies for a bit of convenience.

crates/octez/src/async/baker.rs
huancheng-trili: See https://github.com/jstz-dev/jstz/pull/612#discussion_r1816703016

crates/jstzd/src/task/file.rs
huancheng-trili: There are already too many `File`s in this module, so I decided to add `Wrapper` to avoid confusion.

crates/jstzd/src/task/octez_node.rs
huancheng-trili: Builidng `OctezBakerConfig` from `OctezNodeConfig` and `OctezClient` becomes impossible because `OctezNodeConfig` does not know where the temporary directory is. It used to work because the temporary path was determined by the config builder. Now the temporary path is managed by `jstzd::task::octez_node::OctezNode` and since `OctezNodeConfig` needs to be built before `jstzd::task::octez_node::OctezNode` is spawned, it has no idea what the temporary path is.

There might be some other way of arranging these things and what I'm doing here now is not necessarily tidy enough of course.

PR #613, opened 2024-10-23
refactor(jstzd): refactor tests

huancheng-trili:
# Context

Pulling common parts out of integration test cases since we have too many of them now.

# Description

Test cases for octez-client usually require setting up the node and the client. These steps are pretty much identical for those test cases and thus can be moved into the utils module, which makes those test cases neater.

Somehow clippy complains that the utils module is not used, so `#![allow(dead_code)]` is added to that entire module.

# Manually testing the PR

All existing test cases should pass.

### DIFF
--- a/crates/jstzd/tests/utils.rs
+++ b/crates/jstzd/tests/utils.rs
@@ -1,3 +1,18 @@
+#![allow(dead_code)]
+use jstzd::task::{octez_baker, octez_node::OctezNode, Task};
+use octez::r#async::{
+    baker::{BakerBinaryPath, OctezBakerConfigBuilder},
+    client::{OctezClient, OctezClientBuilder},
+    endpoint::Endpoint,
+    node_config::{OctezNodeConfigBuilder, OctezNodeRunOptionsBuilder},
+    protocol::Protocol,
+};
+use regex::Regex;
+use std::path::{Path, PathBuf};
+
+pub const SECRET_KEY: &str =
+    "unencrypted:edsk31vznjHSSpGExDMHYASz45VZqXN4DPxvsa4hAyY8dHM28cZzp6";
+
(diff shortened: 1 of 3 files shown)


crates/jstzd/tests/utils.rs
```diff
@@ -14,7 +29,129 @@ where
     false
 }
 
-#[allow(dead_code)]
+pub async fn setup() -> (OctezNode, OctezClient, octez_baker::OctezBaker) {
```
ryutamago, line 32: 💯🧹

crates/jstzd/tests/utils.rs
```diff
@@ -14,7 +29,129 @@ where
...
+        .map(|level_match| level_match.as_str().parse::<i32>().unwrap())
+        .unwrap()
+}
+
+pub async fn import_bootstrap_keys(octez_client: &OctezClient) {
+    for (idx, key) in [
+        "unencrypted:edsk3gUfUPyBSfrS9CCgmCiQsTCHGkviBDusMxDJstFtojtc1zcpsh",
+        "unencrypted:edsk39qAm1fiMjgmPkw1EgQYkMzkJezLNewd7PLNHTkr6w9XA2zdfo",
+        "unencrypted:edsk4ArLQgBTLWG5FJmnGnT689VKoqhXwmDPBuGx3z4cvwU9MmrPZZ",
+        "unencrypted:edsk2uqQB9AY4FvioK2YMdfmyMrer5R8mGFyuaLLFfSRo8EoyNdht3",
```
ryutamago, line 120: nit: i dodn't include in my previous pr but there's another bootstrap5 account  w "unencrypted:edsk4QLrcijEffxV31gGdN2HU7UpyJjA8drFoNcmnB28n89YjPNRFm"

huancheng-trili: added that account

PR #616, opened 2024-10-25
feat(jstzd): implement skeleton of jstz server

huancheng-trili:
# Context

Closes JSTZ-162.
[JSTZ-162](https://linear.app/tezos/issue/JSTZ-162/initialise-jstzd-as-a-wrapper-over-octeznode)

# Description

Creates `Jstzd` that wraps over `OctezNode` and `JstzdServer` that spins up `Jstzd`.
Note that `JstzdServer::run` spins up `Jstzd` and also starts the server. This method returns while the other things keep running. Everything can be shut down with `JstzdServer::stop`.

# Manually testing the PR

* Integration test: added one test case
```sh
$ cargo test --test jstzd_test

running 1 test
< octez logs >
test jstzd_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.69s

### CODE REVIEW

File: crates/jstzd/src/task/jstzd.rs [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+        if let Some(jstzd) = &self.jstzd {
+            if let Ok(v) = jstzd.health_check().await {
+                return v;
+            }
+        }
+
+        false
+    }
+
+    pub async fn stop(&mut self) -> anyhow::Result<()> {
```
ryutamago, line 86: nit: 
remove anyhow

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,117 @@
...
+        if let Some(server) = self.server_handle.take() {
+            server.abort();
+        }
+        match err {
+            Some(e) => Err(e),
+            None => Ok(()),
+        }
+    }
+
+    pub async fn run(&mut self) -> anyhow::Result<()> {
```
ryutamago, line 102: nit: 
remove anyhow

crates/jstzd/src/task/jstzd.rs
huancheng-trili: done

crates/jstzd/src/task/jstzd.rs
huancheng-trili: done

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let router =
+            axum::Router::new().route("/", axum::routing::get(http::StatusCode::OK));
+        let listener =
+            tokio::net::TcpListener::bind(("0.0.0.0", self.jstzd_server_port)).await?;
+
+        let handle = tokio::spawn(async {
+            axum::serve(listener, router).await.unwrap();
+        });
+        self.server_handle.replace(handle);
```
ryutamago, line 114: don't we need to await on the handle here? otherwise the server will shutdown if the main fnction exits.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,117 @@
...
+        self.jstzd
+            .replace(Jstzd::spawn(self.jstzd_config.clone()).await?);
+
+        let router =
+            axum::Router::new().route("/", axum::routing::get(http::StatusCode::OK));
+        let listener =
+            tokio::net::TcpListener::bind(("0.0.0.0", self.jstzd_server_port)).await?;
+
+        let handle = tokio::spawn(async {
+            axum::serve(listener, router).await.unwrap();
```
ryutamago, line 112: would also be nice to test with `curl` and see if the server spins up properly . Would be nice to use `with_graceful_shutdown()` to handle the shutdown with `Crtl C` without forecefully shuting down the process

crates/jstzd/src/task/jstzd.rs
huancheng-trili: No. According to the [doc](https://docs.rs/tokio/latest/tokio/task/fn.spawn.html),

> The provided future will start running in the background immediately when spawn is called, even if you don’t await the returned JoinHandle.

If this handle is awaited, this method only returns when the server is shutdown. Testing is all included in the new integration test.

crates/jstzd/src/task/jstzd.rs
huancheng-trili: There's no need for gracefully shutting down this server since it's pretty much stateless and all it does is to serve the configs. Even if it needs to be shut down gracefully, it should be implemented at jstzd main and the trap there handles closing everything.

crates/jstzd/src/task/jstzd.rs
ryutamago: i understand the server starts correctly based on the integration test. i was wondering say if we want to spin up the server like the following code:
```
#[tokio]
fn main () {
  // this will return immediately and main would exit as we don't await on the join handler
  JstzdServer::run().await;
}
```
then the spawned jstzd would be shutdown as the main function would exit immeidiately?
The doc says:

> There is no guarantee that a spawned task will execute to completion. When a runtime is shutdown, all outstanding tasks are dropped, regardless of the lifecycle of that task.

huancheng-trili: Yes, then the question is how this jstzd server will be launched and this is totally up to discussions. If we just want to run the server like how most servers are run, we just need call `axum::serve` and don't return the handle at all, which will work with the code snippet. However, from what I see now in the issues defined so far, it seems that in jstzd main there will be a readiness check after jstzd server is launched, kind of like the following based on what I understand:

* start jstzd server
* wait until all the components are up and running
  * ok: inform users that they are good to go and then keep everything running until users terminate the sandbox
  * timeout: shut down everything and show errors

Then in this case we won't run the server as what you described but somehow keep the join handle alive.

ryutamago: > ok: inform users that they are good to go and then keep everything running until users terminate the sandbox

In this case, don't we still need a way to await on the join handler? maybe we can add a method that waits for the join handler?

huancheng-trili: Not really. As long as the join handler does not get dropped, the task keeps running. I imagine that jstzd main would be like
```rs
let handle = jstzd.run(...)
tokio::select!{
  // listen to signals and cancel the handle or call stop on request
}
```

If we change the implementation so that it's like `jstzd_server.run().await;`, it should also work. It's just that in main there will be a `tokio::spawn` if we want to do something else after `server.run().await`. I'm not sure about this because then the server instance will be moved into spawn and I'm not sure if rust will complain when we try to do call that instance.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,117 @@
...
+            None => Ok(()),
+        }
+    }
+
+    pub async fn run(&mut self) -> Result<()> {
+        self.jstzd
+            .replace(Jstzd::spawn(self.jstzd_config.clone()).await?);
+
+        let router =
+            axum::Router::new().route("/", axum::routing::get(http::StatusCode::OK));
```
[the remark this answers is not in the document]
huancheng-trili: This `/` route is just a placeholder for now. Those other endpoints are out of the scope of this PR and ticket and will be implemented later on.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,117 @@
+use super::{octez_node::OctezNode, Task};
+use anyhow::Result;
+use async_dropper_simple::AsyncDrop;
+use async_trait::async_trait;
+use octez::r#async::node_config::OctezNodeConfig;
+use tokio::task::JoinHandle;
+
+pub struct Jstzd {
```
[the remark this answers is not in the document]
huancheng-trili: Updated

crates/jstzd/tests/jstzd_test.rs
```diff
@@ -0,0 +1,74 @@
...
+    let run_options = run_option_builder
+        .set_synchronisation_threshold(0)
+        .set_network("sandbox")
+        .build();
+    let mut config_builer = OctezNodeConfigBuilder::new();
+    config_builer
+        .set_binary_path("octez-node")
+        .set_network("sandbox")
+        .set_rpc_endpoint(&rpc_endpoint)
+        .set_run_options(&run_options);
```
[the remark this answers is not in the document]
huancheng-trili: Yes. I've removed these.

crates/jstzd/tests/jstzd_test.rs
```diff
@@ -0,0 +1,74 @@
...
+use utils::retry;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn jstzd_test() {
+    let mut run_option_builder = OctezNodeRunOptionsBuilder::new();
+    let rpc_endpoint = Endpoint::localhost(unused_port());
+    let run_options = run_option_builder
+        .set_synchronisation_threshold(0)
+        .set_network("sandbox")
+        .build();
```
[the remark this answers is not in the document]
huancheng-trili: Removed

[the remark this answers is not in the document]
huancheng-trili: There's actually a bug related to `OctezNodeRunOptions`. That is fixed in #640.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,115 @@
...
+    type Config = JstzdConfig;
+
+    async fn spawn(config: Self::Config) -> Result<Self> {
+        Ok(Self {
+            octez_node: OctezNode::spawn(config.octez_node_config.clone()).await?,
+        })
+    }
+
+    async fn kill(&mut self) -> Result<()> {
+        self.octez_node.kill().await
```
zcabter, line 34: Note: When the other octez processes are handled, we will want to kill them concurrently.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,115 @@
...
+        self.jstzd
+            .replace(Jstzd::spawn(self.jstzd_config.clone()).await?);
+
+        let router =
+            axum::Router::new().route("/", axum::routing::get(http::StatusCode::OK));
+        let listener =
+            tokio::net::TcpListener::bind(("0.0.0.0", self.jstzd_server_port)).await?;
+
+        let handle = tokio::spawn(async {
+            axum::serve(listener, router).await.unwrap();
```
zcabter, line 110: Nit: My preference for conciseness/readability is to avoid qualified references in general and reserve them only when calling module level functions or disambiguate from the current context norm. 

So here I would use `Router` and `TcpListener` because devs should know that we are using `axum` and `tokio` as our runtimes but use `axum::serve` as it a module level function.

crates/jstzd/src/task/jstzd.rs
zcabter: I initially scoped this function to do the run checks and wait for sigs but splitting the server orchestration and web server launch makes sense. Could you create a linear issue to track the server orchestration work please?

crates/jstzd/tests/jstzd_test.rs
```diff
@@ -0,0 +1,64 @@
...
+use octez::r#async::endpoint::Endpoint;
+use octez::r#async::node_config::OctezNodeConfigBuilder;
+use octez::unused_port;
+use utils::retry;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn jstzd_test() {
+    let rpc_endpoint = Endpoint::localhost(unused_port());
+    let mut config_builer = OctezNodeConfigBuilder::new();
+    config_builer.set_rpc_endpoint(&rpc_endpoint);
```
zcabter, line 11: Nit: 
```suggestion
    let mut config_builer = OctezNodeConfigBuilder::new().set_rpc_endpoint(&rpc_endpoint);
```

crates/jstzd/src/task/jstzd.rs
```diff
@@ -0,0 +1,115 @@
...
+            server_handle: None,
+        }
+    }
+
+    pub async fn health_check(&self) -> bool {
+        let res =
+            reqwest::get(format!("http://localhost:{}", self.jstzd_server_port)).await;
+        if res.is_err() {
+            return false;
+        }
```
zcabter, line 73: Making a request to itself is quite odd. Maybe check `server_handle`  instead?

crates/jstzd/src/task/jstzd.rs
huancheng-trili: Updated

crates/jstzd/tests/jstzd_test.rs
huancheng-trili: `OctezNodeConfigBuilder::new()` needs to be held at a temporary place so that it lives long enough. I've replaced this with `OctezNodeConfigBuilder::new()....build()`.

crates/jstzd/src/task/jstzd.rs
huancheng-trili: I think it's fine. I'm not quite sure about checking if `server_handle` is finished since it might be possible that somehow the server gets stuck but the task is still running, though the chance is relatively low since this server is pretty much serving static content only.

crates/jstzd/tests/jstzd_test.rs
zcabter: Ah, the builder patter was implemented with `&mut self` instead of `mut self`... I think a consuming builder would have been nicer to work with since octez configurations are relatively simple

crates/jstzd/src/task/jstzd.rs
zcabter: Yeah, I reckon that would hardly happen on the jstzd server.

And I guess the `health_check` would be bound to something like `/health` which means that handling the endpoint proves that is isn't stuck?

huancheng-trili: Alright. I've updated this check.

crates/jstzd/src/task/jstzd.rs
huancheng-trili: https://linear.app/tezos/issue/JSTZ-189

PR #618, opened 2024-10-28
feat(jstz_node): axum migration

zcabter:
# Context
Migrates the jstz node runtime from `actix_web` to `axum`.

Historically, `actix_web` used an actor based design to for their concurrency model. As time has proven itself, green threads have become the de-facto way to implement concurrency within the rust ecosystem in particular using `tokio`. As a result, `actix` implement its own wrappers over `tokio` to support both the actor and green threading model which means the using tokio directly for task management is not easy. 

This migration makes jstz align on `tokio` throughout the cli, node, and jstzd. This also lays the ground work for [client binding generation through open api.](https://linear.app/tezos/project/generated-bindings-forjstz-node-06f8bdb66c4e/overview)

Closes: [JSTZ-85](https://linear.app/tezos/issue/JSTZ-85/switch-to-using-axum-in-jstz-node)

# Description
The main changes relate to how routes and app state get registered. 

# Manually testing the PR
Test coverage is 0 due to the change touching the whole module which is already untested. We will need progress here soon.

Additionally, the kernel is currently broken due to a yet to be understood bug in the stack below jstz, making it difficult to test these changes.


crates/jstz_node/src/services/logs/mod.rs
```diff
@@ -122,42 +128,29 @@ mod persistent_logging {
...
-    Ok(broadcaster.new_client(address).await)
-}
+use super::error::{ServiceError, ServiceResult};
 
 pub struct LogsService;
 
-impl Service for LogsService {
-    fn configure(cfg: &mut ServiceConfig) {
-        let scope = Scope::new("/logs").service(stream_logs);
+impl LogsService {}
```
huancheng-trili, line 140: Is this necessary?

crates/jstz_node/src/services/logs/broadcaster.rs
```diff
@@ -97,11 +102,18 @@ impl Broadcaster {
         if let Some(clients) = clients.get(contract_address) {
             let send_futures = clients
                 .iter()
-                .map(|client| client.send(Data::new(msg).into()));
-
-            // try to send to all clients, ignoring failures
-            // disconnected clients will get swept up by `remove_stale_clients`
+                .map(|client| client.send(Ok(sse::Event::default().data(msg))));
+            // // try to send to all clients, ignoring failures
```
huancheng-trili, line 106:
```suggestion
            // try to send to all clients, ignoring failures
```

crates/jstz_node/src/services/logs/mod.rs
```diff
@@ -122,42 +128,29 @@ mod persistent_logging {
...
 
-        #[cfg(not(feature = "persistent-logging"))]
-        cfg.service(scope);
+impl Service for LogsService {
+    fn router() -> Router<AppState> {
+        let routes = Router::new().route("/{address}/stream", get(stream_log));
```
huancheng-trili, line 144: Does `/{...}` also work with axum? They only mention `/:` in the [doc](https://docs.rs/axum/latest/axum/struct.Router.html#captures).

crates/jstz_node/src/services/logs/mod.rs
zcabter: Nope, removed.

crates/jstz_node/src/services/logs/broadcaster.rs
zcabter: Done

crates/jstz_node/src/services/logs/mod.rs
zcabter: Oops, missed it. Updated.

huancheng-trili (approved):

PR #619, opened 2024-10-29
feat(openapi): switch to openapi router and add scalar API docs to js…

zcabter:
# Context
This PR makes the `OpenApiRouter` as the base router for axum routes and introduces the `ApiDoc` type that will hold top level information about the OAS spec. This PR is in preparation for:

* https://linear.app/tezos/issue/JSTZ-158/implement-openapi-for-operationsservice
* https://linear.app/tezos/issue/JSTZ-157/implement-openapi-for-accountservice
* https://linear.app/tezos/issue/JSTZ-159/implement-openapi-for-logservice

Closes:  [JSTZ-156](https://linear.app/tezos/issue/JSTZ-156/implement-openapi-router-base-struct)

# Description
Besides the changes above, this PR also regression checks the generated `openapi.json` spec. A `jstz-node spec` command has been implemented to make it convenient for generating the spec.

# Manual Testing
Spec can be tested with:
```
cargo test -p jstz_node
```
Scalar docs:
```
cargo run -- sandbox start
# browse to http://localhost:8933/scalar
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build
         run: nix --accept-flake-config --log-format raw -L build -j auto .#{all,js_jstz}
       - name: Flake check
-        run: nix --accept-flake-config --log-format raw -L flake check -j auto
+        run: nix --accept-flake-config --log-format raw --log-lines 70 -L flake check -j auto
       # Coverage is part of nix flake check, but we want to upload it to Codecov
       # So we run it again (it's cached) and upload the result
       - name: Coverage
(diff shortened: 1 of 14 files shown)


huancheng-trili (commented):
Looks fine to me. So the next step for those 3 tickets is to apply the `#[utoipa::path]` macro to route handlers and update the way they are registered?

zcabter:
> Looks fine to me. So the next step for those 3 tickets is to apply the `#[utoipa::path]` macro to route handlers and update the way they are registered?

Yes, and additionally to implement their schemas (where derivation is not possible)

huancheng-trili (approved):

PR #623, opened 2024-10-30
remove default trait, endpoint for octez node

ryutamago:
# Context

 It’s simpler to make the node endpoint of the client builder required so we don’t have to think about the default behavior. 
 Will make a separate PR for the baker 
 
# Description

* clean up the config builder 
* the client will not have a fallback default endpoint

# Manually testing the PR

existing tests pass

### DIFF
--- a/crates/jstzd/tests/utils.rs
+++ b/crates/jstzd/tests/utils.rs
@@ -88,8 +88,7 @@ pub async fn spawn_octez_node() -> OctezNode {
 }
 
 pub fn create_client(node_endpoint: &Endpoint) -> OctezClient {
-    OctezClientBuilder::new()
-        .set_endpoint(node_endpoint.clone())
+    OctezClientBuilder::new(node_endpoint.clone())
         .build()
         .unwrap()
 }
--- a/crates/octez/src/async/client.rs
+++ b/crates/octez/src/async/client.rs
@@ -1,5 +1,4 @@
 use anyhow::{anyhow, bail, Context, Result};
-use http::Uri;
 use jstz_crypto::{
     public_key::PublicKey, public_key_hash::PublicKeyHash, secret_key::SecretKey,
 };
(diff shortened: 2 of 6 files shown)


zcabter:
Why shouldn't the node fallback to a random port?

crates/jstzd/tests/utils.rs
```diff
@@ -1,11 +1,14 @@
 #![allow(dead_code)]
```
zcabter, line 1: thread

ryutamago:
> Why shouldn't the node fallback to a random port?

1. The binary falls back to localhost:8732 so it's consistent
2. the client also falls back to node endpoint w localhost:8732 and rollup should follow the same pattern
3. we added the unused port just for the parallel test to pass and there would be only one isntance of octez for user anyway and they might find it confusing if the octez node is running on differnt port

crates/jstzd/tests/utils.rs
ryutamago: ?

zcabter:
> > Why shouldn't the node fallback to a random port?
> 
> 1. The binary falls back to localhost:8732 so it's consistent
> 2. the client  / baker also falls back to node endpoint w localhost:8732 and rollup should follow the same pattern. otherwise we will have to explicitly pass the endpoint to both the rollup / client / baker when builing their configs
> 3. we added the unused port just for the parallel test to pass and there would be only one isntance of octez for user anyway and they might find it confusing if the octez node is running on differnt port
> 
> will make the baker fallback to the default node endpoin as well

The initial motivation for random ports is to allow tests to run in parallel, particularly useful when using `octez` to set up integration tests.

crates/jstzd/tests/utils.rs
zcabter: Ah, opening a thread will disallow merging the PR while a top level discussion is going on.

huancheng-trili:
but we can just manually assign unused ports as done in [crates/jstzd/tests/utils.rs](https://github.com/jstz-dev/jstz/pull/623/files#diff-f27117d2178931e631e9b5752e21ce1921bd53a411e39e945728281e3b666c19R83)

zcabter:
Yes, we could always set port explicitly but the frequency of setting unused ports will be much higher than a setting the default RPC as the `octez` crate will be used in integration tests. Another small but fair point is that falling back to the default port would clash with the user's running node.

crates/octez/src/async/client.rs
```diff
@@ -25,8 +23,7 @@ pub struct OctezClientBuilder {
     binary_path: Option<PathBuf>,
     // if None, use temp directory
     base_dir: Option<PathBuf>,
-    // if None, use localhost:8732 (the default endpoint for octez-node)
```
zcabter, line 28: I suggest to make this a required field and add `fn with_node_config`

crates/octez/src/async/client.rs
```diff
@@ -25,8 +23,7 @@ pub struct OctezClientBuilder {
...
-    endpoint: Option<Endpoint>,
+    octez_node_endpoint: Option<Endpoint>,
```
zcabter, line 26: I suggest to make this a required field and add fn with_node_config

ryutamago: i think we can have two constructions: 
```
fn new(octez_node_endpoint) -> Self;
fn with_node_config(config: OctezNodeConfig);
```

zcabter:
I am happy with the changes. I will let @huancheng-trili review and approve if he is happy with them too.

huancheng-trili (approved):

huancheng-trili:
Please add more details to the PR description.

PR #624, opened 2024-10-30
fix(octez): allow float in octez rollup node

zcabter:
# Context
Floats are disallowed by default in the wasmer run by the rollup due to an Etherlink constraint. This PR adds a patch to allow that within jstz. 

# Description

# Manually testing the PR


johnyob:
@zcabter I added a fixup commit, would you be able to review?

The commit enables better support for floats in the rollup debugger and fixes a possible issue with nan canonicalisation for wasmer

johnyob (approved):

PR #625, opened 2024-10-30
feat: add `jstz_engine` crate

johnyob:
# Context

This PR adds the skeleton for the much long-awaited `jstz_engine` crate -- the crate that
will provide memory-safe wrappers on SpiderMonkey's `mozjs` crate. 

There is a simple unit test in `jstz_engine` demonstrating that SpiderMonkey does correctly
work in our CI. 

# Description

## Part 1. Adding `mozjs` from crates.io

Initially I tried adding `mozjs` from crates.io, but this backfired since the released version of mozjs is
**5 years old**. 

## Part 2. Building `mozjs` in Nix

So we'll need to use `patch.crates-io` 😔

An initial goal of this PR was to add the patched `mozjs` in such a way that would avoid 
us rebuilding SpiderMonkey each time (~10-15 minutes on a reasonably powerful machine). 

`mozjs-sys`'s build system luckily provides such an approach, where one builds an 'archive' 
containing `libjs_static.a` (statically compiled SpiderMonkey) and some autogenerated Rust files. 

Unfortunately, the compilation of `libjs_static.a` (orchestrated by `mozjs-sys`'s build script) doesn't
work on MacOS (in a Nix environment). The following was the initial attempt:

```nix
  # `mozjs` builds SpiderMonkey automatically. However, this is quite slow. To avoid re-building
  # SpiderMonkey each time, we create a single derivation that builds an archive with all
  # the necessary build artifacts from SpiderMonkey. `mozjs` can then link against these artifacts
  # when we build the crate as part of `cargoDeps`.
  mozjs = let
    src = pkgs.fetchFromGitHub {
      owner = "servo";
      repo = "mozjs";
      # TODO():
      # We should the revision from our Cargo.toml file (since we specify
      # it for `patch.crates-io`)
      rev = "a02aaf1e11fd275f2f129d0c7ca80a9d07460036";
      hash = "sha256-SghYDAnUr2BEuzt/j1NcExqNMxLgCCXttkTnTbreoas=";
    };
    common = {
      pname = "mozjs";
      version = "0.14.1";
      inherit src;
      doCheck = false;
      # mozjs (for some reason?) doesn't use a `Cargo.lock` file
      # so we need to provide one explicitly.
      #
      # Note: We generate this by simply running `cargo build` on the `mozjs` crate
      # (at the given revision). This MUST be updated whenever the `mozjs` crate is
      # updated.
      cargoLock = ./patches/mozjs/Cargo.lock;
    };
  in
    craneLib.cargoBuild (common
      // {
        cargoArtifacts = craneLib.buildDepsOnly common;

        MOZJS_CREATE_ARCHIVE = 1;
        postConfigure =
          ''
            # Build process requires a $HOME directory
            export HOME=$(mktemp -d)
          ''
          + lib.optionalString stdenv.isDarwin ''
            # standalone as(1) doesn't correctly handle `-D` flags for preprocessor `#define`s
            # This ensures we override the environment variables set in `cc-wrapper/setup-hook.sh`
            export AS="$CC -c"
          '';

        buildInputs = lib.optionals stdenv.isDarwin [
          apple-sdk_15
        ];

        nativeBuildInputs = with pkgs;
          [
            python3 # for mozbuild
            gnum4 # for m4
            llvmPackages.llvm # for llvm-objdump
          ]
          ++ lib.optionals stdenv.isDarwin [
            xcbuild # for xcrun
          ];

        installPhase =
          # Building with `MOZJS_CREATE_ARCHIVE=1` creates a tarball at
          # `target/libmozjs-<target>.tar.gz`
          ''
            mkdir $out
            cp target/libmozjs-${stdenv.hostPlatform.rust.rustcTarget}.tar.gz $out/libmozjs.tar.gz
          '';
      });

```

Unfortunately this derivation suffers from two issues (despite building):
1. Linking the `mozjs` archive results in the following error:
    ```
      ld: too many compact unwind infos in anon function '...(RegExp.o)'
      clang: error: linker command failed with exit code 1 (use -v to see invocation)
    ```
    This is due to the use of labels starting with `L` being treated as local by the 
    assembler. When targeting aarch64-darwin (Mach-O), having local labels 
    break the generation of compact (stack) unwinding info. 
  
    This happens because local labels are not seen as function
    boundaries and multiple .cfi_startproc/.cfi_endproc are
    mashed into a single function.
  
    The solution is to force Rust (and clang) to generate dwarf unwind info
    using the clang option: `-femit-dwarf-unwind=always`. Once we have
    the dwarf unwinding infos, we can strip the compact unwind sections using
    `strip --remove-section __LD,__compact_unwind` and link with
    `-Wl,-keep_dwarf_unwind -Wl,-no_compact_unwind`. 

2. The objects in `libjs_static.a` have no symbols, resulting in a missing symbol error when linking.
    *At this point I gave up and back tracked, this still remains a mystery to me* 

Unfortunately, we *will* need to solve this problem in future when targeting `riscv64gc-unknown-linux-musl`, but that is a problem for later I guess 🤷‍♂️

## Part 3. Using external builds for `mozjs`

Instead I decided to look if `mozjs` packages these archives anywhere (since the released `mozjs`)
Trying this manually (using several local builds) worked very nicely!

After looking at the [`mozjs` CI](https://github.com/servo/mozjs/blob/main/.github/workflows/build.yml) I found that these archives are indeed uploaded to GitHub, found with each release [here](https://github.com/servo/mozjs/releases/tag/mozjs-sys-v0.128.3-1). 

Hooray! A reproducible (but note very hermetic) way for including `mozjs` in `jstz_engine`.  

# Manually testing the PR

```sh
cargo nextest --package "jstz_engine"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,6 +309,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.69.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
+dependencies = [
+ "bitflags 2.6.0",
+ "cexpr",
+ "clang-sys",
+ "itertools 0.12.1",
+ "lazy_static",
+ "lazycell",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash 1.1.0",
+ "shlex",
+ "syn 2.0.85",
+ "which",
+]
+
+[[package]]
 name = "bit-set"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -402,7 +423,7 @@ dependencies = [
  "icu_normalizer",
  "indexmap 2.6.0",
  "intrusive-collections",
- "itertools",
+ "itertools 0.13.0",
  "num-bigint 0.4.6",
  "num-integer",
  "num-traits",
@@ -596,12 +617,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ac0150caa2ae65ca5bd83f25c7de183dea78d4d366469f148435e2acfbad0da"
 
 [[package]]
+name = "calendrical_calculations"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cec493b209a1b81fa32312d7ceca1b547d341c7b5f16a3edbf32b1d8b455bbdf"
+dependencies = [
+ "core_maths",
+ "displaydoc",
+]
+
+[[package]]
 name = "cc"
 version = "1.1.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2e7962b54006dcfcc61cb72735f4d89bb97061dd6a7ed882ec6b8ee53714c6f"
 dependencies = [
  "shlex",
+]
+
+[[package]]
+name = "cexpr"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
+dependencies = [
+ "nom",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


zcabter (approved):
LGTM 🚒

PR #627, opened 2024-10-31
feat(octez): add checks for bootstrap accounts in parameters

huancheng-trili:
# Context

Completes JSTZ-149.

# Description

Adds two checks for bootstrap accounts in the parameter set:
* there must be at least one bootstrap account
* at least one bootstrap account needs to have at least 6000 tez

Adding these two specific checks here because the error message from octez client is pretty useless. I think these are pitfalls that users might easily run into and it would definitely help if we catch the errors before the parameter file goes to the octez executables. The issue probably not going to happen using the template parameter file included in jstz because we will put some well funded accounts there, but it can occur when users use their own parameter file as the base.

# Manually testing the PR

* Unit test: added test cases
```sh
$ cargo test --lib async::protocol::tests::merge_bootstrap_accounts
running 2 tests
test r#async::protocol::tests::merge_bootstrap_accounts_no_account ... ok
test r#async::protocol::tests::merge_bootstrap_accounts_insufficient_fund ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 60 filtered out; finished in 0.00s
```
* Integration test: updated the test to show that the second check is valid

### DIFF
--- a/crates/jstzd/tests/protocol_test.rs
+++ b/crates/jstzd/tests/protocol_test.rs
@@ -16,14 +16,26 @@ use utils::{
 #[tokio::test(flavor = "multi_thread")]
 async fn protocol_parameters() {
     let mut param_builder = ProtocolParameterBuilder::new();
-    let bootstrap_account_address = "tz1Ke5At2oVQGyjtU325kUEANe1XGmJBHj2Y";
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
LGTM 🙌

PR #631, opened 2024-11-04
feat(octez): add OctezClientConfig

huancheng-trili:
# Context

Part of JSTZ-163; in preparation for #634.

[JSTZ-163](https://linear.app/tezos/issue/JSTZ-163/spawn-octezbaker-in-jstzd)

# Description

Make `OctezClientBuilder` build `OctezClientConfig` instead of `OctezClient` directly. `OctezClient` then can be instantiated by calling `new(config)`. `OctezClientConfig` is needed because configs for octez client will be added to `JstzdConfig` and I don't think the client itself should be part of the config.

# Manually testing the PR

Updated relevant test cases.


ryutamago (approved):
💪

PR #634, opened 2024-11-04
feat(jstzd): spawn baker in jstzd

huancheng-trili:
# Context

Completes JSTZ-163.

[JSTZ-163](https://linear.app/tezos/issue/JSTZ-163/spawn-octezbaker-in-jstzd)

# Description

Spawn baker in jstzd.

Note that there is a small usability issue regarding bootstrap accounts: these accounts cannot be used by the baker except for the hard-coded activator account. This is because bakers only use accounts with secret keys imported and bootstrap accounts in parameter files only have public keys. This is especially a problem when users use their own template parameter files with bootstrap accounts attached. There is no solution for now. I don't think hardcoding a "base baker" account inside `ProtocolParameterBuilder` is a good idea since it adds irrelevant logic to that module.

# Manually testing the PR

Updated `crates/jstzd/tests/jstzd_test.rs`.

```sh
$ cargo test --test jstzd_test
running 1 test
< ... baker logs >
test jstzd_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.52s
```

### DIFF
--- a/crates/jstzd/src/task/jstzd.rs
+++ b/crates/jstzd/src/task/jstzd.rs
@@ -1,23 +1,42 @@
-use super::{octez_node::OctezNode, Task};
+use super::{octez_baker::OctezBaker, octez_node::OctezNode, utils::retry, Task};
 use anyhow::Result;
 use async_dropper_simple::AsyncDrop;
 use async_trait::async_trait;
 use axum::Router;
-use octez::r#async::node_config::OctezNodeConfig;
+use octez::r#async::{
+    baker::OctezBakerConfig,
+    client::{OctezClient, OctezClientConfig},
+    node_config::OctezNodeConfig,
+    protocol::ProtocolParameter,
+};
 use tokio::{net::TcpListener, task::JoinHandle};
 
 struct Jstzd {
     octez_node: OctezNode,
+    baker: OctezBaker,
 }
 
 #[derive(Clone)]
 pub struct JstzdConfig {
     octez_node_config: OctezNodeConfig,
+    baker_config: OctezBakerConfig,
+    octez_client_config: OctezClientConfig,
+    protocol_params: ProtocolParameter,
 }
 
 impl JstzdConfig {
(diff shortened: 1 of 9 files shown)


crates/jstzd/src/task/jstzd.rs
```diff
@@ -26,17 +45,101 @@ impl Task for Jstzd {
...
+    }
+}
+
+impl Jstzd {
+    async fn import_activator(octez_client: &OctezClient) {
+        let activator = "activator".to_string();
+        octez_client
+            .import_secret_key(
+                &activator,
+                "unencrypted:edsk31vznjHSSpGExDMHYASz45VZqXN4DPxvsa4hAyY8dHM28cZzp6",
```
ryutamago, line 107: nit:
can store the alias and secret key in const for now?

crates/jstzd/src/task/jstzd.rs
```diff
@@ -26,17 +45,101 @@ impl Task for Jstzd {
...
+            )
+            .await
+            .expect("Failed to import account 'activator'");
+    }
+
+    async fn activate_protocol(
+        octez_client: &OctezClient,
+        protocol_params: &ProtocolParameter,
+    ) -> anyhow::Result<()> {
```
ryutamago, line 116: nit: remove anyhow

crates/jstzd/src/task/jstzd.rs
```diff
@@ -26,17 +45,101 @@ impl Task for Jstzd {
...
+    }
+
+    async fn wait_for_node(octez_node: &OctezNode) -> Result<()> {
+        let ready = retry(10, 1000, || async {
+            if let Ok(v) = octez_node.health_check().await {
+                if v {
+                    return Ok(true);
+                }
+            }
+            Err(anyhow::anyhow!(""))
```
ryutamago, line 134: we could return Ok(false) instead of error w empty string?

ryutamago (approved):
LGTM, left some comments for small nits 💪

PR #635, opened 2024-11-04
Jstzd: rollup node

ryutamago:
# Context

# Description

Add rollup node.
For the temp data dir in the ConfigBuiilder, users will have to specify the path to the preimage direcotries as otherwise the kernel installer won't be able to retreive the binaries
```
#[derive(Clone, PartialEq, Debug)]
pub enum RollupDataDir {
    /// Path to the rollup data directory. This directory
    /// should contain the kernel pre image files under `wasm_2_0_0/`
    Path { data_dir: PathBuf },
    /// Path to the directory containing the kernel pre image files
    /// This will be copied to `wasm_2_0_0/` in the temp directory that will be created
    Temp { preimages_dir: PathBuf },
}

pub struct OctezRollupConfigBuilder {
    ...
    /// Path to the smart rollup data directory
    data_dir: RollupDataDir,
   ...
}
```

# Manually testing the PR

added both unit test & integration test


crates/octez/src/async/rollup.rs
```diff
@@ -0,0 +1,228 @@
...
+            data_dir: data_dir.to_path_buf(),
+            octez_client_base_dir: octez_client_base_dir.to_path_buf(),
+            octez_node_endpoint: octez_node_endpoint.clone(),
+            rpc_endpoint: rpc_endpoint.clone(),
+        }
+    }
+}
+
+impl OctezRollup {
+    fn command(&self) -> Command {
```
huancheng-trili, line 138: Why not merge this into `run`?

crates/octez/src/async/rollup.rs
```diff
@@ -0,0 +1,228 @@
...
+            "with",
+            "operators",
+            operator,
+            "--data-dir",
+            &self.data_dir.to_string_lossy(),
+            "--rpc-addr",
+            self.rpc_endpoint.host(),
+            "--rpc-port",
+            &self.rpc_endpoint.port().to_string(),
+            "--boot-sector-file",
```
huancheng-trili, line 170: I think `--boot-sector-file` should be optional since it's necessary only for bootstrap rollups

crates/jstzd/src/task/octez_rollup.rs
```diff
@@ -0,0 +1,114 @@
...
+impl Task for OctezRollup {
+    type Config = OctezRollupConfig;
+
+    async fn spawn(config: Self::Config) -> Result<Self> {
+        let data_dir = match &config.data_dir {
+            RollupDataDir::Temp {
+                preimages_dir: from,
+            } => {
+                let temp_dir = Directory::default();
+                let to = PathBuf::from(&temp_dir).join("wasm_2_0_0");
```
huancheng-trili, line 41: `wasm_2_0_0` should not be hardcoded. This depends on the type of the rollup to be executed by the node.

crates/jstzd/tests/utils.rs
```diff
@@ -39,7 +47,34 @@ pub async fn setup() -> (OctezNode, OctezClient, octez_baker::OctezBaker) {
     activate_alpha(&octez_client).await;
 
     let baker = spawn_baker(&octez_node, &octez_client).await;
-    (octez_node, octez_client, baker)
+    let rollup = spawn_rollup(&octez_node, &octez_client).await;
```
huancheng-trili, line 50: Adding this here means that all existing test cases calling this method are going to spin up a rollup node. I don't think this is what we want. Better to call `spawn_rollup` in `crates/jstzd/tests/octez_rollup_test.rs` instead.

crates/octez/src/async/rollup.rs
ryutamago: in our usecase we are only using bootstrap jstz rollup IIUC so i think it's better to make it required for simplicity. 
cc @zcabter

crates/octez/src/async/rollup.rs
ryutamago: though it would be more readable if we separate the commands with the global options from the specific command like run. Also, we can reuse if we were to implement other commands like the config init.

crates/octez/src/async/directory.rs
```diff
@@ -37,12 +37,11 @@ impl TryFrom<PathBuf> for Directory {
     }
 }
 
-impl TryFrom<&Directory> for PathBuf {
```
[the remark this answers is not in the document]
ryutamago: done
https://github.com/jstz-dev/jstz/pull/636

crates/octez/src/async/rollup.rs
```diff
@@ -0,0 +1,228 @@
...
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+use tezos_crypto_rs::hash::SmartRollupHash;
+use tokio::process::{Child, Command};
+
+const DEFAULT_BINARY_PATH: &str = "octez-smart-rollup-node";
+
+#[derive(Clone, PartialEq, Debug)]
+pub enum RollupDataDir {
```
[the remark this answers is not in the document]
ryutamago: i thought about that too. My rational was for the rollup to follow the same api with other components where the `temp dir` is supported (e.g. the node and client), and it wouldn't make sense for the rollup to support `temp dir` without specifying the path to the pre images. i agree this is a bit awkward but it can also just be a convenient api on top of the rollup.

so we either 
1. support temp dir that requires preimages 
2. do not support temp dir at all. 


cc @zcabter

crates/jstzd/tests/utils.rs
ryutamago: done

crates/octez/src/async/rollup.rs
ryutamago: actually this is in octez so we can keep it optional, fixed

huancheng-trili (approved):
Let's see how it goes

PR #637, opened 2024-11-05
feat(jstzd): add bootstrap contracts exchanger and jstz_native_bridge

huancheng-trili:
# Context

Closes JSTZ-166.

[JSTZ-166](https://linear.app/tezos/issue/JSTZ-166/deploy-xtz-bridge)

# Description

Adds contracts `exchanger` and `jstz_native_bridge` to the sandbox as bootstrap contracts.

# Manually testing the PR

Integration test:
* Updated `jstzd_test`
* Added `contract_test` to ensure that the contract json files always match the michelson code in `contracts/`

### DIFF
--- a/crates/jstzd/src/lib.rs
+++ b/crates/jstzd/src/lib.rs
@@ -1,6 +1,10 @@
 pub mod docker;
 pub mod task;
 
+pub const EXCHANGER_ADDRESS: &str = "KT1F3MuqvT9Yz57TgCS3EkDcKNZe9HpiavUJ";
+pub const JSTZ_ROLLUP_ADDRESS: &str = "sr1PuFMgaRUN12rKQ3J2ae5psNtwCxPNmGNK";
+pub const JSTZ_NATIVE_BRIDGE_ADDRESS: &str = "KT1GFiPkkTjd14oHe6MrBPiRh5djzRkVWcni";
+
 /// The `main` function for running jstzd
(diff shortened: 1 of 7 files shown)


crates/jstzd/src/lib.rs
```diff
@@ -1,8 +1,67 @@
...
+pub const JSTZ_NATIVE_BRIDGE_ADDRESS: &str = "KT1GFiPkkTjd14oHe6MrBPiRh5djzRkVWcni";
 
 /// The `main` function for running jstzd
 pub async fn main() -> anyhow::Result<()> {
     println!("Hello, world!");
     Ok(())
 }
+
+// this requires a running octez node with any activated protocol version
+pub async fn generate_bootstrap_contract_files(
```
ryutamago, line 18: i'm thinking of the use cases here and was wondering if we have to expose a function that generates the contract files. we probably need something like `generate_l1_params_files` that will build the rollup installer kernel, preimages and the contracts but do we ever need the user to generate the contract files like this?

huancheng-trili: We don't need to expose this. I placed it in `lib.rs` simply because it contains some lines related to jstzd, specifically the contract init data part, and thus I don't think putting it in `contract_test` is that appropriate. It's `pub` just because it needs to be imported by the test.

I'll move it to `contract_test` to avoid exposing this unnecessarily.

ryutamago (approved):
🚀

PR #638, opened 2024-11-06
refactor(octez): remove async from node

ryutamago:
# Description
remove async from octez node


huancheng-trili:
I can't recall why I put it like that, but for `config_init` and `generate_identity` specifically, we should wait for the task to complete in these methods and keep them async.

ryutamago:
@huancheng-trili 
Just removed async for run

huancheng-trili (approved):
You can create a PR to update those two methods if you want. I'll do that later otherwise.

PR #639, opened 2024-11-06
feat(jstzd): call contract

ryutamago:
# Context
call contract:
[leounoki-jstz-139](https://linear.app/tezos/issue/JSTZ-139/implement-octezclientcall-contract)
transfer:
[leounoki-jstz-138](https://linear.app/tezos/issue/JSTZ-138/implement-octezclienttransfer)

# Description

This PR implements the private method `transfer` and a public method `call_contract` on top of the transfer.

# Manually testing the PR

Unit tests for transfer options builder and integration test for call_contract

### DIFF
--- a/crates/octez/src/async/client.rs
+++ b/crates/octez/src/async/client.rs
@@ -159,6 +159,74 @@ impl TryFrom<StdOut> for Address {
     }
 }
 
+#[derive(Default)]
+struct TransferOptionsBuilder {
+    from: Option<String>,
+    to: Option<String>,
+    amount: Option<f64>,
+    entrypoint: Option<String>,
+    arg: Option<String>,
+    burn_cap: Option<f64>,
+}
+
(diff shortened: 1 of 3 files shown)


crates/octez/src/async/client.rs
```diff
@@ -159,6 +159,74 @@ impl TryFrom<StdOut> for Address {
...
+            arg: self.arg,
+            burn_cap: self.burn_cap,
+        })
+    }
+}
+
+struct TransferOptions {
+    from: String,
+    to: String,
+    amount: u64,
```
huancheng-trili, line 224: This needs to be `f64` since `amount` is in tez and fractions are accepted.

ryutamago: good point, thanks!

huancheng-trili (approved):

PR #640, opened 2024-11-07
fix(octez): set default network for run options

huancheng-trili:
# Context

Fix the default value for `OctezNodeRunConfig::network`.

# Description

`OctezNodeConfigBuilder::build` uses the default value for `OctezNodeRunOptions` when `run_config` is not set, but this default value is not the one expected because `OctezNodeRunOptions::network` is a string and defaults to `""` instead of `"sandbox"`. This is fixed by implementing `Default` for `OctezNodeRunConfig` explicitly.

# Manually testing the PR

Added one test case.

### DIFF
--- a/crates/octez/src/async/node_config.rs
+++ b/crates/octez/src/async/node_config.rs
@@ -30,7 +30,7 @@ impl Display for OctezNodeHistoryMode {
     }
 }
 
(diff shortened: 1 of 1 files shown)


zcabter (approved):
lgm

PR #641, opened 2024-11-07
Jstz-157: implement openapi for accounts service

zcabter:
# Context
Implements OpenAPI for `AccountsService`

# Description
Adds `utoipa::path` annotation to `AccountsService` handlers.

Generate updated openapi schema

# Manually testing the PR
This PR does not make changes to the logic. OpenAPI schema can be checked with
```
cargo test -p jstz_node api_doc_regression
```

### DIFF
--- a/crates/jstz_node/openapi.json
+++ b/crates/jstz_node/openapi.json
@@ -14,6 +14,201 @@
     "version": "0.1.0-alpha.0"
   },
   "paths": {
+    "/accounts/{address}/balance": {
+      "get": {
+        "tags": [
+          "Accounts"
+        ],
+        "summary": "Get balance of an account",
+        "operationId": "get_balance",
+        "parameters": [
+          {
+            "name": "address",
+            "in": "path",
+            "required": true,
+            "schema": {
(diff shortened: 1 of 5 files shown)


crates/jstz_node/openapi.json
```diff
@@ -14,6 +14,201 @@
...
+        }
+      }
+    },
+    "/accounts/{address}/kv": {
+      "get": {
+        "tags": [
+          "Accounts"
+        ],
+        "summary": "Get KV value under a given key path",
+        "description": "Get KV value under a given key path for an account. If `key` is not provided,\nthe empty key path will be used.",
```
huancheng-trili, line 102: It seems that multiline comments are not concatenated and we have a `\n` here. Not sure if there's any config for this.

huancheng-trili (approved):

crates/jstz_node/openapi.json
zcabter: Yeah not sure we could get a round that but it generates ok on Scalar and, hopefully, client libs.

PR #643, opened 2024-11-08
feat(jstz_node): implement openapi for logs service

zcabter:
# Context
Implements OpenAPI for LogsService

# Description
 We probably want a single client gen binding that supports all node endpoints. With that in mind, the persistent log endpoints have been removed from conditional compilation and instead return a Bad Request if it is called on a node that was not compiled with persistent logging.

(Note: We should change enabling/disabling persistent logging enable from conditional compilation to runtime configuration)

Removed the `QueryResponse` which is a wrapper over LogRecord.

Removed `QueryParams` which acts as a function dispatch over the db api. Instead, just called the db api directly because they are 1 to 1 with their handlers.

# Manually testing the PR

```
cargo test -p jstz_node api_doc_regression
cargo test --features persisitent-logging -p jstz_node api_doc_regression
```

### DIFF
--- a/crates/jstz_node/src/services/logs/db.rs
+++ b/crates/jstz_node/src/services/logs/db.rs
@@ -1,8 +1,10 @@
 #![cfg(feature = "persistent-logging")]
 use std::fs;
 
-use super::{Line, QueryResponse};
+use super::Line;
 use anyhow::{anyhow, Result};
+use jstz_api::js_log::LogLevel;
+use jstz_crypto::public_key_hash::PublicKeyHash;
 use jstz_proto::{
     context::account::Address, js_logger::LogRecord, request_logger::RequestEvent,
 };
@@ -13,7 +15,7 @@ use tokio::task::spawn_blocking;
 
 pub type SqliteConnectionPool = Pool<SqliteConnectionManager>;
 pub type SqliteConnection = PooledConnection<r2d2_sqlite::SqliteConnectionManager>;
-type QueryResponseResult = Result<Vec<QueryResponse>>;
+type QueryResponseResult = Result<Vec<LogRecord>>;
 
 const DB_PATH: &str = ".jstz/log.db";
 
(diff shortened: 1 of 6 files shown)


crates/jstz_node/src/services/logs/db.rs
```diff
@@ -1,8 +1,10 @@
-#![cfg(feature = "persistent-logging")]
+// #![cfg(feature = "persistent-logging")]
```
huancheng-trili, line 1: Do we need to keep this line?

zcabter: My bad, I have uncommented the line.

crates/jstz_node/src/services/logs/db.rs
```diff
@@ -130,17 +132,30 @@ impl Db {
...
+        let query_result = stmt
             .query_map(params, |row| {
-                Ok(QueryResponse::Log {
-                    level: row.get(1)?,
-                    content: row.get(2)?,
-                    function_address: row.get(3)?,
-                    request_id: row.get(4)?,
-                })
+                Ok((
+                    row.get::<usize, String>(0)?,
```
huancheng-trili, line 138: Why are the indices changed to `0, 1, 2, 3`? Did the log table's schema change?

crates/jstz_node/src/services/logs/mod.rs
```diff
@@ -258,6 +204,26 @@ impl LogsService {
     }
 }
 
+#[derive(Deserialize, Debug)]
+#[cfg_attr(not(feature = "persistent-logging"), allow(dead_code))]
+pub struct Pagination {
```
ryutamago, line 209: nit: 
could implement default here given that it's only used for persistent-logging?

zcabter: Done

crates/jstz_node/src/services/logs/db.rs
zcabter: My bad, fixed

zcabter:
Will need to squash fix up before commit

huancheng-trili (approved):

PR #645, opened 2024-11-11
feat(octez): implement execute_rollup_outbox_message

huancheng-trili:
# Context

Completes JSTZ-145.

[JSTZ-145](https://linear.app/tezos/issue/JSTZ-145/implement-octezclientexecute-rollup-outbox-message)

# Description

Implement `execute_rollup_outbox_message` that performs `octez-client execute outbox message of smart rollup ...`.

# Manually testing the PR

* Integration test: added one test. The setup is as follows:
  * contract: takes input `int(x)` and does something.
  * rollup: writes a message `int(1)` targeting the contract to the outbox whenever an external message arrives at the inbox. Source file is included in this PR as well. Since there is no one else writing messages to the inbox and there is no other rollup running, it's safe to assume that the only outbox message is produced by this exact rollup based on the inbox messages sent by the only call to `octez-client write inbox message` in the test.

  The addresses and contract files are brand new and separated from existing parameter files in order to prevent unexpected changes to the contract files that affect the tests. Housekeeping will be done in a separate PR to organise other test files and keep the base test parameter template clean, i.e. bootstrap contracts and rollups should be injected with `ParameterBuilder` rather than hardcoded in the template.


crates/jstzd/tests/octez_client_test.rs
```diff
@@ -351,3 +358,185 @@ async fn send_rollup_inbox_message() {
...
+        rollup_address,
+        contract_address,
+        &installer_path,
+        &contract_path,
+    )
+    .await;
+
+    let (mut octez_node, octez_client, mut baker) =
+        setup(Some(params.parameter_file().path().to_path_buf())).await;
+    let mut rollup = spawn_rollup(
```
zcabter, line 391: Nit: Would be nice if the builder could take an octez node and client to instantiate it. Not required for this PR.

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -351,3 +358,185 @@ async fn send_rollup_inbox_message() {
...
+        .await
+        .unwrap()
+        .write_all(content.as_bytes())
+        .await
+        .unwrap();
+
+    params
+}
+
+async fn wait_for_outbox_proof(rollup_rpc_endpoint: &str) -> anyhow::Result<OutputProof> {
```
zcabter, line 507: Note: we should turn this into an `OctezRollupClient` function

crates/jstzd/tests/utils.rs
```diff
@@ -17,13 +17,33 @@ pub const ACTIVATOR_SECRET_KEY: &str =
     "unencrypted:edsk31vznjHSSpGExDMHYASz45VZqXN4DPxvsa4hAyY8dHM28cZzp6";
 pub const ROLLUP_ADDRESS: &str = "sr1PuFMgaRUN12rKQ3J2ae5psNtwCxPNmGNK";
 
-pub async fn setup() -> (OctezNode, OctezClient, octez_baker::OctezBaker) {
+pub async fn poll<'a, F, T>(
```
zcabter, line 20: Nit: `retry` is effectively `poll` + predicate check. Would it be possible to implement retry using poll? `poll` might need some changes.

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -351,3 +358,185 @@ async fn send_rollup_inbox_message() {
...
+    })
+    .await
+    .expect("should be able to find outbox message soon enough");
+
+    let url = format!("{rollup_rpc_endpoint}/global/block/head/helpers/proofs/outbox/{outbox_level}/messages?index={message_index}");
+    if let Ok(res) = reqwest::get(&url).await {
+        if let Ok(v) = res.json::<OutputProof>().await {
+            return Ok(v);
+        }
+    }
```
zcabter, line 540:
```suggestion
    let res = reqwest::get(&url)?;
     let v = res.json::<OutputProof>().await?
    }
```

crates/jstzd/tests/octez_client_test.rs
```diff
@@ -351,3 +358,185 @@ async fn send_rollup_inbox_message() {
...
+    Err(anyhow::anyhow!("cannot find outbox proof"))
```
zcabter, line 541:
```suggestion
    let res = reqwest::get(&url)?;
    res.json::<OutputProof>().await
```
```

File: crates/jstzd/tests/octez_client_test.rs [hunk 6]
```diff
@@ -351,3 +358,185 @@ async fn send_rollup_inbox_message() {
...
+        if let Ok(res) = reqwest::get(&url).await {
+            let vs = res.json::<Vec<Executable>>().await.unwrap();
+            // using the first message here since any of those should work
+            if let Some(v) = vs.first() {
+                if let Some(m) = v.messages.first() {
+                    return Some((v.outbox_level, m.message_index));
+                }
+            }
+        }
+        None
```
zcabter, line 530: Perhaps try to avoid nested `if let` in favour of `?`

zcabter:
Logic of the actual function looks fine.

crates/jstzd/tests/octez_client_test.rs
huancheng-trili: You mean `OctezRollupConfigBuilder`? If so, it was like that at some point but I changed it because it introduced dependencies on the node and the client instances and I ran into some issues because of that. Perhaps we can revisit that later on.

crates/jstzd/tests/octez_client_test.rs
huancheng-trili: For now I think it's better to keep this function here because there are too many assumptions based on the test setup in this implementation. We can polish and move it when we need it.

crates/jstzd/tests/utils.rs
huancheng-trili: Yes. I'm keeping it here as is and moving it properly in #681 since I'm adding some tests for it and thus there are quite a few lines of changes.

crates/jstzd/tests/octez_client_test.rs
huancheng-trili: Updated

crates/jstzd/tests/octez_client_test.rs
huancheng-trili: Updated

crates/jstzd/tests/octez_client_test.rs
zcabter: Yes the `OctezRollupConfigBuilder`, because that is kind of what `spawn_rollup` is doing.

crates/jstzd/tests/octez_client_test.rs
zcabter: Can you add a ticket to track it?

huancheng-trili: https://linear.app/tezos/issue/JSTZ-206/polish-wait-for-outbox-proof-for-general-use

zcabter (approved):

PR #648, opened 2024-11-11
JSTZD: implement Task for jstz node

ryutamago:
# Context

[JSTZ-190: implement task for `JstzNode`](https://linear.app/tezos/issue/JSTZ-190/implement-task-for-jstznode)
Next step:
[spwan rollup node](https://linear.app/tezos/issue/JSTZ-164/spawn-octezrollupnode-in-jstzd)
[JSTZ-165: Spawn `JstzNode` in Jstzd](https://linear.app/tezos/issue/JSTZ-165/spawn-jstznode-in-jstzd)

# Description  

* add health check endpoint to jstz node `/ping`
* add jstz_node config
* implement task for jstz_node

# Manually testing the PR

* add integration test

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -1,12 +1,14 @@
-use std::{path::PathBuf, sync::Arc};
-
+use anyhow::Result;
 use api_doc::ApiDoc;
+use axum::{http, routing::get};
+use config::JstzNodeConfig;
 use octez::OctezRollupClient;
 use services::{
     accounts::AccountsService,
     logs::{broadcaster::Broadcaster, db::Db, LogsService},
     operations::OperationsService,
 };
+use std::{path::PathBuf, sync::Arc};
 use tokio::net::TcpListener;
 use tower_http::cors::{Any, CorsLayer};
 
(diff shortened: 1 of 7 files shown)


crates/jstz_node/src/lib.rs
```diff
@@ -66,9 +83,14 @@ fn router() -> OpenApiRouter<AppState> {
         .merge(OperationsService::router_with_openapi())
         .merge(AccountsService::router_with_openapi())
         .merge(LogsService::router_with_openapi())
+        .route("/ping", get(health_check))
```
huancheng-trili, line 86: Maybe rename this route to something more obvious for health checks, e.g. `/health`?

And since this handler doesn't really do anything, I think something like this is sufficient:
```rs
.route("/health", get(http::StatusCode::OK))
```

crates/jstzd/src/task/jstz_node.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    async fn kill(&mut self) -> Result<()> {
+        self.handle.abort();
+        Ok(())
+    }
+
+    async fn health_check(&self) -> Result<bool> {
+        let res = reqwest::get(format!("{}/ping", self.config.endpoint)).await;
+        if res.is_ok_and(|res| res.status().is_success()) {
```
huancheng-trili, line 29: What about replacing this `if` with
```rs
Ok(res.is_ok_and(|res| res.status().is_success()))
```
?

Though I think `Ok(res.is_ok())` should be sufficient since 4xx and 5xx status code leads to `Err`.

crates/jstz_node/src/lib.rs
ryutamago: done!

huancheng-trili (approved):

PR #649, opened 2024-11-11
feat(jstzd): implement health check endpoint in jstzd server

huancheng-trili:
# Context

Part of JSTZ-121.

# Description

Added route `/health` to `JstzdServer`. Note that in this PR the octez node and baker instances in Jstzd are wrapped with `Arc<RwLock<_>>` because they need to be passed to the server as the state. I know we can just make baker derive `Clone` and then clone `JstzdServer.jstzd` and things under the hood are also wrapped with `Arc`, but I think it's clearer like this with explicit `Arc` and `RwLock` in this module and we know for sure that these instances are not duplicated but shared when they are cloned.

# Manually testing the PR

* Integration test: updated `crates/jstzd/tests/jstzd_test.rs` to use the health check endpoint


ryutamago (approved):
neat 💪

PR #650, opened 2024-11-11
fix(jstzd): mitigate test flakiness

huancheng-trili:
# Context

[CI](https://github.com/jstz-dev/jstz/actions/runs/11781846583/job/32815571186) just randomly failed (in some PR checks as well.) This happens a bit too often. For some reason,
* the baker process was not removed, i.e. still listed in `ps` output, after 10 seconds
* calling `octez-client wait for` took longer than expected when it should have returned almost immediately when the operation was present
* etc

and tests failed because of these. This never happened to me when I ran tests locally though.

# Description

Tweaked some test settings so that the tests are less likely to be flakey:
* Wait a bit longer in integration tests so that tests don't fail due to possibly some delay in the runner machine. That octez-client test might be longer, but it's still better than tests randomly failing.
* Give the `activator` account a bit more tez in `jstzd_test` so that it does not run out of fund after baking a few blocks.

# Manually testing the PR

All tests should still pass.


zcabter (approved):
lgtm

PR #653, opened 2024-11-13
feat(octez): implement serialize for endpoint

huancheng-trili:
# Context

Part of JSTZ-121; in preparation for #655.

# Description

Implement `Serialize` for `Endpoint` so that it can be dumped in JSON.

# Manually testing the PR

Added one unit test.

### DIFF
--- a/crates/octez/src/async/endpoint.rs
+++ b/crates/octez/src/async/endpoint.rs
@@ -1,7 +1,6 @@
-#![allow(dead_code)]
-use std::fmt::{self, Display};
-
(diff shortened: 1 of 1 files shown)


ryutamago (approved):
LGTM!
IIRC i think we were able to use the following syntax. i like it as we can see that Endpint implemnts Seriazlie in one glance but it's pretty much the same in terms of LoC so it doesn't matter much :) 
```rs
#[serde(serialize_with = "serialize_as_string")]
struct Endpoiont {
...
}
fn serialize_as_string ... 
```

PR #655, opened 2024-11-13
feat(jstzd): serve octez node config in jstzd server

huancheng-trili:
# Context

Part of JSTZ-121.

# Description

Serve octez node config in Jstzd server. The following endpoints are implemented:
* `GET /config/octez-node`: returns octez node config in JSON.
* `GET /config/`: returns `{"octez-node": <node config>}`.

# Manually testing the PR

* Integration test: updated jstzd_test to check if the responses are as expected.


crates/jstzd/tests/jstzd_test.rs
```diff
@@ -22,13 +23,41 @@ const CONTRACT_NAMES: [(&str, &str); 2] = [
 #[tokio::test(flavor = "multi_thread")]
 async fn jstzd_test() {
     let rpc_endpoint = Endpoint::localhost(unused_port());
+    let jstzd_port = unused_port();
+    let (mut jstzd, config) = create_jstzd_server(&rpc_endpoint, jstzd_port).await;
+
+    jstzd.run().await.unwrap();
+    ensure_jstzd_components_are_up(&rpc_endpoint, jstzd_port).await;
```
zcabter, line 30: Isn't making a request to jstzd health check endpoint enough since it checks the health of other components?

huancheng-trili: Yes, but here I want to make sure that the jstzd health check endpoint is indeed covering all components as expected since there are more components involved now. Of course this test does not cover all possible cases, e.g. when any of the components is down, this health check endpoint should return false, due to the difficulty of setting up such cases, but at least for the cases when everything is up and everything is down, we can ensure that it's working.

zcabter: Ah I see. I was a bit confused because `ensure_jstzd_components_are_up` polls the node and baker after running health succeeds. I think that it should not need to poll node and baker health but jus call it once to assert that those components are up.

zcabter:
GTM overall, just small comment

crates/jstzd/tests/jstzd_test.rs
```diff
@@ -22,13 +23,41 @@ const CONTRACT_NAMES: [(&str, &str); 2] = [
...
+    reqwest::Client::new()
+        .put(&format!("http://localhost:{}/shutdown", jstzd_port))
+        .send()
+        .await
+        .unwrap();
+
+    ensure_jstzd_components_are_down(&rpc_endpoint, jstzd_port).await;
+
+    // stop should be idempotent and thus should be okay after jstzd is already stopped
+    jstzd.stop().await.unwrap();
```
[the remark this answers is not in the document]
huancheng-trili: Yes. I want to test `.stop()` but at the same time I don't feel like running a separate test just for that. I've removed this line since it does not do much here.

zcabter (approved):

PR #656, opened 2024-11-13
feat(jstzd): use AsyncDropper in JstzdServer

huancheng-trili:
# Context

Closes JSTZ-121.

# Description

Wrap the test in a future so that we can still call `stop` at the end of the test to clean up everything before the program exits should the test fail and panic. Otherwise, the test program does not actually close after the panic and the baker and octez-node  processes remain running even after the test program terminates.

In a tokio task, `panic` is captured and returned as `Err`, so handling failed assertions like this is much simpler than trying to do something like `catch_unwind`.

# Manually testing the PR

The test should still work as is.

### DIFF
--- a/crates/jstzd/src/task/jstzd.rs
+++ b/crates/jstzd/src/task/jstzd.rs
@@ -1,6 +1,6 @@
 use super::{octez_baker::OctezBaker, octez_node::OctezNode, utils::retry, Task};
 use anyhow::Result;
-use async_dropper_simple::AsyncDrop;
+use async_dropper_simple::{AsyncDrop, AsyncDropper};
 use async_trait::async_trait;
 use axum::{
     extract::{Path, State},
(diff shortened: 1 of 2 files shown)


crates/jstzd/src/task/jstzd.rs
zcabter, line 1: Isn't the problem because `JstzServer`, although it implements `AsyncDrop`, isn't instantiated with `AsyncDropper<JstzServer>`?

zcabter (changes requested):
Isn't the problem because `JstzServer`, although it implements `AsyncDrop`, isn't instantiated with `AsyncDropper<JstzServer>`?

huancheng-trili:
> Isn't the problem because JstzServer, although it implements AsyncDrop, isn't instantiated with AsyncDropper<JstzServer>?

Yes. I thought implementing `AsyncDrop` would be sufficient and forgot about `AsyncDropper`. I've changed this PR and added `AsyncDropper` to `JstzServer`. The implementation might look a bit awkward because of the locks. Ideally `AsyncDrop` should be implemented for `ServerState`, but there will be some borrow_mut issues without having the lock. The only way to get the lock involved is to wrap everything with a struct and implement `AsyncDrop` for that struct instead. Now `AsyncDropper` is embedded in `JstzServer`. At least with this we don't need to write `AsyncDropper<JstzServer>` everywhere.

> Since AsyncDropper requires impl Default, ServerState.jstzd_config field might have to wrapped in Option<> which is rather inconvenient :(

That's true, but it's not that inconvenient since the server does not really access `jstzd_config` that much. I think it's okay to just wrap it.

> If there is a jstzd panics for some reason, we definitely want the octez resources to be cleaned up. Is there a better way to to hook self.stop()when the server starts up?

`AsyncDrop` should be able to handle what we can handle for now. Based on my observation, `AsyncDrop` for `JstzdServer` does clean up octez resources in the following cases:
* when something in the program outside of `JstzdServer` panics and causes the program to terminate
* when something inside `JstzdServer` itself, e.g. some route handler, panics
* when something inside `Jstzd` itself, e.g. something bad in any of its methods, panics

At this moment, there aren't really that many places in `Jstzd` where it can panic. It has only a few methods and `AsyncDrop` for `JstzdServer` can handle failures at those places. For `spawn`, after calling `spawn`, the child processes continue on their owns and `Jstzd` does not get affected should those processes fail somehow. I think there was some discussion before regarding making jstzd "recover" from subprocess failures and we can look into this later on, but for now `Jstzd` won't be aware of those events.

zcabter (approved):
LGTM thanks

PR #657, opened 2024-11-13
JSTZD: spawn rollup in jstzd

ryutamago:
# Context

[Spawn rollup in jstzd](https://linear.app/tezos/issue/JSTZ-164/spawn-octezrollupnode-in-jstzd)
# Description

* jstzd now spawns rollup
* add the jstz preimages and kernel installer, parameters_ty under `test/jstz_rollup`
* add util functions to wait for the block to reach certain level. Needed because we can't start the rollup until the block level 3 when the alpha protocol is enabled 

NOTE: it could be helpful to have a script that will generate the jstzd binaries in the future but for now we can just manually add them.

# Manually testing the PR

integration test:
```
cargo test --package jstzd --test jstzd_test       
```


ryutamago:
closing to split

PR #658, opened 2024-11-14
feat(jstz_engine): add skeletons for `Context` and `Compartment`

johnyob:
# Context

A SpiderMonkey `Runtime` contains the state of the garbage collector. The collector partitions the JavaScript heap into compartments. `Compartment`s have the invariant that no object in the compartment can contain pointers to another compartment. 

A `Context` contains the state for a virtual machine that executes and manages JavaScript objects within a `Runtime`. It can compile and execute scripts, get and set object properties, call JavaScript functions, convert JavaScript data from one type to another, create objects, and so on. 

# Description

This PR adds the skeletons for the `Compartment` and `Context` types & traits. 

We track the compartment that an object or context is in at the type-level using capability traits. 

# Manually testing the PR

```
```

### DIFF
--- a/crates/jstz_engine/src/compartment.rs
+++ b/crates/jstz_engine/src/compartment.rs
@@ -1,0 +1,26 @@
+//! SpiderMonkey partitions the heap into compartments. Some key properties of
+//! compartments are:
+//!  - Every GC object (cell) belongs to one compartment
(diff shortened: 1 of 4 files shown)


zcabter (approved):
LGTM 🚀

PR #660, opened 2024-11-14
feat(jstz_engine): add safe wrapper for JS `Script`s

johnyob:
# Context

Adds memory and type-safe version of `JSScript`.

# Description

Adds ability to compile and evaluate scripts. 

# Manually testing the PR

```
```

### DIFF
--- a/crates/jstz_engine/src/script.rs
+++ b/crates/jstz_engine/src/script.rs
@@ -1,0 +1,137 @@
+//! This module provides an interface for JavaScript scripts in SpiderMonkey.
+//! A `Script` encapsulates the parsed and compiled form of a JavaScript program,
+//! enabling efficient execution and interaction with the JavaScript engine.
+//!
+//! ## Overview
+//! - **Source Representation**: Links to the source text, including metadata
+//!     like line numbers and column offsets for debugging / stack traces.
+//! - **Bytecode**: Contains the compiled bytecode generated during parsing,
+//!     optimized for SpiderMonkey's interpreter.
+//!
+//! For more details, refer to the [ECMAScript Specification on Scripts and Modules](https://tc39.es/ecma262/#sec-scripts).
+
+use std::{marker::PhantomData, path::Path, ptr::NonNull};
+
+use mozjs::{
+    jsapi::{Compile1, JSScript, JS_ExecuteScript},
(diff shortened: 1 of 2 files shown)


zcabter (approved):
Other than the minor comment, LGTM

crates/jstz_engine/src/script.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        // TODO(https://linear.app/tezos/issue/JSTZ-196):
+        // Remove once we have a proper way of rooting things.
+        // The script is rooted in the context in `eval`, but this doesn't work due to lifetimes.
+        // So we need to transmute it here.
+        let rooted_script: Script<'_, compartment::Ref<'_>> =
+            unsafe { std::mem::transmute(script) };
+
+        // Evaluate the script
+        let res = rooted_script.evaluate(&mut cx);
```
zcabter, line 128: I wonder if you thought about Script holding a `*mut JSContext` since, in most cases,

crates/jstz_engine/src/script.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Evaluates a script and returns its return value
+
+    // TODO(https://linear.app/tezos/issue/JSTZ-210):
+    // Add support for error handling / exceptions instead of using `Option`
+    // TODO(https://linear.app/tezos/issue/JSTZ-211):
+    // TODO: `JSVal` is not safe, we should return a safe wrapper instead
+    pub fn evaluate<'b, S>(&self, cx: &'b mut Context<S>) -> Option<JSVal>
+    where
+        S: InCompartment<C> + CanAlloc,
+        'a: 'b,
```
zcabter, line 65: Unused 'a?

johnyob: It's used in the type of script `Script<'a, C>`

crates/jstz_engine/src/script.rs
johnyob: Not always, scripts are nothing more than compiled bits of bytecode. They require a *context* in order to execute. We should also be weary of implicitly storing contexts since they are the capability for doing anything in SpiderMonkey (and we almost should never implicitly copy / clone a context)

PR #662, opened 2024-11-14
feat(jstz_engine): define the `Trace` and `Finalize` traits

johnyob:
# Context

We'll need to trace native Rust objects. To do so, we'll need a `Trace` trait (and a `Finalize` for drops)

# Description

This PR:
- Defines the `Trace` and `Finalize` traits + some helpful macros
- Implements the `Trace` and `Finalize` trait for many common types.

# Manually testing the PR

Code isn't easily testable (relies on calling out to SM GC where `Tracer` is defined). I suggest we ignore the codecov check for this PR -- these implementations will become tested as we implement APIs on top of them

### DIFF
--- a/crates/jstz_engine/src/gc/trace.rs
+++ b/crates/jstz_engine/src/gc/trace.rs
@@ -1,0 +1,395 @@
+use std::{
+    any::TypeId,
+    marker::PhantomData,
+    num::{
+        NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize,
+        NonZeroU128, NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize,
+    },
+    ops::Deref,
+    path::{Path, PathBuf},
+    pin::Pin,
+    rc::Rc,
+    sync::Arc,
+};
+
+use mozjs::jsapi::{
+    gc::{
+        TraceExternalEdge, TraceExternalEdge1, TraceExternalEdge3, TraceExternalEdge5,
+        TraceExternalEdge6, TraceExternalEdge7, TraceExternalEdge8, TraceExternalEdge9,
+    },
+    jsid as JSId, BigInt as JSBigInt, GCTraceKindToAscii, JSFunction, JSObject, JSScript,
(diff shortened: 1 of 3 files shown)


crates/jstz_engine/src/gc/trace.rs
```diff
@@ -0,0 +1,378 @@
...
+use mozjs::jsapi::{
+    gc::{
+        TraceExternalEdge, TraceExternalEdge1, TraceExternalEdge3, TraceExternalEdge5,
+        TraceExternalEdge6, TraceExternalEdge7, TraceExternalEdge8, TraceExternalEdge9,
+    },
+    jsid as JSId, BigInt as JSBigInt, GCTraceKindToAscii, JSFunction, JSObject, JSScript,
+    JSString, Symbol as JSSymbol, TraceKind, Value as JSValue,
+};
+
+/// Visitor passed to trace methods. ALl managed pointers
```
zcabter, line 24:
```suggestion
/// Visitor passed to trace methods. ALL managed pointers
```

zcabter (dismissed):
LGTM

crates/jstz_engine/src/gc/trace.rs
```diff
@@ -0,0 +1,378 @@
...
+// All elements inside the array are correctly marked.
+unsafe impl<T: Trace, const N: usize> Trace for [T; N] {
+    custom_trace!(this, mark, {
+        for v in this {
+            mark(v);
+        }
+    });
+}
+
+impl<T: Trace> Finalize for Arc<T> {}
```
zcabter, line 168: Do we need another `impl` for Rc<T>?

zcabter (commented):
`Finalize` implementation seems to be left empty throughout and also is left empty in the follow up PR (https://github.com/jstz-dev/jstz/pull/680/files). Am I missing something here? Do we need it?

crates/jstz_engine/src/gc/trace.rs
johnyob: Yes, I must've missed it!

zcabter (approved):

johnyob:
CI failed due to coverage. I'm bypassing the check based on @zcabter approval.

PR #644, opened 2024-11-08
feat(ci): integrates stainless client lib gen into ci

zcabter:
# Context
This PR sets up [Stainless](https://app.stainlessapi.com/trilitech/projects) Github Action to automatically generate the client library and push changes to `jstz-client` repo.  The auto generated bindings will be useful as we continue to evolve the Jstz node endpoints and communication bindings. 

Generated code: https://github.com/jstz-dev/jstz-client

# Description
* Adds generate client job
* Adds stainless.yaml config
* Represents operation hash as byte array

## Evaluation
Evaluation doc:  https://docs.google.com/document/d/1kU_xjPYoHaqgmewAZlJDpkJ25FTNnccq8Wb88YQhcFk/edit?usp=sharing

| | Description | Stainless
|-- | -- | --
| Code quality | Coherent generated code: (M1) Separation by resource, (M2) Sensible calling conventions and method names, (M3) Able to return the body type and Response, async, (M4) Idiomatic use of Typescript (M5) Package size(M6) Fetch configurability | (M1) + Types /Classes are separated by resource(M1) - Sub-resources are child types of resource ie. Type for subkey list is Jstz.Accounts.Kv.SubkeyListResponse(M2) + Calling convention are resource based ie. client.accounts.kv.subkeys.list(address)(M3) + Response are returned as Core.APIPromise<T>, an awaitable promises that returns T or can be converted into Response with asResponse() or withResponse()(M4) + OO Typescript(M4) + Namespaces to void conflicts(M6) + Client customisation eg. timeout, retries, custom fetch, agent for managing connections (M6) ~ Streamable response support via fetch API - cannot mark endpoint as returning a stream (could be an OAS definition problem)
Runtime compatibility |   | Node.js, Deno, Bun, browsers, and variousedge runtimes, as well as both CommonJS (CJS) and EcmaScript Modules (ESM).
Mock and testing |   | `Fetch` can be mocked
Customisability | Degree and ease of customising the code generation | + Method names, model references, and organisation can be controlled via stainless.yml config+ Patching with custom code by replaying patches for each generation
Language support |   | + Typescript- No Rust, not in roadmap~ SDK docs in roadmap
Developer experience |   | - No offline generation- No LSP to aid writing stainless configs~ Basic documentation
CI Integration | Degree and ease of integrating into CI | + GHA support- Dependent on Stainless platform to perform the code generation
Compatibility with OpenAPI 3.1 | Utopia generates OAS v3.1 | Yes

# Manually testing the PR

PR #663, opened 2024-11-14
Fix: serde on enums

zcabter:
# Context
This PR fixes the awkward external tagging that serde uses for Json serialisation with the following changes:
1. Cryptos enums are marked  with `#[serde(untagged)]`
2. Non-cryptos enums are marked as  `#[serde(tag = "_type")]`

These changes ensures that we get generated code that makes sense in Typescript from https://github.com/jstz-dev/jstz/pull/644

# Manually testing the PR
Code should build, tests should pass
`make`

### DIFF
--- a/crates/jstz_node/openapi.json
+++ b/crates/jstz_node/openapi.json
@@ -465,7 +465,10 @@
             ],
             "title": "RunFunction"
           }
-        ]
+        ],
+        "discriminator": {
+          "propertyName": "_type"
+        }
       },
       "DeployFunction": {
         "type": "object",
@@ -662,13 +665,13 @@
         "type": "object",
         "required": [
           "hash",
-          "inner"
+          "result"
         ],
         "properties": {
           "hash": {
             "type": "string"
           },
-          "inner": {
+          "result": {
             "$ref": "#/components/schemas/ReceiptResult"
           }
         }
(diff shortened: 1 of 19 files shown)


crates/jstz_node/openapi.json
```diff
@@ -422,26 +422,46 @@
       "Content": {
         "oneOf": [
           {
-            "type": "object",
-            "required": [
-              "DeployFunction"
-            ],
-            "properties": {
-              "DeployFunction": {
+            "allOf": [
```
huancheng-trili, line 425: Code changes look fine to me, but the generated spec seems a bit weird, for instance
```json
      "Content": {
        "oneOf": [
          {
            "allOf": [
              {
                "$ref": "#/components/schemas/DeployFunction"
              },
              {
                "type": "object",
                "required": [
                  "#type"
                ],
                "properties": {
                  "#type": {
                    "type": "string",
                    "enum": [
                      "DeployFunction"
                    ]
                  }
                }
              }
            ]
          },
          ...
```
ideally we want just the first element of `allOf` without `allOf` itself. Is this doable?

zcabter: The correct way to add discriminator in OAS is the `discriminator` field, however, utoipa only generates this if the enum is marked as `serde(untagged)` which then generates json like:
```
{
  "Ok": {
    "account": "tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx",
    "updated_balance": 20
  }
}
```
Instead of:
```
{
  "Ok": {
    "_type": "Deposit",
    "account": "tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx",
    "updated_balance": 20
  }
}
```
(Note: I changed `#type` to `_type`)

The `_type` will end up being a field on ts structs and is used to discriminate between variants in match statements.

huancheng-trili: I get your point and that's fine. I see that now [ReceiptResult](https://github.com/jstz-dev/jstz/pull/663/files#diff-6a082d1d5ac19606f82468d82595327a0ea053178edbaeee2e6121b5e1f036d1R84) has this `content = "inner"` thing. Should we apply this to all enums? Then we can get rid of this `allOf` thing here so that it looks like the `ReceiptResult` [schema](https://github.com/jstz-dev/jstz/pull/663/files#diff-ed955db8c0de308210b3093969021b0e41c7cd2af81a8ca3c665e1e69fe89141R788).

zcabter: No, I think its rather unconventional to apply this format for types that are not meant to be container types

zcabter: I think tag + content is useful for container types but we shouldn't overuse it as it becomes a burden in the generated code.

huancheng-trili (approved):

crates/jstz_node/openapi.json
huancheng-trili: leaving this unresolved temporarily for https://github.com/jstz-dev/jstz/pull/644#pullrequestreview-2442122051

PR #664, opened 2024-11-14
fix(jstzd): revise baker check in jstzd test

huancheng-trili:
# Context

Changes the way the test checks if baker has been destroyed because it does not always work.

# Description

The original idea of checking if a baker process is indeed destroyed is to check if there is any process with `octez-baker` in its name with `ps`. This works fine when there is only one such test running each time, but it becomes problematic when there are multiple tests and even multiple workers running in parallel -- the baker launched by each test might have been destroyed, but `ps` also shows baker processes launched by other tests.

This check is now performed with a different approach. ~The baker binary is copied to a temp file and then jstzd launches the baker with that clone. The test still looks for that name in `ps` and removes that clone at the end of the test. This approach should be fine for now since it's only used this jstzd_test~. One public method is implemented in jstzd server so that we can check if the baker is alive directly by asking the underlying baker task. This is not really ideal, but let's keep it for now so that we don't get blocked here.

# Manually testing the PR

`jstzd_test` should still pass.

### DIFF
--- a/crates/jstzd/src/task/jstzd.rs
+++ b/crates/jstzd/src/task/jstzd.rs
@@ -207,6 +207,14 @@ impl JstzdServer {
         self.state.write().await.server_handle.replace(handle);
         Ok(())
     }
+
+    pub async fn baker_healthy(&self) -> bool {
+        if let Some(v) = &self.state.read().await.jstzd {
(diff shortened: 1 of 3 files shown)


huancheng-trili:
Offline discussion: expose a simple public method in `Jstzd` that checks if the baker is still running and use this in the test.

PR #665, opened 2024-11-14
test(jstzd): add jstz rollup preimages

ryutamago:
# Context

[Spawn rollup in jstzd](https://linear.app/tezos/issue/JSTZ-164/spawn-octezrollupnode-in-jstzd)
# Description

* add the jstz preimages and kernel installer, parameters_ty under `jstzd/resource/jstz_rollup`


huancheng-trili (commented):
I guess this is fine. Please move the files from `jstzd/tests` to `jstzd/resources` as these files are to be used in production.

zcabter (changes requested):
Please do not check in the preimages. You should write a `build.rs` that copies the kernel, runs the installer kernel at compile time and packages the preimages with the binary.

PR #673, opened 2024-11-20
feat(octez): serialise octez client config

huancheng-trili:
# Context

Part of JSTZ-121; in preparation for #655.

# Description

Serialise octez client config. Note that the `base_dir` field is omitted if it's a temp path because it's generated and managed by the program and users should not see it when they fetch the serialised config from the jstzd server.

# Manually testing the PR

* Unit test: added new test cases

### DIFF
--- a/crates/octez/src/async/client.rs
+++ b/crates/octez/src/async/client.rs
@@ -3,6 +3,7 @@ use jstz_crypto::{
     public_key::PublicKey, public_key_hash::PublicKeyHash, secret_key::SecretKey,
 };
 use regex::Regex;
+use serde::Serialize;
 use std::{
     ffi::OsStr,
     fmt,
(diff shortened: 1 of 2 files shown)


crates/octez/src/async/client.rs
```diff
@@ -19,14 +20,19 @@ const DEFAULT_BINARY_PATH: &str = "octez-client";
 
 type StdOut = String;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize)]
 pub struct OctezClientConfig {
     binary_path: PathBuf,
+    #[serde(skip_serializing_if = "skip_serializing_base_dir")]
```
zcabter, line 26: I think useful to see the exact temp dir that the client uses to quickly inspect baker logs, keys, highwatermarks etc. mostly for advanced users

huancheng-trili: Updated. octez node config needs to be updated later on to be consistent.

ryutamago (approved):
🛳️

zcabter (approved):

PR #684, opened 2024-11-22
feat(jstzd): insert jstz bootstrap contracts

huancheng-trili:
# Context

Part of JSTZ-168.

[JSTZ-168](https://linear.app/tezos/issue/JSTZ-168/jstzd-run-cli)

# Description

Insert jstz smart contracts into user config.

Note that user contracts whose addresses collide with those reserved for jstz contracts will overwrite jstz contracts. This aligns with the current implementation where bootstrap contracts in the base parameter file take precedence, even if it means that jstz won't launch in such cases.

# Manually testing the PR

* Unit test: added test cases

### DIFF
--- a/crates/jstzd/src/config.rs
+++ b/crates/jstzd/src/config.rs
@@ -1,6 +1,10 @@
 #![allow(dead_code)]
+use std::path::{Path, PathBuf};
+
 use crate::task::jstzd::JstzdConfig;
+use crate::{EXCHANGER_ADDRESS, JSTZ_NATIVE_BRIDGE_ADDRESS};
 use anyhow::{Context, Result};
+use octez::r#async::protocol::{BootstrapContract, ProtocolParameter};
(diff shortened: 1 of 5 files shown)


crates/jstzd/src/config.rs
```diff
@@ -105,6 +113,51 @@ fn populate_baker_config(
...
+        .await
+        .context(format!(
+            "error loading bootstrap contract '{contract_name}'"
+        ))?;
+        contracts.push(BootstrapContract::new(script, 1_000_000, Some(hash)).unwrap());
+    }
+    Ok(contracts)
+}
+
+async fn read_json_file(path: PathBuf) -> anyhow::Result<serde_json::Value> {
```
ryutamago, line 132: nit: can remove anyhow

ryutamago (approved):
💪

PR #686, opened 2024-11-25
feat(jstzd): listen to os signals

huancheng-trili:
# Context

Part of JSTZ-189.

[JSTZ-189](https://linear.app/tezos/issue/JSTZ-189/monitor-health-and-capture-signals-after-calling-jstzdserverrun)

# Description

Terminate jstzd on SIGTERM and SIGINT.

Note that we still need to find a way to shut down all processes when SIGKILL hits the server process. What happens now is that the subprocesses become orphans after the server is terminated.

# Manually testing the PR

* Integration test: added back `default_config` with SIGTERM and added one test case with SIGINT

### DIFF
--- a/crates/jstzd/src/lib.rs
+++ b/crates/jstzd/src/lib.rs
@@ -8,6 +8,7 @@ pub mod jstz_rollup_path {
     include!(concat!(env!("OUT_DIR"), "/jstz_rollup_path.rs"));
 }
 use std::process::exit;
+use tokio::signal::unix::{signal, SignalKind};
 
 include!("../build_config.rs");
 pub const JSTZ_ROLLUP_ADDRESS: &str = "sr1PuFMgaRUN12rKQ3J2ae5psNtwCxPNmGNK";
@@ -35,8 +36,14 @@ async fn run(port: u16, config: JstzdConfig) {
         exit(1);
     }
 
-    server.wait().await;
+    let mut sigterm = signal(SignalKind::terminate()).unwrap();
+    let mut sigint = signal(SignalKind::interrupt()).unwrap();
 
+    tokio::select! {
+        _ = server.wait() => (),
+        _ = sigterm.recv() => (),
+        _ = sigint.recv() => (),
+    };
     println!("Shutting down");
     server.stop().await.unwrap();
 }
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
lgtm! 
For handling the SIGKILL, i'm thikning we could assgin subprocesses to the same process group as the parent? 
if we could set the process group ID of the main server process to its own PID, and make sure all the octez tasks inherit it, if we kill the jstzd server process, it should clean up all the sub processses. wdyt ? @huancheng-trili @zcabter

huancheng-trili:
Killing the server process won't take down the entire process group. Child processes do have a new parent but they keep running in the same group. It's tricky because we don't have control over the octez executables and thus we cannot make them listen to anything relevant. (btw with the current setup octez process are already in the same process group as the server process.)

If we want to guarantee that the sandbox is clean, we can use a container with everything running inside and create a wrapper server that handles path mapping and spawns the container, but that's a bit messy. For now we can probably just ignore this issue and deal with it later on.

ryutamago:
@huancheng-trili 
> Killing the server process won't take down the entire process group
you are right here. i've done more digging into it and seems like we could set the `PR_SET_PDEATHSIG` for Child Processes:
if we modify the the spawning of child processes to include a pre_exec closure that sets the `PR_SET_PDEATHSIG` to say `SIGTERM`, This allows that if the parent process is terminated (e.g., via SIGKILL), the child processes receive a SIGTERM and can shut down gracefully. The downside of this approach is that we will have to implment the listener for all of the octez child processes (node, baker, rollup)

crates/jstzd/src/main.rs
```diff
@@ -1,7 +1,21 @@
...
+#[derive(Parser, Debug)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+enum Commands {
+    /// Run the sandbox
+    Run { config_path: Option<String> },
```
zcabter, line 12:
```suggestion
    Run { config: Option<String> },
```

zcabter:
We discussed potential solutions offline, namely requiring a sidecar process. I think we can put SIGKILL support on the backlog for now

PR #687, opened 2024-11-25
Leounoki jstz 165/spawn jstz node 2

ryutamago:
# Context

[task](https://linear.app/tezos/issue/JSTZ-165/spawn-jstznode-in-jstzd)

The ci fails for the same reason as [the rollup PR](https://github.com/jstz-dev/jstz/tree/leounoki-jstz-164/spawn-rollup-3)
# Description

* spawn jstz-node in jstzd
* introduce a type aliass `Shared` and refactored the jstzd by adding a blanket implementation so we can avoid having to write `Arc<RwLock<..` everywhere

# Manually testing the PR

add integration tests the runs successfully locally 
```
nix flake check
```


crates/jstzd/src/task/jstzd.rs
```diff
@@ -1,4 +1,7 @@
 use super::{
+    // child_wrapper::Shared,
```
huancheng-trili, line 2:
```suggestion
```

PR #688, opened 2024-11-25
feat(jstzd): implement shutdown signal in jstzd server

huancheng-trili:
# Context

Part of JSTZ-168.

[JSTZ-168](https://linear.app/tezos/issue/JSTZ-168/jstzd-run-cli)

# Description

Implement shutdown signal in jstzd server.

If a program spins up jstzd server and keeps it running, it needs a way to know when jstzd server is terminated, e.g. after someone calls the shutdown endpoint. Otherwise, the program would just hang there not knowing when it can exit. This PR adds a `wait` method to achieve that. When `wait` finally returns, it means that jstzd server is shut down.

# Manually testing the PR

* Integration test: updated jstzd test to verify that `wait` actually works

### DIFF
--- a/crates/jstzd/src/task/jstzd.rs
+++ b/crates/jstzd/src/task/jstzd.rs
@@ -16,7 +16,11 @@ use octez::r#async::{
 };
 use serde::Serialize;
 use std::sync::Arc;
-use tokio::{net::TcpListener, sync::RwLock, task::JoinHandle};
+use tokio::{
+    net::TcpListener,
+    sync::{oneshot, RwLock},
+    task::JoinHandle,
+};
 
 struct Jstzd {
     octez_node: Arc<RwLock<OctezNode>>,
@@ -182,6 +186,7 @@ struct ServerState {
     jstzd_config_json: serde_json::Map<String, serde_json::Value>,
     jstzd: Option<Jstzd>,
     server_handle: Option<JoinHandle<()>>,
+    shutdown_tx: Option<oneshot::Sender<()>>,
 }
 
 #[async_trait]
@@ -195,6 +200,7 @@ impl AsyncDrop for JstzdServerInner {
 pub struct JstzdServer {
     port: u16,
     inner: Arc<AsyncDropper<JstzdServerInner>>,
+    shutdown_rx: Option<oneshot::Receiver<()>>,
 }
 
 impl JstzdServer {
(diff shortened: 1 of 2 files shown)


crates/jstzd/src/task/jstzd.rs
```diff
@@ -182,6 +186,7 @@ struct ServerState {
     jstzd_config_json: serde_json::Map<String, serde_json::Value>,
     jstzd: Option<Jstzd>,
     server_handle: Option<JoinHandle<()>>,
+    tx: Option<mpsc::Sender<()>>,
```
ryutamago, line 189: we only need to send a single shutdown signal so we can use `oneshot`

ryutamago: nit:
shutdown_tx 
to make it more explicit

crates/jstzd/src/task/jstzd.rs
```diff
@@ -195,10 +200,12 @@ impl AsyncDrop for JstzdServerInner {
 pub struct JstzdServer {
     port: u16,
     inner: Arc<AsyncDropper<JstzdServerInner>>,
+    rx: mpsc::Receiver<()>,
```
ryutamago, line 203: nit:
shutdown_rx 
to make it more explicit

crates/jstzd/tests/jstzd_test.rs
```diff
@@ -27,11 +27,18 @@ async fn jstzd_test() {
...
+            .put(&format!("http://localhost:{}/shutdown", jstzd_port))
+            .send()
+            .await
+            .unwrap();
+    });
+
+    tokio::time::timeout(tokio::time::Duration::from_secs(30), jstzd.wait())
         .await
-        .unwrap();
+        .expect("should not wait too long for the server to be taken down");
```
ryutamago, line 41: maybe we can create a separate test for a shut down ?

crates/jstzd/src/task/jstzd.rs
huancheng-trili: what's the problem with using mpsc::Sender?

crates/jstzd/tests/jstzd_test.rs
huancheng-trili: why?

crates/jstzd/src/task/jstzd.rs
ryutamago: oneshot is more efficient as it's optimized for single-use :) i assume for shutdown signal we just need to send one message, once.

crates/jstzd/tests/jstzd_test.rs
ryutamago: i just realized jstzd server is only shut down via the shutdonw signal and otherwise will keep running as a daemon so it will never `Drop`

ryutamago (approved):
🚀

PR #691, opened 2024-11-27
ci: terminate slow tests

huancheng-trili:
# Context

Closes JSTZ-208.

[JSTZ-208](https://linear.app/tezos/issue/JSTZ-208/ci-terminate-slow-tests)

# Description

Terminate tests that take too much time in CI to save resources and prevent unexpected long-running tests from blocking the CI runner.

# Manually testing the PR

A test branch contains changes in this PR and a dummy test that does nothing but sleeps for 2 minutes. Workflow run: https://github.com/jstz-dev/jstz/actions/runs/12035957074/job/33556133877

The test was terminated and the overall check failed as expected.

### DIFF
--- a/nix/crates.nix
+++ b/nix/crates.nix
@@ -15,7 +15,7 @@
   src = let
     # Include all Rust files / cargo related files
     # Include all files in the `contracts` and `crates` directories
-    regexes = [".*\.toml$" ".*\.rs$" "^\.cargo.*$" "^Cargo.lock$" "^crates.*$" "^contracts.*$"];
+    regexes = [".*\.toml$" ".*\.rs$" "^\.cargo.*$" "^Cargo.lock$" "^crates.*$" "^contracts.*$" "\.config.*$"];
   in
     lib.sourceByRegex (lib.cleanSource ../.) regexes;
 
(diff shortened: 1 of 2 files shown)


nix/crates.nix
```diff
@@ -135,7 +135,7 @@ in {
         # Don't run the `jstz_api` integration tests until they've been paralellized
         #
         # Note: --workspace is required for --exclude. Once --exclude is removed, remove --workspace
-        cargoNextestExtraArgs = "--workspace --test \"*\" --exclude \"jstz_api\"";
+        cargoNextestExtraArgs = "--workspace --test \"*\" --exclude \"jstz_api\" --config-file ${src}/.config/nextest.toml";
```
johnyob, line 138: Won't this be needed for the `cargo-llvm-cov` check as well?

johnyob: Looking at `cargo llvm-cov test` the option isn't available, but it is for `cargo llvm-cov nextest`. #658 changes the command to `nextest` -- though this could be split out/included in this PR to avoid PR dependencies

johnyob (changes requested):
See comment about `cargo-llvm-cov`

nix/crates.nix
huancheng-trili: I see. Do you want to split that `nextest` change into an individual PR? You can apply `config-file` in that PR after this one is merged or I can wait for that and patch `cargo-llvm-cov` in this PR.

johnyob (approved):

PR #693, opened 2024-11-27
feat(node): add hash operation endpoint

zcabter:
# Context
Adds an endpoint to hash operations in Jstz so that user's can have consistent hashing without wasm dependency. A slightly nicer UX would be an indutry standard binary encoding that is consistent across rust, typescript and other languages but existing solutions are not effortless drop ins and will require additional effort at a later time.

# Description
* Adds the hashing endpoint
*  Updates openapi
* 

# Manually testing the PR
`cargo test -p jstz_node api_doc_regression`

### DIFF
--- a/crates/jstz_node/openapi.json
+++ b/crates/jstz_node/openapi.json
@@ -378,6 +378,43 @@
         }
       }
     },
+    "/operations/hash": {
+      "post": {
+        "tags": [
+          "Operations"
+        ],
+        "summary": "Returns the hash of an Operation",
+        "operationId": "hash_operation",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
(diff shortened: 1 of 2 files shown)


huancheng-trili (commented):
`POST /operations/hash` doesn't sound right to me, though I don't think there's a better verb for this operation.

What is the actual use case of this endpoint? Can we replace it by making `POST /operations` return the hash?

zcabter:
> `POST /operations/hash` doesn't sound right to me, though I don't think there's a better verb for this operation.
> 
> What is the actual use case of this endpoint? Can we replace it by making `POST /operations` return the hash?

We need the operation hash (as part of signature generation) to inject the operation :(

huancheng-trili (approved):

PR #694, opened 2024-11-29
feat(jstzd): use symlink for rollup

ryutamago:
# Context

[use symlink instead of copy for rollup preimages](https://linear.app/tezos/issue/JSTZ-207/symlink-for-rollup-preimages)

# Description

* Use symlink instead of copying all the preimages

# Manually testing the PR

the rollup test passess meaning the preimages are used correctly

### DIFF
--- a/crates/jstzd/src/task/octez_rollup.rs
+++ b/crates/jstzd/src/task/octez_rollup.rs
@@ -1,5 +1,5 @@
 use std::{
-    fs::{self, create_dir_all},
+    os::unix::fs::symlink,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


huancheng-trili:
I don't think that `cannot unlink` error has anything to do with your changes. It only happens when a build fails. Nix performs a cleanup after each build and removes build directories. Build directories are always empty at the end of successful builds while failed builds have leftover files inside (because they never reach the end when things get removed.) Somehow nix has the assumption that build directories should always be empty and thus the error occurs for failed builds when nix tries to remove the non-empty directories. This is also partly why the runner machine needs the additional garbage collection I implemented.

ryutamago:
> I don't think that `cannot unlink` error has anything to do with your changes. It only happens when a build fails. 

I wonder what made the build fail when i pushed my changes

PR #700, opened 2024-12-04
fix(jstzd): use protocol version from proto params

huancheng-trili:
# Context

If baker path is not provided in user config, jstzd should fall back to the default baker path _according to the target protocol version_. If protocol version is specified in user config, that protocol version should be referenced by the baker config builder. The bug is that it still defaults to `alpha` when protocol version is provided in user config.

# Description

Fixes the bug by checking the protocol version provided during baker config population.

# Manually testing the PR

* Unit test: updated one test case
* Manual test: launched the sandbox locally and confirmed that this was fixed


ryutamago (approved):
💪

PR #703, opened 2024-12-04
feat(jstzd): add jstz rollup to config

ryutamago:
# Context

[use jstz kernel in config](https://linear.app/tezos/issue/JSTZ-238/use-jstz-rollup-kernel-in-configrs)

# Description
* move some of the const values in config.rs 
* replace the dummy rollup with jstz rollup

# Manually testing the PR

add unit test to check if the rollup config is correctly set
```
cargo test --package jstzd --lib -- config::tests::build_config --exact --show-output
```

### DIFF
--- a/crates/jstzd/src/config.rs
+++ b/crates/jstzd/src/config.rs
@@ -1,8 +1,10 @@
 use rust_embed::Embed;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::task::jstzd::JstzdConfig;
-use crate::{EXCHANGER_ADDRESS, JSTZ_NATIVE_BRIDGE_ADDRESS, JSTZ_ROLLUP_ADDRESS};
+use crate::{
+    jstz_rollup_path, EXCHANGER_ADDRESS, JSTZ_NATIVE_BRIDGE_ADDRESS, JSTZ_ROLLUP_ADDRESS,
+};
 use anyhow::{Context, Result};
 use jstz_node::config::JstzNodeConfig;
 use octez::r#async::endpoint::Endpoint;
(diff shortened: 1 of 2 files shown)


crates/jstzd/src/config.rs
```diff
@@ -461,19 +499,24 @@ mod tests {
             .unwrap();
         let params = serde_json::from_str::<serde_json::Value>(&buf).unwrap();
 
-        // one bootstrap account should have been inserted: the activator account
+        // one bootstrap account should have been inserted: the activator account and the rollup operator account
```
huancheng-trili, line 502:
```suggestion
        // two bootstrap accounts should have been inserted: the activator account and the rollup operator account
```

crates/jstzd/src/config.rs
```diff
@@ -69,18 +80,18 @@ pub(crate) async fn build_config(
...
         octez_client_config.base_dir().into(),
         SmartRollupHash::from_base58_check(JSTZ_ROLLUP_ADDRESS).unwrap(),
-        "bootstrap1".to_string(),
-        "dummy-kernel".into(),
+        ROLLUP_OPERATOR_ACCOUNT_ALIAS.to_string(),
+        jstz_rollup_path::kernel_installer_path(),
     )
-    .set_data_dir(RollupDataDir::Temp)
-    .set_rpc_endpoint(&Endpoint::localhost(8000))
+    .set_data_dir(RollupDataDir::TempWithPreImages {
```
huancheng-trili, line 91: Note from the offline discussion: this will need to be updated later on so that
* `Config` gets one more member `octez_rollup: Option<OctezRollupConfigBuilder>`
* if `config.octez_rollup` is provided, use that as the config builder and fill in the missing bits, e.g. `octez_node_endpoint`
* if not, create a new builder with a default path, possibly taken from some environment variable so that we can easily control the default path (which will be essential when it comes to packaging and distribution)

Builds with this implementation do not work outside the jstz repo. We're proceeding with this for now just so that we can at least get a concrete (though not distributable) full implementation.

crates/jstzd/src/task/jstzd.rs
```diff
@@ -97,6 +102,10 @@ impl JstzdConfig {
         &self.baker_config
     }
 
+    pub fn octez_rollup_config(&self) -> &OctezRollupConfig {
```
huancheng-trili, line 105: Is this in use now?

ryutamago: yes in the test! probably need it later when returing the config to the endpoint too

huancheng-trili:
fix your test please

huancheng-trili (approved):

PR #705, opened 2024-12-06
Jstzd: serialize rollup config

ryutamago:
# Context

[serialzie rollup config](https://linear.app/tezos/issue/JSTZ-241/implemetn-rollup-config-endpoint)

# Description

serialize the rollup config and return as part of the config endpoint
# Manually testing the PR

add integration test to jstzd
add unit test

### DIFF
--- a/crates/octez/src/async/rollup.rs
+++ b/crates/octez/src/async/rollup.rs
@@ -3,7 +3,7 @@ use crate::unused_port;
 use super::{bootstrap::SmartRollupPvmKind, endpoint::Endpoint};
 use anyhow::Result;
 use http::Uri;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::{
     path::{Path, PathBuf},
     str::FromStr,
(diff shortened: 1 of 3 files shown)


crates/octez/src/async/rollup.rs
```diff
@@ -257,4 +257,42 @@ mod test {
...
+
+        let json = serde_json::to_value(config.clone()).unwrap();
+        assert_eq!(
+            json,
+            serde_json::json!({
+                "binary_path": "octez-smart-rollup-node",
+                "octez_client_base_dir": "/base_dir",
+                "octez_node_endpoint": "http://localhost:1234",
+                "pvm_kind": "wasm_2_0_0",
+                "data_dir": {
```
huancheng-trili, line 285: This should show the path to the actual data directory used by the rollup node

huancheng-trili (changes requested):

crates/octez/src/async/rollup.rs
ryutamago: we discussed in slack that we could skip this for now as the actaul data direcotory path isn't avaiable until the rollup is spawned
Ticket: https://linear.app/tezos/issue/JSTZ-243/include-rollup-data-dir-in-config

huancheng-trili (approved):

PR #724, opened 2024-12-31
fix(wpt): run async tests and collect results

huancheng-trili:
# Context

Closes JSTZ-256.
[JSTZ-256](https://linear.app/tezos/issue/JSTZ-256/run-async-wpt-test-cases)

# Description

Fixes the wpt runner in jstz_api so that it covers async test cases. Asynchronous tests, including those wrapped in promises, were not properly executed because `eval` only runs the synchronous tests and puts async ones in the task queue. The runtime task queue was never touched afterwards and thus the async tests were simply abandoned. With changes in this PR, async test cases can be executed as expected, and also because of this, we don't need the `Null` test status because at the end of the async task chain, test suites are completed and therefore there is always a meaningful test status returned.

Note that the task queue must be executed after all synchronous tests finish, i.e. after all `eval` calls are made. I haven't fully figured out why, but my guess is that the test harness API has some assumptions about the test flow for each test suite, e.g. after async tests are done, other things are ignored, such that some test cases are dropped if `run_jobs` is called right after an `eval` call.

# Manually testing the PR

`env UPDATE_EXPECT=1 cargo test --test wpt`

and more tests show up in `wptreport.json`. These tests are already defined in the test suites but they were never executed.

### DIFF
--- a/crates/jstz_api/tests/wpt.rs
+++ b/crates/jstz_api/tests/wpt.rs
@@ -102,6 +102,8 @@ pub struct TestsResult {
 #[derive(Default, Debug, Trace, Finalize, JsData)]
 pub struct TestHarnessReport {
     #[unsafe_ignore_trace]
+    // `status` is an Option because it is set at the end of a test suite
+    // and we need a placeholder for it before that.
     status: Option<WptTestStatus>,
     #[unsafe_ignore_trace]
     subtests: Vec<WptSubtest>,
@@ -282,6 +284,9 @@ pub fn run_wpt_test_harness(bundle: &Bundle) -> JsResult<Box<TestHarnessReport>>
         }
     }
 
+    // Execute promises after all sync tests have completed after `eval` returns
+    rt.run_jobs();
+
     // Return the test harness report
 
     let test_harness_report = {
@@ -305,7 +310,10 @@ fn run_wpt_test(
             return Ok(WptReportTest::new(WptTestStatus::Err, vec![]));
         };
 
-        let status = report.status.clone().unwrap_or(WptTestStatus::Null);
+        // It should be safe to unwrap here because each test suite should have a
+        // status code attached after it completes. If unwrap fails, it means something
+        // is wrong and we should fix that
+        let status = report.status.clone().unwrap();
 
         let subtests = report.subtests.clone();
 
--- a/crates/jstz_api/tests/wptreport.json
+++ b/crates/jstz_api/tests/wptreport.json
@@ -30,7 +30,7 @@
                         "message": null
                       }
                     ],
-                    "status": "Null",
+                    "status": "Ok",
                     "metrics": {
                       "passed": 4,
                       "failed": 0,
@@ -82,7 +82,7 @@
                     "message": null
                   }
                 ],
-                "status": "Null",
+                "status": "Ok",
                 "metrics": {
                   "passed": 6,
                   "failed": 0,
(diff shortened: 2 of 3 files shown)


crates/jstz_api/tests/wpt.rs
```diff
@@ -282,6 +282,9 @@ pub fn run_wpt_test_harness(bundle: &Bundle) -> JsResult<Box<TestHarnessReport>>
         }
     }
 
+    // Execute promises after all sync tests have completed after `eval` returns
+    rt.run_jobs();
+
```
johnyob, line 287: Prefer our event loop over Boa's:
```suggestion
    // Execute promises after all sync tests have completed after `eval` returns
    rt.run_event_loop()

```

crates/jstz_api/tests/wpt.rs
```diff
@@ -305,7 +308,7 @@ fn run_wpt_test(
             return Ok(WptReportTest::new(WptTestStatus::Err, vec![]));
         };
 
-        let status = report.status.clone().unwrap_or(WptTestStatus::Null);
+        let status = report.status.clone().unwrap();
```
johnyob, line 311: Why not use `WptTestStatus::Null`? Seems like this change would only introduce panics that we can recover from?

huancheng-trili: From what I see in `WptTestStatus ::try_from`, I suppose that `WptTestStatus::Null` is not an actual status returned by wpt. Since now `WptTestStatus::Null` is replaced by other status codes in the report, I assume that every regular test report must come with some status code, and therefore it should be fine to call unwrap if all tests are properly executed. If unwrap fails, it means something was not executed properly and we should fix that. I guess I can also remove `Option` from `TestHarnessReport::status` if the assumption that there will always be some status code in a report is correct.

johnyob: Yes, `WptTestStatus::Null` was designed to capture when the test harness failed to set status without the entire integration test panicking

johnyob: I guess we could trust the tests themselves more -- removing the `Option` as you suggest

johnyob: Having thought about this argument, I'm more convinced that we can remove `WptTestStatus::Null`

huancheng-trili: It seems that I cannot just remove `Option` now. The harness report status is set at the end of a test suite, so before that we need a placeholder like `None`. I'll leave the `Option` as is for now.

crates/jstz_api/tests/wpt.rs
huancheng-trili: I tried this but `rt.run_event_loop().await` simply hanged there. I'm not sure if this has anything to do with the async context blocking itself or so. Do you have any idea? If not, I can look into this later.

johnyob: I'm not too sure why that would be the case -- it is likely something odd with async since `run_event_loop` essentially loops `poll_event_loop` until `Poll::Ready(())` is returned. I just noticed that we override the default implementation of `run_jobs` for our job queue so `rt.run_jobs()` should fine anyway

huancheng-trili:
@johnyob do you have any other comments? I added some notes in `wpt.rs` explaining the use of `unwrap` and `Option`.

johnyob (approved):

PR #725, opened 2024-12-31
feat: implement smart function hash

ryutamago:
# Context

[Introduce smart function hash](https://linear.app/tezos/issue/JSTZ-251/introduce-smartfunctionhash)

In preparation for [254](https://linear.app/tezos/issue/JSTZ-254/migrate-to-newaddress-type)

# Description

1. Introduce a `Hash` trait that both the public key hash for user account and sf hash for sf account will use
```
pub trait Hash<'de>:
    Sized
    + ...
{
    fn to_base58(&self) -> String;

    fn from_base58(data: &str) -> Result<Self>;

    fn as_bytes(&self) -> &[u8];

    fn digest(data: &[u8]) -> Result<Self>;
}
```
2. make `PublicKeyHash` implement the JstzHash

3. Introduce `SmartFunctionHash` that implements the JstzHash. this hash will be used for sf address and starts with `TZ1...`

# Manually testing the PR

unit test

### DIFF
--- a/crates/jstz_core/src/kv/outbox.rs
+++ b/crates/jstz_core/src/kv/outbox.rs
@@ -333,7 +333,7 @@ pub enum OutboxError {
 
 #[cfg(test)]
 mod test {
-    use jstz_crypto::public_key_hash::PublicKeyHash;
+    use jstz_crypto::{hash::Hash, public_key_hash::PublicKeyHash};
     use tezos_data_encoding::nom::NomReader;
     use tezos_smart_rollup::{
         michelson::{
--- a/crates/jstz_cli/src/bridge/deposit.rs
+++ b/crates/jstz_cli/src/bridge/deposit.rs
@@ -1,3 +1,4 @@
+use jstz_crypto::hash::Hash;
 use log::{debug, info};
 
 use crate::{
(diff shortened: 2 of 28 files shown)


crates/jstz_core/src/kv/outbox.rs
```diff
@@ -387,10 +387,8 @@ mod test {
     fn flush_empty_snapshot_flushes_rollup_queue() {
         let mut host = MockHost::default();
 
-        let accounts = [
-            PublicKeyHash::digest(b"account1").unwrap(),
-            PublicKeyHash::digest(b"account2").unwrap(),
-        ];
+        let digest = PublicKeyHash::digest(b"account1");
```
huancheng-trili, line 390: It seems that these lines don't need to be changed?

ryutamago: fixed!

huancheng-trili (approved):
Just one comment

huancheng-trili (approved):

huancheng-trili (changes requested):
Wait why's the new commit including changes out of scope?

huancheng-trili (approved):
sorry false alert

PR #728, opened 2025-01-02
introduce new address type

ryutamago:
# Context

[introduce new address](https://linear.app/tezos/issue/JSTZ-252/introduce-enum-newaddress)
in preparation for [254](https://linear.app/tezos/issue/JSTZ-254/migrate-to-newaddress-type)

# Description

introduce NewAddress type that supports both user(tz..) and sf(KT1..) address

# Manually testing the PR

add unit test

### DIFF
--- a/crates/jstz_proto/src/context/new_account.rs
+++ b/crates/jstz_proto/src/context/new_account.rs
@@ -1,0 +1,179 @@
+use std::{
+    fmt::{self, Display},
+    str::FromStr,
+};
+
+use crate::error::{Error, Result};
+use boa_gc::{empty_trace, Finalize, Trace};
+use jstz_crypto::public_key_hash::PublicKeyHash;
+use jstz_crypto::smart_function_hash::SmartFunctionHash;
+use serde::{Deserialize, Serialize};
+
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    Finalize,
+    // TODO: Add back when renamed
+    // https://linear.app/tezos/issue/JSTZ-253/remove-old-accountrs
+    // ToSchema,
+)]
+#[serde(untagged)]
+/// TODO: rename to Address
+// https://linear.app/tezos/issue/JSTZ-253/remove-old-accountrs
+pub enum NewAddress {
+    User(PublicKeyHash),
+    SmartFunction(SmartFunctionHash),
+}
+
+unsafe impl Trace for NewAddress {
+    empty_trace!();
+}
+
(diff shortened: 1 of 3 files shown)


crates/jstz_proto/src/context/new_account.rs
```diff
@@ -0,0 +1,178 @@
...
+    Deserialize,
+    Finalize,
+    ToSchema,
+)]
+#[serde(untagged)]
+/// TODO: rename to Address
+// https://linear.app/tezos/issue/JSTZ-253/remove-old-accountrs
+pub enum NewAddress {
+    User(PublicKeyHash),
+    SF(SmartFunctionHash),
```
huancheng-trili, line 31: Why use the acronoym here?

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -0,0 +1,178 @@
...
+        match &s[..3] {
+            "KT1" => Ok(NewAddress::SF(SmartFunctionHash::from_str(s)?)),
+            "tz1" | "tz2" | "tz3" => Ok(NewAddress::User(PublicKeyHash::from_str(s)?)),
+            _ => Err(Error::InvalidAddress),
+        }
+    }
+}
+
+impl NewAddress {
+    pub fn check_is_user(&self) -> Result<()> {
```
huancheng-trili, line 60: Why not return a `bool`? `Error::InvalidAddress` does not really make sense here even if we want a `Result` since all addresses wrapped by `NewAddress` instances are already valid.

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -0,0 +1,178 @@
...
+    Clone,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    Finalize,
+    ToSchema,
```
huancheng-trili, line 24: Perhaps this should not implement `ToSchema` at this point since we would then get a schema called `NewAddress`.

crates/jstz_proto/src/context/new_account.rs
ryutamago: i've changed the error name to `AddressTypeMismatch` but will keep the signature as it is. 
This function will be used everywhere in the new account type and i think it's cleaner to use it like the following rather than repeating the logic of checking the boolean and throwing error everytime
```
check_is_user()?
...
```

crates/jstz_proto/src/context/new_account.rs
ryutamago: felt it was long, idm either way so used `SmartFunction` instead!

huancheng-trili (approved):

PR #729, opened 2025-01-02
fix sf api

ryutamago:
# Context

quick fix in preparation for supporting the [kt1 address](https://linear.app/tezos/issue/JSTZ-171/change-smart-function-addresses-to-kt1)

# Description

* fix the length for the sf function api `call` and `create`
* the sf account won't have a nonce so removed the logic that increments it

# Manually testing the PR

existing test works


johnyob (approved):
LGTM 🚀

PR #730, opened 2025-01-02
feat(jstz_cli): make Config::load async

huancheng-trili:
# Context

Part of JSTZ-248.
[JSTZ-248](https://linear.app/tezos/issue/JSTZ-248/revise-jstz-cli-so-that-it-works-with-jstzd)

In preparation for #732.

# Description

We want to make jstz_cli make http requests to fetch jstzd config so that it can be used against jstzd. The `load` method is synchronous now, but we cannot use `reqwest::blocking` because this method is normally called inside async functions elsewhere. This means that we need to change this method itself and other functions where this is called to async.

Note that `load_sync` and `reload_sync` are created as a synchronous version of the `load` and `reload` methods just for the sandbox. Since the sandbox has nothing to do with jstzd, it will never make http calls to fetch jstzd config and thus it can keep using the synchronous version. This avoids unnecessary and tricky updates to the sandbox code.

# Manually testing the PR

Manually tested jstz_cli and it still worked.

I'm not fixing the low test coverage at this moment.

### DIFF
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -32,8 +32,8 @@ impl User {
     }
 }
 
-fn add_smart_function(alias: String, address: Address) -> Result<()> {
-    let mut cfg = Config::load()?;
+async fn add_smart_function(alias: String, address: Address) -> Result<()> {
+    let mut cfg = Config::load().await?;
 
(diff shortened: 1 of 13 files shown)


huancheng-trili:
Not fixing test coverage in this PR

ryutamago (approved):
LGTM!

PR #732, opened 2025-01-02
feat(jstz_cli): make jstz_cli compatible with jstzd

huancheng-trili:
# Context

Completes JSTZ-248.
[JSTZ-248](https://linear.app/tezos/issue/JSTZ-248/revise-jstz-cli-so-that-it-works-with-jstzd)

# Description

Make `jstz_cli` compatible with `jstzd`. This is more like a temporary patch before `jstz sandbox` is taken down.

To use `jstz_cli` with a running `jstzd` instance, an environment variable `USE_JSTZD` needs to be set to `true` or `1`. `jstz_cli` will then load config from jstzd server at 127.0.0.1:55555 (default jstzd server endpoint) and find the nodes and client directory.

Note that if a user uses `jstz_cli` with `jstz sandbox` first and then switch to `jstzd`, the next usage against `jstz sandbox` will fail because the data directory path stored in the config will still point to `jstzd`. I think this is fine for now since we don't plan to keep `jstz sandbox` alive in the long run.

# Manually testing the PR

* Unit test: added some test cases for the testable logic
* Manual run:
  ```sh
  # with jstzd running
  $ USE_JSTZD=1 cargo run -- account create test
  $ USE_JSTZD=1 cargo run -- bridge deposit --from tz1TGu6TN5GSez2ndXXeDX6LgUDvLzPLqgYV --to <test_address> --amount 1 --network dev
  ```
  and it worked as expected.

### DIFF
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -19,8 +19,12 @@ use crate::{
     sandbox::{
         SANDBOX_JSTZ_NODE_PORT, SANDBOX_LOCAL_HOST_ADDR, SANDBOX_OCTEZ_NODE_RPC_PORT,
     },
-    utils::AddressOrAlias,
+    utils::{using_jstzd, AddressOrAlias},
 };
+
+// hardcoding it here instead of importing from jstzd simply to avoid adding jstzd
+// as a new depedency of jstz_cli just for this so that build time remains the same
+const JSTZD_SERVER_BASE_URL: &str = "http://127.0.0.1:55555";
 
 pub fn jstz_home_dir() -> PathBuf {
     if let Ok(value) = env::var("JSTZ_HOME") {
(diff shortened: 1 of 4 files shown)


crates/jstz_cli/src/config.rs
```diff
@@ -323,6 +358,28 @@ impl Config {
         Ok(config)
     }
 
+    fn fill_in_jstzd_config(&mut self, jstzd_config: JstzdConfig) -> Result<()> {
+        self.sandbox = Some(SandboxConfig {
+            octez_client_dir: PathBuf::from_str(&jstzd_config.octez_client.base_dir)?,
```
ryutamago, line 363: isn't this value going to get overwriten by the sandbox?

crates/jstz_cli/src/sandbox/mod.rs
```diff
@@ -48,6 +48,11 @@ pub async fn restart(detach: bool) -> Result<()> {
 }
 
 pub async fn exec(command: Command) -> Result<()> {
+    if using_jstzd() {
+        return Err(anyhow::anyhow!(
```
ryutamago, line 52: nit:
bail!

crates/jstz_cli/src/config.rs
huancheng-trili: Why?

ryutamago: inside `run_sandbox` in daemon.rs

huancheng-trili: Yes, but this method is only called when jstzd is in use. Logic in `daemon.rs` does not get executed. This PR is not about replacing `jstz sandbox ...` with `jstzd` and `USE_JSTZD=1` does not change the behaviour of `jstz sandbox start`.

ryutamago (approved):

PR #749, opened 2025-01-09
fix(operation): remove serde tag from operation variants

zcabter:
# Context
Fixes an issue with json serialization  of `Content` printing the discriminator (`_type`) twice which causes problems when deserializing. 

This PR also identifies an issue with the binary encoding that is broken because `tezos_cryptos` cannot decode enums with `serde(untagged)`. This will be fixed in a follow up PR.

# Description
* Removes serde tag `_type` from operation struct types.

# Manually testing the PR
`cargo nextest run -p jstz_proto`

### DIFF
--- a/crates/jstz_crypto/src/public_key.rs
+++ b/crates/jstz_crypto/src/public_key.rs
@@ -132,4 +132,16 @@ mod test {
             "tz3QxNCB8HgxJyp5V9ZmCVGcTm6BzYc14k9C"
         );
     }
+
+    #[test]
+    #[ignore = "Fails because deserialization cannot handle untagged crypto enums"]
+    // FIXME: https://linear.app/tezos/issue/JSTZ-272/fix-binary-round-trip-for-tezos-cryptos
+    fn bin_round_trip() {
+        let pk = PublicKey::from_base58(TZ1).unwrap();
+        let bin = bincode::serialize(&pk).unwrap();
+        // Error message:
+        //      Result::unwrap()` on an `Err` value: DeserializeAnyNotSupported
+        let decoded = bincode::deserialize::<PublicKey>(bin.as_ref()).unwrap();
+        assert_eq!(pk, decoded);
+    }
 }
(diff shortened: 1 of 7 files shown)


huancheng-trili:
I'm implementing a fix for this whole serde tag thing. This PR can be closed.

crates/jstz_crypto/src/public_key.rs
```diff
@@ -132,4 +132,16 @@ mod test {
             "tz3QxNCB8HgxJyp5V9ZmCVGcTm6BzYc14k9C"
         );
     }
+
+    #[test]
+    #[ignore = "Fails because deserialization cannot handle untagged crypto enums"]
```
huancheng-trili, line 137: Why not just remove the tags and enable these tests in `jstz_crypto`?

huancheng-trili (commented):
Okay I realised that the fix that I had in mind is more difficult to handle than I thought, so let's at least remove the tags if you cannot implement your fix soon enough.

crates/jstz_crypto/src/public_key.rs
zcabter: Are you referring to `serde(untagged)`? If naively removed, the json encoding will not be correct.

zcabter:
> Okay I realised that the [fix](https://github.com/jstz-dev/jstz/commit/a476b415f8ea243e79acd4de4cf411e6688472f8) that I had in mind is more tricky than I thought, even if it preserves the API schema and keeps bincode happy, so let's at least remove the tags if you cannot implement your fix soon enough.

I've implemented the fix here: https://github.com/jstz-dev/jstz/pull/761  

I would still prefer to merge this then the followup #761

crates/jstz_crypto/src/public_key.rs
huancheng-trili: Yes, but how come? If we remove them, the structure of messages will change (or be reverted to what it used to be,) but then bincode should be able to deserialise the messages right?

huancheng-trili (approved):

crates/jstz_crypto/src/public_key.rs
zcabter: Yes, the bincode will be succeed but json encoding will tagged :(  we want untagged json cryptos which are simple strings as that is their canonical way of representation in json.

PR #752, opened 2025-01-10
feat(jstzd): set default history mode for octez node

huancheng-trili:
# Context

Closes JSTZ-276.
[JSTZ-276](https://linear.app/tezos/issue/JSTZ-276/run-octez-node-in-rolling-mode-by-default)

# Description

Set the default history mode for octez node in jstzd to `rolling:15`. This makes octez node keep a bit more context and should reduce the chance of the rollup node falling into degraded mode when it takes too much time loading the kernel. With the current sandbox parameters, octez node should keep data within a 2-minute window with `rolling:15`, which should be sufficient for the rollup node.

Some changes are made to the octez crate. I put them in this PR as well since those are trivial getter implementations.

# Manually testing the PR

* Unit test: added tests
```sh
$ cargo test --lib config::tests::patch_octez_node_config
running 1 test
test config::tests::patch_octez_node_config ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 27 filtered out; finished in 0.00s
```

### DIFF
--- a/crates/jstzd/src/config.rs
+++ b/crates/jstzd/src/config.rs
@@ -1,3 +1,4 @@
+use octez::r#async::node_config::{OctezNodeHistoryMode, OctezNodeRunOptionsBuilder};
 use octez::unused_port;
 use rust_embed::Embed;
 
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
🛳️

PR #753, opened 2025-01-12
Remove enum from smart function hash

ryutamago:
# Context
Since there's only one type of contract address on tezos, we don't need to use an enum.

# Description

* removed the enum from sfh
* renamed KT1 to `SmartFunctionHash`

# Manually testing the PR

existing tests work


huancheng-trili:
It probably should not be done like this now due to the wip fix in #761. You need to check with @zcabter.

zcabter:
Tbh SF hash doesn't need to be an enum, that should fix the problem with openapi entirely?

ryutamago:
@zcabter 
since the serde untagged is fixed, i think we can keep the enum for extensibility

zcabter (approved):

PR #757, opened 2025-01-12
remove validation check for new address 

ryutamago:
# Context

preparation step for [758](https://github.com/jstz-dev/jstz/pull/758) where new account is introduced

# Description

1. The validation check for the address type is removed as we intend to use the sf hash address directly in the following PRs.
2. the value for entry.or_insert_with doesn't need a `Default` trait

# Manually testing the PR

existing test pass

### DIFF
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -342,11 +342,10 @@ pub enum Command {
     },
 }
 
+// TODO: use sf address
+// // TODO: https://linear.app/tezos/issue/JSTZ-260/add-validation-check-for-address-type
 fn parse_smart_function_address(s: &str) -> Result<NewAddress> {
     let address = NewAddress::from_str(s).map_err(|e| user_error!("{}", e))?;
-    address
-        .check_is_smart_function()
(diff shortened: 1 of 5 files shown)


crates/jstz_cli/src/account.rs
```diff
@@ -342,11 +342,10 @@ pub enum Command {
     },
 }
 
+// TODO: use sf address
+// // TODO: https://linear.app/tezos/issue/JSTZ-260/add-validation-check-for-address-type
 fn parse_smart_function_address(s: &str) -> Result<NewAddress> {
     let address = NewAddress::from_str(s).map_err(|e| user_error!("{}", e))?;
-    address
```
huancheng-trili, line 347: Can we keep this until smart function address is introduced? Otherwise, this seems a bit risky without the check.

ryutamago: this wasn't meant to be here in first place as the codebase doesn't support KT1.. yet. i could reintroduce this once the migration to the new account is supported and we could start using KT1.. .

huancheng-trili (approved):

huancheng-trili:
change your commit title please

PR #758, opened 2025-01-12
jstz proto: introduce new account

ryutamago:
# Context

first step of migrarting to a new account
[task link](https://linear.app/tezos/issue/JSTZ-257/introduce-new-account-type)
in preparation for [255](https://linear.app/tezos/issue/JSTZ-255/migrate-to-newaccount-type)

# Description

Added a new account context that handles address of both user and smart function. For a clearer api, we should use `SmartFunctionHash` for methods that's only intended for sf. This can be implemented after the migration to the new account type is completed. 

In the next step, the old account should be replaced with the new one.
# Manually testing the PR
Unit tests are added.

### DIFF
--- a/crates/jstz_proto/src/context/new_account.rs
+++ b/crates/jstz_proto/src/context/new_account.rs
@@ -1,16 +1,85 @@
 use std::{
     fmt::{self, Display},
+    result,
     str::FromStr,
 };
 
 use crate::error::{Error, Result};
 use bincode::{Decode, Encode};
+use boa_engine::{Context, JsError, JsResult, Module, Source};
 use boa_gc::{empty_trace, Finalize, Trace};
+use jstz_core::{
+    host::HostRuntime,
+    kv::{Entry, Transaction},
+};
 use jstz_crypto::hash::Hash;
 use jstz_crypto::public_key_hash::PublicKeyHash;
 use jstz_crypto::smart_function_hash::SmartFunctionHash;
 use serde::{Deserialize, Serialize};
+use tezos_smart_rollup::storage::path::{self, OwnedPath, RefPath};
 use utoipa::ToSchema;
+
+pub type Amount = u64;
+
+#[derive(
+    Clone,
+    Copy,
+    Default,
+    Debug,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
+    ToSchema,
+    Encode,
+    Decode,
+)]
+pub struct Nonce(pub u64);
+
+impl Nonce {
+    pub fn next(&self) -> Nonce {
+        Nonce(self.0 + 1)
+    }
+
+    pub fn increment(&mut self) {
+        self.0 += 1
+    }
+}
+
+impl Display for Nonce {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+// Invariant: if code is present it parses successfully
+#[derive(
+    Default, PartialEq, Eq, Debug, Clone, Serialize, Deserialize, ToSchema, Encode, Decode,
+)]
+#[schema(
+    format = "javascript",
+    example = "export default (request) => new Response('Hello world!')"
+)]
+pub struct ParsedCode(pub String);
+impl From<ParsedCode> for String {
+    fn from(ParsedCode(code): ParsedCode) -> Self {
+        code
+    }
+}
+impl Display for ParsedCode {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> result::Result<(), fmt::Error> {
+        Display::fmt(&self.0, formatter)
+    }
+}
+impl TryFrom<String> for ParsedCode {
+    type Error = JsError;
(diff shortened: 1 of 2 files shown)


zcabter:
To aid with review, it would be helpful to add the specific commands that execute your tests.

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -85,172 +142,682 @@ impl NewAddress {
...
+pub enum Account {
+    User {
+        amount: Amount,
+        nonce: Nonce,
+    },
+    SmartFunction {
+        amount: Amount,
+        nonce: Nonce,
+        function_code: ParsedCode,
+    },
```
zcabter, line 158: Its often better to factor these out into their own type and use single unnamed field variants.

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -85,172 +142,682 @@ impl NewAddress {
...
+            Self::SmartFunction { nonce, .. } => Ok(nonce),
+        }
+    }
+
+    pub fn create_smart_function(
+        hrt: &impl HostRuntime,
+        tx: &mut Transaction,
+        // TODO: use smart function hash
+        // https://linear.app/tezos/issue/JSTZ-260/add-validation-check-for-address-type
+        addr: &NewAddress,
```
zcabter, line 228: We should generate the address from the parsed code

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -85,172 +142,682 @@ impl NewAddress {
...
+    ) -> Result<()> {
+        let account = Self::get_mut(hrt, tx, addr)?;
+        match account {
+            Self::SmartFunction { function_code, .. } => {
+                *function_code = new_function_code.try_into()?;
+                Ok(())
+            }
+            _ => Err(Error::AddressTypeMismatch),
+        }
+    }
```
zcabter, line 270: Probably don't need this as we expect function code to be immutable on creation.

crates/jstz_proto/src/context/new_account.rs
ryutamago: i am keeping the same logic from the old Account type. i think we could improve on the api once we can verify that the changes here maintains the same behaviour as [it](https://github.com/jstz-dev/jstz/pull/762) touches a lot of files

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -85,172 +142,682 @@ impl NewAddress {
...
+
+    fn balance_mut<'a>(
```
zcabter, line 272: Super nit: Move private fns above pub ones

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -85,172 +142,682 @@ impl NewAddress {
...
+            let (host, mut tx) = setup_test_env();
+            let (user_addr, sf_addr) = create_test_addresses();
+
+            // Test inserting user account
+            let user_account = Account::User {
+                amount: 100,
+                nonce: Nonce(0),
+            };
+            assert!(user_account.try_insert(&host, &mut tx, &user_addr).is_ok());
+
```
zcabter, line 756: Perhaps you need to assert that the account can be retrieved

crates/jstz_proto/src/context/new_account.rs
zcabter: The old logic was trying to implement account abstraction at the `Account` level but we are not doing that anymore. Now is the appropriate time to change it since we have a clear way to distinguish user and SF accounts.

crates/jstz_proto/src/context/new_account.rs
zcabter: Oops I didnt realize it was used in the repl. Let's keep it around, could you add a todo to conditionally compile for repl?

zcabter:
LGTM overall. Good job. 🚀

crates/jstz_proto/src/context/new_account.rs
ryutamago: > The old logic was trying to implement account abstraction at the Account level
i'm not sure if i follow this. what does this account abstraction have to do with migration to KT1 address? 

can i create a ticket to fix this once the [migratoin pr](https://github.com/jstz-dev/jstz/pull/762) is merged. i think it's easier to make a separate PR once the migration with the current interface is merged. i want to avoid mixing up the code for migration and adding a new logic for the change in the interface.

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -87,171 +156,733 @@ impl NewAddress {
...
+        let account = Self::get_mut(hrt, tx, addr)?;
+        match account {
+            Self::SmartFunction(SmartFunctionAccount { function_code, .. }) => {
+                Ok(&function_code.0)
+            }
+            Self::User(_) => Err(Error::AddressTypeMismatch),
+        }
+    }
+
+    // Used only in repl, conditionally compile
```
zcabter, line 288: Nit: Prefix with `TODO`

zcabter (approved):

PR #761, opened 2025-01-13
fix(crypto): handle enums when bin deserialize

zcabter:
# Context
Fixes the issue where bincode [fails with the existence of `serde(tag/untagged)` attribute](https://github.com/bincode-org/bincode/issues/272). This happens because when serde will drive bincode to `deseriaize_any`  when tagged/untagged which bincode does not support. Luckily, bincode v2 takes a totally new approach, where its derive is independent but compatible with serde. 

This PR fixes the problem by upgrading to bincode v2. 

Closes: https://linear.app/tezos/issue/JSTZ-272/fix-binary-round-trip-for-tezos-cryptos

# Description
* Since bincode v2 is not fully released yet, we upgrade to the latest rc ie `2.0.0-rc.v3`
* Crypto variants are wrapped with new type so that `bincode::{Encode, Decode}` can be implemented for them. The implementation is generated by `impl_bincode_for_hash` macro
* Bincode v2 deprecates `de/serialize` in favour of `encode_*/decode_*` which needs updating throughout the code base. Legacy configuration is kept to maintain the same byte string encoding behaviour.
*  Added test for cryptos to assert that json is encoded as simple strings while binary round trips succeed.
* Move openapi regressions to unit tests instead of integration test for faster run cycles
* Remove `Ord` and `PartialOrd` from cryptos because I don't think it makes sense to order crypto hashes? and it is not used anywhere
* Added `cargo-expand` utility to nix output since it is very useful to see how macros expand

# Manually testing the PR
```
make build && make test-unit
```
➕  no changes in openapi schema 🎉


zcabter:
It turns out that adding `serde(tag)` causes the visitor to be driven to `deserialize_any` (the thing that blows up bincode) which causes issues with de/serializing `Content` 😓

huancheng-trili (approved):
Then I suppose we need to implement `Encode/Decode` for those enums with `#[serde(tag = "_type")]` as well?

zcabter:
> Then I suppose we need to implement `Encode/Decode` for those enums with `#[serde(tag = "_type")]` as well?

Yes, you're right. I've added `Encode/Decode` accordingly.

PR #763, opened 2025-01-13
feat(jstzd): create dockerfile

huancheng-trili:
# Context

Closes JSTZ-278.
[JSTZ-278](https://linear.app/tezos/issue/JSTZ-278/create-docker-file-for-jstzd)

# Description

Create a dockerfile that builds jstzd as an image based on a designated octez image. One feature flag is added specifically for this docker build because
* jstzd in the image must not suffer from the slow PVM issue, which means the base image needs either the old version v20.3 or newer versions with the slow PVM fix
* jstzd in the image needs to have quebec support, which means the old version v20.3 is not suitable as the base image

and therefore we need to use a newer version as the base image. Since alpha is not yet stabilised, it's better that we don't allow users to use alpha at all, so this new feature flag includes `octez/disable-alpha`. The default protocol version used by jstzd will be quebec.

# Manually testing the PR

```
$ cd $(git rev-parse --show-toplevel) && \
docker build -t ghcr.io/jstz-dev/jstz/jstz:0.1.0 --build-arg OCTEZ_TAG=master_d21359af_20241220171048 \
-f crates/jstzd/Dockerfile .
...
$ docker images
REPOSITORY                   TAG           IMAGE ID       CREATED          SIZE
ghcr.io/jstz-dev/jstz/jstz   0.1.0         b5f373cb09e1   58 seconds ago   1.64GB
```

### DIFF
--- a/crates/jstzd/Cargo.toml
+++ b/crates/jstzd/Cargo.toml
@@ -51,6 +51,7 @@ tezos_crypto_rs.workspace = true
 
 [features]
 skip-rollup-tests = []
+build-image = ["octez/disable-alpha", "octez/disable-slow-pvm"]
 
 [[bin]]
 name = "jstzd"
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
🔥

PR #764, opened 2025-01-14
feat(octez): introduce feature 'disable-alpha'

huancheng-trili:
# Context

Part of JSTZ-273.

# Description

Introduce one feature flag `disable-alpha` in the octez crate that hides the alpha protocol version. This is handy when the octez executables for alpha might not be available in an environment, e.g. when the base image of the jstzd image does not have alpha executables at all. This happens when the base image is tagged, e.g. `octez-v20.3`, `octez-v21.0`, including release candidates. The alpha executables are removed when those tagged images are built.

Also replaced Alpha with ParisC in some test cases just to avoid having to add the flag check.

# Manually testing the PR

Tested locally with a feature flag temporarily added to the jstzd crate:
```
[features]
build-image = ["octez/disable-alpha"]
```
The jstzd executable built with this feature flag then launched using `ParisC` as the default protocol version. The jstzd executable built without this feature flag still ran with `Alpha` by default.

### DIFF
--- a/crates/octez/Cargo.toml
+++ b/crates/octez/Cargo.toml
@@ -28,3 +28,6 @@ tempfile.workspace = true
 tezos_crypto_rs.workspace = true
 tezos-smart-rollup-encoding.workspace = true
 tokio.workspace = true
+
+[features]
+disable-alpha = []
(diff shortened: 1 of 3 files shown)


ryutamago (approved):

ryutamago (commented):
actually i find it hard to get double negation like
not(disable-alpha)

instead of disable-alpha, i think using a positive feature name like `alpha-protocol` is easier to understand.
we could enable alpha protocol by default, what do you think? @huancheng-trili 

```
[features]
default = ["alpha-protocol"] 
alpha-protocol = []
```

PR #766, opened 2025-01-14
feat(jstzd): add quebec params

huancheng-trili:
# Context

Part of JSTZ-273.

# Description

Add a parameter file for protocol version quebec. This is basically the same as the current alpha version.

# Manually testing the PR

Launched jstzd locally with a config file `{"protocol":{"protocol":"quebec"}}` and observed that jstzd could understand the protocol version and run normally.


ryutamago (approved):
💪

PR #767, opened 2025-01-14
feat(storage): use bincode for value storage

zcabter:
# Context
This PRs switches the encoding library for value storage from `serde` to `bincode` which ensure binary serialization is independent of serde. This is important because serde manipulation on enums can break bincode, thus an encoding scheme independrnt of serde increases reliability.

# Description
* The main change is o embed encode as part of `Value` and replace `DeserializeOwned` with `bincode::Decode` in `Storage`.
* Changed anywhere using `bincode::serde` to strictly `bincode` to avoid serde-bincode complications
* Cleaned up some redundant serde tags

# Manually testing the PR
```
make build && make test-unit
```

### DIFF
--- a/crates/jstz_core/src/kv/mod.rs
+++ b/crates/jstz_core/src/kv/mod.rs
@@ -3,7 +3,6 @@
 //! This module provides a persistent transactional key-value store.
 
 use boa_gc::{Finalize, Trace};
-use serde::de::DeserializeOwned;
 use tezos_smart_rollup_host::runtime::ValueType;
 use tezos_smart_rollup_host::{path::Path, runtime::Runtime};
 
@@ -42,10 +41,10 @@ pub struct Storage;
 
 impl Storage {
     /// Retrieve a value from the persistent store if it exists
-    pub fn get<V>(rt: &impl Runtime, key: &impl Path) -> Result<Option<V>>
-    where
-        V: Value + DeserializeOwned,
-    {
+    pub fn get<V: bincode::Decode>(
+        rt: &impl Runtime,
+        key: &impl Path,
+    ) -> Result<Option<V>> {
         match rt.store_has(key)? {
             Some(ValueType::Value | ValueType::ValueWithSubtree) => {
                 let bytes = rt.store_read_all(key)?;
@@ -66,11 +65,12 @@ impl Storage {
     }
 
     /// Insert a key-value pair into the persistent store
-    pub fn insert<V>(rt: &mut impl Runtime, key: &impl Path, value: &V) -> Result<()>
-    where
-        V: Value + ?Sized + serde::Serialize,
-    {
-        rt.store_write(key, &value::serialize(value)?, 0)?;
+    pub fn insert<V: Value + ?Sized>(
+        rt: &mut impl Runtime,
+        key: &impl Path,
+        value: &V,
+    ) -> Result<()> {
+        rt.store_write(key, value.encode()?.as_slice(), 0)?;
         Ok(())
     }
 
--- a/crates/jstz_core/src/kv/mod.rs
+++ b/crates/jstz_core/src/kv/mod.rs
@@ -41,14 +41,11 @@ pub struct Storage;
 
 impl Storage {
     /// Retrieve a value from the persistent store if it exists
-    pub fn get<V: bincode::Decode>(
-        rt: &impl Runtime,
-        key: &impl Path,
-    ) -> Result<Option<V>> {
+    pub fn get<V: Value>(rt: &impl Runtime, key: &impl Path) -> Result<Option<V>> {
         match rt.store_has(key)? {
             Some(ValueType::Value | ValueType::ValueWithSubtree) => {
                 let bytes = rt.store_read_all(key)?;
-                let value = value::deserialize(&bytes)?;
+                let value = V::decode(&bytes)?;
                 Ok(Some(value))
             }
             _ => Ok(None),
--- a/crates/jstz_core/src/kv/value.rs
+++ b/crates/jstz_core/src/kv/value.rs
@@ -1,41 +1,30 @@
+use crate::BinEncodable;
+use derive_more::{Deref, DerefMut};
 use std::any::Any;
 use std::fmt::Debug;
 
-use derive_more::{Deref, DerefMut};
-
-use crate::{Error, Result, BINCODE_CONFIGURATION};
-
-pub fn serialize<T: bincode::Encode>(value: &T) -> Result<Vec<u8>> {
-    bincode::encode_to_vec(value, BINCODE_CONFIGURATION).map_err(|err| {
-        Error::SerializationError {
-            description: format!("{err}"),
-        }
-    })
-}
-
-pub fn deserialize<T: bincode::Decode>(bytes: &[u8]) -> Result<T> {
-    let (result, _) =
-        bincode::decode_from_slice(bytes, BINCODE_CONFIGURATION).map_err(|err| {
-            Error::SerializationError {
-                description: format!("{err}"),
-            }
-        })?;
-
-    Ok(result)
-}
-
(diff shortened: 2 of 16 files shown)


crates/jstz_core/src/kv/value.rs
```diff
@@ -52,6 +48,10 @@ where
     fn clone_box(&self) -> Box<dyn Value> {
         Box::new(self.clone())
     }
+
+    fn encode(&self) -> Vec<u8> {
+        bincode::encode_to_vec(self, BINCODE_CONFIGURATION).unwrap()
```
ryutamago, line 53: why not return Result and add a new Error type that wraps `EncodeError` from bincode?

crates/jstz_core/src/kv/value.rs
```diff
@@ -18,28 +24,18 @@ pub fn serialize<T: ?Sized + serde::Serialize>(value: &T) -> Result<Vec<u8>> {
...
-}
-
 /// A key-value 'value' is a value that is can be dynamically
 /// coerced (using `Any`) and serialized.
-pub trait Value: Any + Debug + erased_serde::Serialize {
+pub trait Value: Any + Debug {
     fn as_any(&self) -> &dyn Any;
     fn as_any_mut(&mut self) -> &mut dyn Any;
     fn clone_box(&self) -> Box<dyn Value>;
+    fn encode(&self) -> Vec<u8>;
```
ryutamago, line 33: q: if we are planning to add different encoding method in the future, woundn't it better to separate the encode method into another trait?

crates/jstz_core/src/kv/mod.rs
```diff
@@ -42,10 +41,10 @@ pub struct Storage;
 
 impl Storage {
     /// Retrieve a value from the persistent store if it exists
-    pub fn get<V>(rt: &impl Runtime, key: &impl Path) -> Result<Option<V>>
-    where
-        V: Value + DeserializeOwned,
-    {
+    pub fn get<V: bincode::Decode>(
```
ryutamago, line 44: why not restrict the V: Value + bincode::Decode sincee this Stroage only deal with Value type

crates/jstz_core/src/kv/value.rs
```diff
@@ -2,14 +2,20 @@ use std::any::Any;
...
 use derive_more::{Deref, DerefMut};
-use erased_serde::serialize_trait_object;
-use serde::de::DeserializeOwned;
 
 use crate::{Error, Result, BINCODE_CONFIGURATION};
 
-pub fn serialize<T: ?Sized + serde::Serialize>(value: &T) -> Result<Vec<u8>> {
-    let result =
-        bincode::serde::encode_to_vec(value, BINCODE_CONFIGURATION).map_err(|err| {
+pub fn serialize<T: bincode::Encode>(value: &T) -> Result<Vec<u8>> {
```
ryutamago, line 8: we don't have to fix it in this PR but i noticed some imporvement in the current design and wanted to hear your opinion on it
- The `Value` trait requires types to implement `encode()`
- But deserialization happens through a standalone `deserialize()` function
- This means a type could implement `Value::encode()` using one serialization format
  while the standalone `deserialize()` uses a different format
- This could lead to silent failures where encoded data cannot be properly decoded

For example:
```rust
// A type could do this:
impl Value for MyType {
  fn encode(&self) -> Vec<u8> {
  // Uses JSON serialization
  serde_json::to_vec(self).unwrap()
  }
}
// But the standalone deserialize uses bincode:
let value = deserialize::<MyType>(&bytes)?; // Uses bincode deserialization
```

i was thinking of moving the serialzie/deserialzie into the Value trait to achieve better encapsulation:
```rust
pub trait Value: Any + Debug {
    fn as_any(&self) -> &dyn Any;
    fn as_any_mut(&mut self) -> &mut dyn Any;
    fn clone_box(&self) -> Box<dyn Value>;
    fn encode(&self) -> Result<Vec<u8>>;
    fn decode(bytes: &[u8]) -> Result<Self>
    where
        Self: Sized;
}

impl<T> Value for T
where
    T: Any + Debug + Clone + bincode::Encode + bincode::Decode,
{
    ...
    fn encode(&self) -> Result<Vec<u8>> {
        bincode::encode_to_vec(self, BINCODE_CONFIGURATION).map_err(|err| {
            Error::SerializationError {
                description: format!("{err}"),
            }
        })
    }

    fn decode(bytes: &[u8]) -> Result<Self> {
        let (value, _) = bincode::decode_from_slice(bytes, BINCODE_CONFIGURATION)
            .map_err(|err| Error::SerializationError {
                description: format!("{err}"),
            })?;
        Ok(value)
    }
}
```
This ensures that each type must provide matching encode/decode implementations. 

Another benefit is that it simplifies the storage/transaction APIs - we don't need a trait bound like `Value + Decode` anymore:
```
impl Storage {
    pub fn get<V: Value>(rt: &impl Runtime, key: &impl Path) -> Result<Option<V>> {
        let value = V::decode(&bytes)?;
        ...
```
```        
impl Transaction {
    pub fn get<V>(&mut self, rt: &impl Runtime, key: Key) -> Result<Option<&V>>
    where
        V: Value,  // Only need Value trait bound
    {
        // ...
``` 
would like to hear your opininos :) @zcabter @huancheng-trili

zcabter: `decode` can't be moved into Value because it is not erased (ie. it has generic on one of its members)- so it will not object safe.

crates/jstz_kernel/src/inbox.rs
```diff
@@ -106,8 +106,10 @@ fn read_transfer(
 }
 
 fn read_external_message(rt: &mut impl Runtime, bytes: &[u8]) -> Option<ExternalMessage> {
-    let (msg, _): (ExternalMessage, _) =
-        bincode::serde::decode_from_slice(bytes, BINCODE_CONFIGURATION).ok()?;
+    let result: Result<(ExternalMessage, usize), bincode::error::DecodeError> =
+        bincode::decode_from_slice(bytes, BINCODE_CONFIGURATION);
+    debug_msg!(rt, "{result:?}");
```
huancheng-trili, line 111:
```suggestion
```

huancheng-trili (commented):
Looks okay overall to me. I'll leave the approval to Leo since he raised some questions.

(I'll come back to Leo's long comment later on after the jstzd stuff is done)

crates/jstz_core/src/kv/value.rs
ryutamago: https://github.com/jstz-dev/jstz/pull/773

crates/jstz_core/src/kv/value.rs
zcabter: Fixed

crates/jstz_kernel/src/inbox.rs
zcabter: fixed

crates/jstz_core/src/kv/value.rs
ryutamago: > in order to hide the encoding configuration, the type needs to implement Value which doesn't make sense for types outside of Kv.
what do you mean by this? Do you mean that we could use the same configuration of encode/decode outside of storage Value? I assumed we only need encode/decoding for Kv store from the current codebase :)

ryutamago: feel free to merge:)
https://github.com/jstz-dev/jstz/pull/773

zcabter: Grafted in: https://github.com/jstz-dev/jstz/pull/767/commits/43d81daf9ac67eabd3370072416fb371ecab6e77

ryutamago (approved):
LGTM🚀
we can merge this to main first and make a separte PR for the [kv value issue](https://github.com/jstz-dev/jstz/pull/775) since this PR is already dooing a lot :)

PR #768, opened 2025-01-14
Update jstz doc

ryutamago:
# Context

# Description

* add jstzd doc
* update from tz1 to KT1 accordingly

# Manually testing the PR

### DIFF
--- a/docs/.vitepress/config.mjs
+++ b/docs/.vitepress/config.mjs
@@ -29,6 +29,7 @@ export default defineConfig({
           { text: "Quick Start", link: "/quick_start" },
           { text: "CLI", link: "/cli" },
           { text: "Asset Bridge", link: "/bridge" },
+          { text: "jstzd", link: "/jstzd" },
         ],
       },
 
--- a/docs/api/url.md
+++ b/docs/api/url.md
@@ -11,9 +11,9 @@ let url: URL = new URL(`tezos://${my_function.address}/entrypoint`);
 let url2: URL = new URL("../entrypoint_2", url.href);
 ```
 
-Each `jstz` smart function is assigned a unique address, akin to an IP address, starting with `tz1` when the function is deployed.
+Each `jstz` smart function is assigned a unique address, akin to an IP address, starting with `KT1` when the function is deployed.
 To decode these addresses, `jstz` employs its own URL scheme `tezos://`.
-An example URL for a `jstz` smart function would therefore be `tezos://tz1w42Gt5zkiGaHPm1ya4MgLomgkL1k7Dy7q/`.
+An example URL for a `jstz` smart function would therefore be `tezos://KT19mYzcaYk55KttezwP4TbMrGCDpVuPW3Jw/`.
 
 It's important to note that if the base URL or the resulting URL is not valid, the constructor will raise a `TypeError` exception.
 To check whether URLs can be parsed correctly, you can use the static method [`URL.canParse()`](#canParse).
@@ -31,11 +31,11 @@ if (URL.canParse(relativePath, baseUrl)) {
 You can also modify a URL by setting its properties.
 
 ```typescript
-let url = new URL("tezos://tz1FakeFakeFakeFakeFakeFakeFakeDDy7q/"); // not a valid address, we'll have to change it
+let url = new URL("tezos://KT19mYzcaYk55KttezwP4TbMrGCDpVuPW3Jw/"); // not a valid address, we'll have to change it
 url.hostname = Ledger.selfAddress;
 url.pathname = "accounts";
 url.hash = "#id";
-console.log(url.href); // tezos://tz1../accounts#id
+console.log(url.href); // tezos://KT1../accounts#id
 ```
 
 The [`URLSearchParams`](./url_search_params.md) API may be used to build and manipulate search parameters. To get the search parameters from the URL, you can make use of the `.searchParams` instance property.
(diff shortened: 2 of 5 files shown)


docs/jstzd.md
```diff
@@ -0,0 +1,83 @@
...
+  +======================================+=====================+
+  | tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx | 60000000000         |
+  +--------------------------------------+---------------------+
+  | tz1TGu6TN5GSez2ndXXeDX6LgUDvLzPLqgYV | 40000000000         |
+  +--------------------------------------+---------------------+
+```
+
+By default, jstzd will start on port 55555. You can customize the configuration through a JSON configuration file:
+
+For example to run jstzd on port 66666 and the octez node on port 8888, you can create a config.json file with the following content:
```
ryutamago, line 50: todo: change

huancheng-trili (commented):
@zamrokk would you please take a look at the new jstzd page? It's rather limited at the moment but it will get enriched as we refine the UX, e.g. more details on what kind of thing is configurable. Please suggest anything that you think should be added and/or point out anything that confuses you.

docs/api/url.md
```diff
@@ -11,9 +11,9 @@ let url: URL = new URL(`tezos://${my_function.address}/entrypoint`);
 let url2: URL = new URL("../entrypoint_2", url.href);
 ```
 
-Each `jstz` smart function is assigned a unique address, akin to an IP address, starting with `tz1` when the function is deployed.
+Each `jstz` smart function is assigned a unique address, akin to an IP address, starting with `KT1` when the function is deployed.
 To decode these addresses, `jstz` employs its own URL scheme `tezos://`.
```
zamrokk, line 15: What does the scheme mean as on the Typescript client I am not more able to use it?

docs/api/url.md
```diff
@@ -11,9 +11,9 @@ let url: URL = new URL(`tezos://${my_function.address}/entrypoint`);
...
-An example URL for a `jstz` smart function would therefore be `tezos://tz1w42Gt5zkiGaHPm1ya4MgLomgkL1k7Dy7q/`.
+An example URL for a `jstz` smart function would therefore be `tezos://KT19mYzcaYk55KttezwP4TbMrGCDpVuPW3Jw/`.
```
zamrokk, line 16: Why this change to KT1? Is KT1 generic to all runtimes or only for Tezos L1?

ryutamago: user account in jstz starts from tz1.. and the smart function starts from KT1.. similar to L1

docs/api/url.md
ryutamago: i don't understand this, could you elaborate? 
the change here is because jstz now separtes user addres and smart function address, just like L1

zamrokk: I use the Typescript client recently and I needed to use **HTTP://** and not **tezos://** like in the first version compiled  last November

ryutamago: changes to the client shoudl be updated by @zcabter later.

huancheng-trili:
I'm okay if @zamrokk does not have other questions.

huancheng-trili (approved):

docs/api/url.md
zcabter: @zamrokk When you want to run a smart function, you call the inject endpoint (http). The body of that endpoint is a request object where the url starts with the tezos scheme.

docs/quick_start.md
```diff
@@ -244,24 +244,24 @@ The [`jstz bridge deposit`](bridge.md) command is used to transfer funds from a
 Within the sandbox environment, there are pre-funded L1 accounts `bootstrap1` through `bootstrap5` that you can use.
 
 ```sh
-jstz bridge deposit --from bootstrap1 --to tz1Tp5wSRWiVJwLoT8WqN1yRapdq6UmdRf6W --amount 10000000
```
zcabter, line 247: This shouldn't be changed. We want users to fund their user account here

docs/jstzd.md
```diff
@@ -0,0 +1,90 @@
+# jstzd
+
+jstzd is a daemon service that orchestrates and manages the core components of the `jstz` environment. It serves as the central coordinator for running a jstz client node and the octez binaries (node, baker, rollup node).
```
zcabter, line 3:
```suggestion
jstzd is a daemon service that orchestrates and manages the core components of the `jstz` local sandbox . It serves as the central coordinator for running a jstz client node and the octez binaries (node, baker, rollup node).
```

docs/jstzd.md
```diff
@@ -0,0 +1,90 @@
...
+
+The core features are:
+
+- Automatic process spawning
+- Graceful shutdown
+- Health monitoring
+- Configuration settings
```
zcabter, line 10:
```suggestion
The core features are:

- Almost instant block time and commitment period
- Pre-loaded bootstrap accounts
- Customisability through user supplied configurations
```

docs/.vitepress/config.mjs
```diff
@@ -29,6 +29,7 @@ export default defineConfig({
           { text: "Quick Start", link: "/quick_start" },
           { text: "CLI", link: "/cli" },
           { text: "Asset Bridge", link: "/bridge" },
+          { text: "jstzd", link: "/jstzd" },
```
zcabter, line 32:
```suggestion
          { text: "Jstzd", link: "/jstzd" },
```

docs/jstzd.md
```diff
@@ -0,0 +1,90 @@
+# jstzd
```
zcabter, line 1:
```suggestion
# Jstzd
```

PR #771, opened 2025-01-16
fix(jstzd): change ports assigned to jstz node and jstzd server

huancheng-trili:
# Context

Part of JSTZ-274.
[JSTZ-274](https://linear.app/tezos/issue/JSTZ-274/run-jstzd-with-jstz-sandbox-start)

# Description

Assign a constant port to jstz node. This is required for the container solution because for now the jstz node in jstzd is not configurable, which means there is no way to control the port assigned to jstz node beforehand and thus it's impossible to expose jstz node to jstz CLI in the host environment. Setting the port number to 8933 aligns with the old sandbox.

Also changed the port assigned to jstzd server from 55555 to 54321 because 55555 is a [secretly reserved port in MacOS](https://developer.apple.com/forums/thread/671197).

# Manually testing the PR

Launch jstzd locally to observe that the occupied ports have changed.

### DIFF
--- a/crates/jstz_cli/src/config.rs
+++ b/crates/jstz_cli/src/config.rs
@@ -23,7 +23,7 @@ use crate::{
 
 // hardcoding it here instead of importing from jstzd simply to avoid adding jstzd
 // as a new depedency of jstz_cli just for this so that build time remains the same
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
the mysterious port 55555 in mac!

PR #559, opened 2024-09-12
feat(jstzd): docker container

ryutamago:
# Context

[runnable_image create](https://linear.app/tezos/issue/JSTZ-62/implement-start-for-runnableimage)
[container](https://linear.app/tezos/issue/JSTZ-54/implement-dockercontainer)

# Description

implement the create method on runnable image and docker container

# Manually testing the PR

run ```cargo test --package jstzd```

PR #772, opened 2025-01-16
feat(cli): run containerised jstzd with 'sandbox start'

huancheng-trili:
# Context

Completes JSTZ-274.
[JSTZ-274](https://linear.app/tezos/issue/JSTZ-274/run-jstzd-with-jstz-sandbox-start)

# Description

Enable users to run containerised jstzd with `jstz sandbox --container start`.

Users can choose to launch the sandbox running containerised jstzd with a newly introduced flag `--container`. All sandbox commands with this flag operates on the container.
* `jstz sandbox --container start`: launch a new container running a pre-defined image with jstzd inside. If the image does not exist, the program then tries to pull that from the repository. This command sets some default config values so that the CLI can interact with the octez node and the jstz node inside the container later on. Note that this command generates some temporary files. They are not going to be cleaned up by the CLI after the sandbox is removed and garbage collection is left to the host OS. It's possible to implement something for this, but I'm not going to do it at the moment just to save some time.
* `jstz sandbox --container stop`: kill the running sandbox container if there is any. This command also cleans up the CLI sandbox config stored locally. I'm not sure why in the old sandbox the sandbox config is not removed when a sandbox is torn down, but I think it makes more sense to clean up the config because every sandbox is new.
* `jstz sandbox --container restart`: restart the sandbox if it exists. Note that this does not raise an error if the sandbox is not running. This command simply prints a reminder message and won't start the sandbox.

Note that the module `container` might look a bit awkward. I have to move testable but trivial logic out of the main handlers so that at least we get some test coverage when integration tests are not available.

# Manually testing the PR

* Manual test: built the CLI and jstzd image locally and ran the CLI
* Unit test: added some test cases
```sh
$ cargo test sandbox
running 6 tests                                                                            
test sandbox::container::tests::create_mounts ... ok
test sandbox::container::tests::new_create_container_options ... ok
test sandbox::container::tests::create_port_bindings ... ok     
test sandbox::container::tests::new_create_container_config ... ok   
test sandbox::container::tests::create_exposed_ports ... ok  
test sandbox::container::tests::create_config_file_and_client_dir ... ok           
 
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 9 filtered out; finished in 0.01s
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2757,6 +2757,7 @@ dependencies = [
  "anyhow",
  "boa_engine",
  "boa_gc",
+ "bollard",
  "bs58",
  "clap 4.5.20",
  "clap_complete",
--- a/crates/jstz_cli/Cargo.toml
+++ b/crates/jstz_cli/Cargo.toml
@@ -19,6 +19,7 @@ ansi_term.workspace = true
 anyhow.workspace = true
 boa_engine.workspace = true
 boa_gc.workspace = true
+bollard.workspace = true
 bs58.workspace = true
 clap.workspace = true 
 clap_complete.workspace = true
--- a/crates/jstz_cli/src/main.rs
+++ b/crates/jstz_cli/src/main.rs
@@ -69,8 +69,13 @@ enum Command {
     Bridge(bridge::Command),
 
     /// 🏝️  Start/Stop/Restart the local jstz sandbox
-    #[command(subcommand)]
-    Sandbox(sandbox::Command),
+    Sandbox {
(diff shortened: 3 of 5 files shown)


ryutamago (approved):
LGTM, just a few nits.

Felt like we could have used the docker module if it was merged eg. ([pull image](https://github.com/jstz-dev/jstz/pull/558), [docker container](https://github.com/jstz-dev/jstz/pull/559)) but given that this is the only use case and not much time left for the release let's go wtih this!

PR #774, opened 2025-01-16
feat(client-lib): update client lib

zcabter:
# Context
This PR updates the jstz sdk to use the client lib.  Technically, users don't need to use the client sdk but it has some convenience features that might be cherry-picked into the generated client lib. 

Will review this in the next step as it is getting quite difficult to work across `jstz` and `jstz-client` repos with varying head branches.

# Description
* Updates jstz sdk to use published version jstz-client. Note that this is published to my personal scope for now until we move `@jstz-dev` to trilitech org
* Updates example dashboard
* The js_jstz build was initially failing because the dependency, `undici-types` was being pulled in with 2 none resolvable versions which was causing`ENOTCACHED` issues. The violating dependency versions were being brought in by `node` dependency where v18 (the one client lib uses) and v20 (the one jstz package uses) used 2 unresolvable versions. To circumvent this, I bumped the node version on client lib to v20.

# Manually testing the PR
```
# Jstzd should be running

npm install
cd examples/dashboard
npm run dev 

# Play around with the dashboard


```

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -9,6 +9,7 @@
     "build": "tsc --outDir dist"
   },
   "dependencies": {
+    "@zcabter/client": "0.1.0-alpha.6",
     "jstz_sdk": "file:../../crates/jstz_sdk/pkg"
   }
 }
--- a/crates/jstz_proto/src/operation.rs
+++ b/crates/jstz_proto/src/operation.rs
@@ -117,7 +117,7 @@ pub struct RunFunction {
     pub method: Method,
     /// Any valid HTTP headers
     #[serde(with = "http_serde::header_map")]
-    #[schema(schema_with= openapi::http_headers)]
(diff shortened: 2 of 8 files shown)


packages/sdk/package.json
```diff
@@ -9,6 +9,7 @@
     "build": "tsc --outDir dist"
   },
   "dependencies": {
+    "@jstz-dev/client": "github:jstz-dev/jstz-client#ryan-jstz-201@patch-client",
```
zcabter, line 12: Todo: Do not merge this, this should point to main

ryutamago (approved):
in general, lgtm🛳️
 please update the docs later too!

zcabter:
> in general, lgtm🛳️ please update the docs later too!

Tracked: https://linear.app/tezos/issue/JSTZ-288/update-docs-for-client-lib

PR #762, opened 2025-01-13
Migrate to new account

ryutamago:
# Context

This PR completes the migration to the new account type. 
[255](https://linear.app/tezos/issue/JSTZ-255/migrate-to-newaccount-type)
In the next step, `NewAddress` enums are replaced with `SmartFunctionHash` for parts that should only use sf(e.g. set_code, the self address of ledger, sf runtime api etc)

# Description
Replaced the old Account with the new Account. the KT1 address is now supported.

# Manually testing the PR

* Added unit tests for smart function deployment.
* Added unit tests for rejecting non KT1 address for adding sf alias

* manually tested sf deployment and running.
```
cargo run --bin jstz -- deploy examples/counter.js --network dev
cargo run --bin jstz -- run tezos://KT1J6BkxBK96qQgJURrut4SJz65EQAA5vDhU/ --network dev --trace
```

* manually tested account alias for sf
```
cargo run --bin jstz -- account alias counter_sf KT1J6BkxBK96qQgJURrut4SJz65EQAA5vDhU 

Added smart function: counter_sf -> KT1J6BkxBK96qQgJURrut4SJz65EQAA5vDhU
```

* manually tested deposit / withdrawal to user acount
```
cargo run --bin jstz -- account create test

cargo run --bin jstz -- bridge deposit --from bootstrap1 --to test --amount 2 --network dev

cargo run --bin jstz -- account balance -a test --network dev

cargo run --bin jstz -- bridge withdraw --to tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx --amount 1 --network dev
```

* manually tested deposit/withdraw to a KT1 address
```
cargo run --bin jstz -- bridge deposit --from bootstrap1 --to counter_sf --amount 2 --network dev

cargo run --bin jstz -- account balance -a counter_sf --network dev

argo run --bin jstz -- bridge withdraw --to KT1J6BkxBK96qQgJURrut4SJz65EQAA5vDhU --amount 1 --network dev
```
* manually tested that adding sf alias is rejected for tz1 address
```
cargo run --bin jstz -- account alias sf tz1cmuyEpW2vg6JxVzzRman9bDJXsJadxD2h

error: invalid value 'tz1cmuyEpW2vg6JxVzzRman9bDJXsJadxD2h' for '<ADDRESS>': Invalid smart function address: tz1cmuyEpW2vg6JxVzzRman9bDJXsJadxD2h

```

* manually tested repl 
```
[...]

PR #776, opened 2025-01-20
Use smart function address

ryutamago:
# Context

Currently the `NewAddress` type (an enum of user and sf address) is used for places where only sf is supposed to be used(e.g. get/set function code). This PR refactors the codebase to restrict the type to sf address where possible.
[Use smart function hash](https://linear.app/tezos/issue/JSTZ-260/add-validation-check-for-address-type)

# Description
* re-exported the smart funciton hash and public key hash as `SmartFunctionAddress` and `UserAddress` respectively
* Replace `NewAddress` with `SmartFunctionAddress` where possible 
* Ledger, SmartFunction, Kv api all takes SmartFunctionAddress as it's self.address

# Manually testing the PR

manually test the sf deployment / execution and deposits/withdrawal
check [here](https://github.com/jstz-dev/jstz/pull/762) for the tests conducted

### DIFF
--- a/crates/jstz_cli/src/repl/mod.rs
+++ b/crates/jstz_cli/src/repl/mod.rs
@@ -7,6 +7,7 @@ use jstz_core::{
     kv::Transaction,
     runtime::{self, Runtime},
 };
+use jstz_crypto::{hash::Hash, smart_function_hash::SmartFunctionHash};
 use jstz_proto::executor::smart_function::{register_jstz_apis, register_web_apis};
 use log::{debug, error, info, warn};
 use rustyline::{
@@ -100,7 +101,7 @@ impl Completer for JsHighlighter {
     type Candidate = String;
 }
 
-const DEFAULT_SMART_FUNCTION_ADDRESS: &str = "tz1KqTpEZ7Yob7QbPE4Hy4Wo8fHG8LhKxZSx";
+const DEFAULT_SMART_FUNCTION_ADDRESS: &str = "KT1KRj5VMNmhxobTJBPq7u2kacqbxu9Cntx6";
 const DEFAULT_GAS_LIMIT: usize = usize::MAX;
 const DEFAULT_RANDOM_SEED: u64 = 42;
 
@@ -108,10 +109,14 @@ pub async fn exec(account: Option<AddressOrAlias>) -> Result<()> {
     let cfg = Config::load().await?;
 
     let address = match account {
-        Some(account) => account.resolve(&cfg)?.clone(),
-        None => DEFAULT_SMART_FUNCTION_ADDRESS
-            .parse()
-            .expect("`DEFAULT_SMART_FUNCTION_ADDRESS` is an invalid address."),
+        Some(account) => account
+            .resolve(&cfg)?
+            .as_smart_function()
+            .expect("expected smart function address")
+            .clone(),
+        None => SmartFunctionHash::from_base58(DEFAULT_SMART_FUNCTION_ADDRESS)
+            .expect("`DEFAULT_SMART_FUNCTION_ADDRESS` is an invalid address.")
+            .clone(),
     };
     debug!("resolved `account` -> {:?}", address);
 
--- a/crates/jstz_kernel/src/parsing.rs
+++ b/crates/jstz_kernel/src/parsing.rs
@@ -27,10 +27,14 @@ pub fn try_parse_fa_deposit(
     let receiver = try_parse_contract(&receiver.0)?;
 
     let proxy_smart_function = (proxy_contract)
-        .map(|c| try_parse_contract(&c.0))
+        .map(|c| {
+            if let addr @ NewAddress::SmartFunction(_) = try_parse_contract(&c.0)? {
+                Ok(addr)
+            } else {
+                Err(jstz_proto::Error::AddressTypeMismatch)
+            }
+        })
         .transpose()?;
-    //TODO: validate sf address
-    // https://linear.app/tezos/issue/JSTZ-260/add-validation-check-for-address-type
 
     let amount = ticket
         .amount()
@@ -50,8 +54,11 @@ pub fn try_parse_fa_deposit(
 #[cfg(test)]
 mod test {
 
-    use jstz_crypto::{hash::Hash, smart_function_hash::SmartFunctionHash};
-    use jstz_proto::{context::new_account::NewAddress, operation::external::FaDeposit};
+    use jstz_crypto::smart_function_hash::SmartFunctionHash;
+    use jstz_proto::{
+        context::new_account::{Addressable, NewAddress},
+        operation::external::FaDeposit,
+    };
     use tezos_smart_rollup::{
         michelson::{
             ticket::{FA2_1Ticket, Ticket},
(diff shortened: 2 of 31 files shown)


crates/jstz_cli/src/repl/mod.rs
```diff
@@ -131,7 +131,14 @@ pub async fn exec(account: Option<AddressOrAlias>) -> Result<()> {
     let realm = rt.realm().clone();
 
     register_web_apis(&realm, &mut rt);
-    register_jstz_apis(&realm, &address, DEFAULT_RANDOM_SEED, &mut rt);
+    register_jstz_apis(
```
zcabter, line 134: In this file, line 103 needs to be updated.

crates/jstz_cli/src/repl/mod.rs
zcabter, line 1: In this file, line 103 needs to be updated.

crates/jstz_cli/src/repl/mod.rs
```diff
@@ -131,7 +131,14 @@ pub async fn exec(account: Option<AddressOrAlias>) -> Result<()> {
...
+        &realm,
+        address
```
zcabter, line 136: Probably want to move to check to line 11

crates/jstz_cli/src/repl/mod.rs
ryutamago: good catch

crates/jstz_kernel/src/parsing.rs
```diff
@@ -27,10 +27,14 @@ pub fn try_parse_fa_deposit(
     let receiver = try_parse_contract(&receiver.0)?;
 
     let proxy_smart_function = (proxy_contract)
-        .map(|c| try_parse_contract(&c.0))
+        .map(|c| {
+            let addr: NewAddress = try_parse_contract(&c.0)?;
+            if let NewAddress::SmartFunction(_) = addr {
+                return Ok(addr);
+            }
+            Err(jstz_proto::Error::AddressTypeMismatch)
```
zcabter, line 35:
```suggestion
            if let addr @ NewAddress::SmartFunction(_)  = try_parse_contract(&c.0)? {
                 Ok(addr)
            }
            else {  Err(jstz_proto::Error::AddressTypeMismatch) }
```

crates/jstz_proto/src/api/ledger.rs
```diff
@@ -58,14 +55,20 @@ impl Ledger {
         dst: &NewAddress,
         amount: Amount,
     ) -> Result<()> {
-        Account::transfer(rt, tx, &self.address, dst, amount)?;
+        Account::transfer(
+            rt,
+            tx,
+            &NewAddress::SmartFunction(self.address.clone()),
```
zcabter, line 61: In general, we should avoid cloning in the kernel code but since Ledger API is going to be deprecated, I think we will let this slide.

crates/jstz_proto/src/api/smart_function.rs
```diff
@@ -71,28 +71,34 @@ impl SmartFunction {
         function_code: ParsedCode,
         initial_balance: Amount,
     ) -> Result<String> {
-        let balance = Account::balance(hrt, tx, &self.address)?;
+        let deployer = NewAddress::SmartFunction(self.address.clone());
```
zcabter, line 74: We should avoid clone.

crates/jstz_kernel/src/parsing.rs
ryutamago: didn't know that syntax! neater 💯

crates/jstz_proto/src/api/smart_function.rs
ryutamago: implemented a trait to avoid it!

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -81,6 +81,42 @@ impl TryFrom<String> for ParsedCode {
...
+    fn kind(&self) -> AddressKind {
+        AddressKind::KT1
+    }
+
+    fn to_base58(&self) -> String {
+        <Self as Hash>::to_base58(self)
+    }
+}
+
+pub type UserAddress = PublicKeyHash;
```
zcabter, line 108: Same here, I don't think we need to further alias this type

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -81,6 +81,42 @@ impl TryFrom<String> for ParsedCode {
...
+}
+
+pub type SmartFunctionAddress = SmartFunctionHash;
+
+impl Addressable for SmartFunctionAddress {
+    fn kind(&self) -> AddressKind {
+        AddressKind::KT1
+    }
+
+    fn to_base58(&self) -> String {
```
zcabter, line 103: If you rename this to `address(&self)` then you don't need to add <as Hash> everyhwere

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -403,7 +405,12 @@ impl HostScript {
             tx.begin();
 
             // 2. Execute jstz host smart function
-            let result = jstz_run::execute_without_ticketer(hrt, tx, self_address, run);
+            let result = jstz_run::execute_without_ticketer(
+                hrt,
+                tx,
+                &NewAddress::SmartFunction(self_address.clone()),
```
zcabter, line 411: Hmm I think you need to change the APIs that use `NewAddress` so that they use `Addressable` instead.

crates/jstz_proto/src/context/new_account.rs
```diff
@@ -81,6 +81,42 @@ impl TryFrom<String> for ParsedCode {
     }
 }
 
+pub enum AddressKind {
+    Tz1,
+    KT1,
```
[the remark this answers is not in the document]
ryutamago: i did that at the start, and i thought the trait `Addressable` should just know the format of the address and whether that's a user or sf should be up to the implementor / Account. in case we support differnt prefix type like Tz2 etc wdyt?

crates/jstz_proto/src/context/new_account.rs
ryutamago: nani?!

crates/jstz_proto/src/executor/smart_function.rs
ryutamago: cool i wasn't sure if this is only restricted to sf or not

crates/jstz_proto/src/context/new_account.rs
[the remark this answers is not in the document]
zcabter: In most of jstz, once a valid address has been parsed, you almost would never need to know what the actually address scheme is. I would image that User and SF is much more convenient to work with.

crates/jstz_proto/src/executor/smart_function.rs
zcabter: Sorry my bad, that field is the source, so indeed it is not a just a smart function address.

Why would you need to cast it back to NewAddress?

zcabter: could you try making Addressable convertable into a NewAddress eg. Addressable: Into<NewAddress>?

ryutamago: if we use &impl Addresasble, we would have to change the api of withdraws to take Addressable, but ```pub struct FaWithdrawReceipt {
    pub source: NewAddress,
    pub outbox_message_id: OutboxMessageId,
}```
requires NewAddress to be returned.

zcabter (approved):
LGTM

crates/jstz_proto/src/context/new_account.rs
zcabter: Yeah ok that make sense. Maybe we update it in the future, but not an issue now.

PR #782, opened 2025-01-21
feat(cli): set up cli npm package

huancheng-trili:
# Context

Part of JSTZ-275.
[JSTZ-275](https://linear.app/tezos/issue/JSTZ-275/ship-cli-as-an-npm-package)

# Description

Create an NPM package for jstz CLI.

This package is usable but not in its complete shape. Ideally the package should have `optionalDependencies` that are used to install the actual CLI binary, but those smaller packages make testing a bit more complicated without publishing them to NPM, and publishing these packages needs some more careful planning, so for simplicity, this package uses a rather naive approach before the publish plan gets finalised. Lines about `BINARY_DISTRIBUTION_PACKAGE` are here in preparation for the optional dependencies. They have no effect at the moment.

What happens when a user runs `npm install -g @jstz-dev/cli` or `yarn global add @jstz-dev/cli` is that the package manager fetches the package from the remote repository and then runs `install.js` that actually downloads the CLI binary from the release page of this repository. The actual executable that gets hit when a user runs `jstz` is this `run.mjs` file that spawns the CLI process.

# Manually testing the PR

To test this before releases are created in this repository, I prepared a test repo with a release. A test branch `huanchengchang-jstz-275-4-test` is also created for demonstration before this branch is merged.

```sh
$ yarn global add yarn add 'https://gitpkg.vercel.app/jstz-dev/jstz/packages/cli/main?huanchengchang-jstz-275-4-test'
$ which jstz
$ jstz account list
```
and the CLI should work. Can also try with a linux container:
```sh
$ docker run -it --rm --entrypoint sh node:16-alpine -c "yarn global add 'https://gitpkg.vercel.app/jstz-dev/jstz/packages/cli/main?huanchengchang-jstz-275-4-test' && sh"
```

Tested with node v14, 16, 18, and 23.

### DIFF
--- a/packages/cli/main/install.js
+++ b/packages/cli/main/install.js
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+
+const fs = require("fs");
+const path = require("path");
+const https = require("https");
+
+const BINARY_NAME = "jstz";
+const FALLBACK_BINARY_PATH = path.join(__dirname, BINARY_NAME);
+const BINARY_DISTRIBUTION = {
+  darwin_arm64: "jstz_darwin_arm64",
+  linux_x64: "jstz_linux_x64",
+  linux_arm64: "jstz_linux_arm64",
+};
+const BINARY_DISTRIBUTION_PACKAGE = {
+  darwin_arm64: "cli-darwin-arm64",
+  linux_arm64: "cli-linux-arm64",
+  linux_x64: "cli-linux-x64",
+};
+const PLATFORM_ARCH_KEY = `${process.platform}_${process.arch}`;
+
+function makeRequest(url) {
+  return new Promise((resolve, reject) => {
+    https
+      .get(url, (response) => {
+        if (response.statusCode >= 200 && response.statusCode < 300) {
+          const chunks = [];
+          response.on("data", (chunk) => chunks.push(chunk));
+          response.on("end", () => {
+            resolve(Buffer.concat(chunks));
+          });
+        } else if (
+          response.statusCode >= 300 &&
+          response.statusCode < 400 &&
+          response.headers.location
+        ) {
+          // Follow redirects
+          makeRequest(response.headers.location).then(resolve, reject);
+        } else {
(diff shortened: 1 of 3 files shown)


ryutamago (approved):
lgtm, i would have @zcabter to have a look as well but i will approve it for now :)

packages/cli/main/install.js
zcabter, line 1: why does package.json  need to be in the `main/` directory? Could it just be put in `cli/`?

zcabter, line 1: Could we rename `/main` to `/install`?

zcabter:
>  the package should have optionalDependencies that are used to install the actual CLI binary

What do you mean?

zcabter (approved):
🚀

huancheng-trili:
> the package should have optionalDependencies that are used to install the actual CLI binary

This means that instead of solely relying on the `install.js` script in the main CLI package to identify the platform and architecture and download the corresponding binary, it's better to create additional individual CLI packages specifically for different platforms and architectures, ship the binaries with those packages, and make the main CLI package list them as optional dependencies. When the main CLI package gets installed, the proper optional dependency that matches the platform and architecture gets installed as well because the package manager can identify the platform and architecture automatically ([ref](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#os)) and therefore the proper binary is downloaded without us having to code out this logic.

For instance, we can have these in the main package.json:
```
optionalDependencies: {
  "cli-darwin-arm64": "@jstz-dev/cli-darwin-arm64",
  "cli-linux-arm64": "@jstz-dev/cli-linux-arm64",
  ...
}
```
where the `cli-...` packages includes nothing but the binaries. The binaries get downloaded ("installed") when the dependencies get installed along with the main package. I've tested this and it works. This is also why this package is now in the `main` directory. I plan to generate and publish these packages in the pipeline only so that we don't need to check in these binaries, and I need to separate the main package with the generated ones.

PR #783, opened 2025-01-23
fix(cli): add `container` flag to sandbox config

huancheng-trili:
# Context

Part of JSTZ-274.
[JSTZ-274](https://linear.app/tezos/issue/JSTZ-274/run-jstzd-with-jstz-sandbox-start).

# Description

Since now jstzd can also run in a container, we need to identify such cases to use the fixed address for the bridge contract rather than using the alias. To achieve this, a flag `container` is added to the sandbox config. This flag is set only when the sandbox container is in use.

# Manually testing the PR

Tested this manually.


ryutamago (approved):
🚀

PR #784, opened 2025-01-23
feat(cli): call contract from jstzd server

huancheng-trili:
# Context

Part of JSTZ-273.

# Description

Jstz CLI makes use of octez client. Having to ship the octez binary along with CLI is quite frustrating because the whole point of the sandbox container is to free users from having to install octez binaries. The solution for now is to expose the octez client inside the sandbox container to some extent through jstzd server so that CLI can still do its job without an actual octez client binary nearby. Since the only usage of the octez client is to transfer tez, one new endpoint `POST /contract_call` is added. It performs `octez-client transfer` when it's called.

# Manually testing the PR

Tested this manually by building a jstzd image and the CLI locally.


crates/jstz_cli/src/config.rs
```diff
@@ -539,16 +539,22 @@ impl Config {
...
-                        SANDBOX_LOCAL_HOST_ADDR, SANDBOX_OCTEZ_NODE_RPC_PORT
-                    ),
-                    jstz_node_endpoint: format!(
-                        "http://{}:{}",
-                        SANDBOX_LOCAL_HOST_ADDR, SANDBOX_JSTZ_NODE_PORT,
-                    ),
-                }),
+                None => {
+                    let addr = match self.sandbox().is_ok_and(|c| c.container) {
+                        true => "localhost",
```
ryutamago, line 544: let's leave a //TODO here indicating that this is a hack and we can come back to it later to solve it

ryutamago (approved):
LGTM, just one nit
could you share the commands you ran to test manually in the PR descritipn as well?

PR #788, opened 2025-01-24
fix(jstzd): remove job limit in docker build

huancheng-trili:
# Context

Closes JSTZ-286.
[JSTZ-286](https://linear.app/tezos/issue/JSTZ-286/build-multiplatform-images)

# Description

Remove the job limit in building docker images for jstzd. The config value `--jobs` was set to prevent the system from using too much resource during cross-compilation, but now there is no need to do cross-compilation, this config value can be removed to speed up the build process.

# Manually testing the PR

N/A


ryutamago (approved):
NICE🛳️

zcabter (approved):

PR #802, opened 2025-02-05
fix(jstz_engine): introduce generative compartment `Id`s

johnyob:
# Context

 **Related Tasks**: [jstz-298](https://linear.app/tezos/issue/JSTZ-298/fix-generative-of-lifetime-parameters-for-compartments) 

While playing around with unrooting this morning, Ryan and I observed the following bug -- namely the following snippet compiles:
```rust
// Initialize JS engine
let engine = JSEngine::init().unwrap();
let rt = Runtime::new(engine.handle());

let rt_cx1 = Runtime::new(engine.handle());
let cx1 = rt_cx1.new_realm().unwrap();

let rt_cx2 = Runtime::new(engine.handle());
let cx2 = rt_cx2.new_realm().unwrap();

// This line shouldn't compile
let foo = vec![cx1, cx2];
```

Morally speaking, each `new_realm` should allocate a 'generative' lifetime parameter -- this parameter is generative in the sense that no other lifetime should unify with it apart from itself. 

# Description

The fix is the following:
1. Make the `'a` parameter of `compartment::Ref` invariant -- meaning that it is never valid to substitute or `Ref<'a>` into `Ref<'b>`, for *any* concrete `'a` or `'b`. 
2. Add a `Region<'a>` as a compiler witness that `'a` is live until `Region` is dropped
3. Add a macro `alloc_compartment` which allocates a fresh compartment reference in a fresh region -- APIs that use this compartment reference (as a value) should use the reference as an owned value. 

# Manually testing the PR

Added some tests to check the generativity of compartment ids
```
```

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,6 @@
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 1 }
+
+[[profile.default.overrides]]
+filter = 'package(jstz_engine) and test(ui)'
+slow-timeout = { period = "3m", terminate-after = 1 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2856,6 +2856,7 @@ version = "0.1.0-alpha.0"
 dependencies = [
  "anyhow",
  "mozjs",
+ "trybuild",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


johnyob:
This fix also has the nice benefit of removing an existing (and unsatisfying) `unsafe` block in `Context::new_realm`

zcabter:
LGTM 🚀

zcabter (approved):
LGTM

PR #806, opened 2025-02-07
fix(wpt): handle utf-16 characters

huancheng-trili:
# Context

Closes JSTZ-304.
[JSTZ-304](https://linear.app/tezos/issue/JSTZ-304/handle-utf-16-strings)

# Description

Handle JsStrings with utf-16 characters properly in the test runner. Some test cases involve utf-16 characters, and the implementation of `TryFromJs for String` does not handle those characters properly, specifically the method [`to_std_string`](https://docs.rs/boa_engine/0.19.0/src/boa_engine/value/conversions/try_from_js.rs.html#39-52), which leads to errors like

`could not convert JsString to Rust string: invalid utf-16: lone surrogate found`

and leaves the runner unable to handle some tests.

To fix this, we need to get `JsString` first and then do the proper conversion ourselves.

# Manually testing the PR

```
```

Test cases showed up for the test suite `encodeInto.any.js` and there is no more error.

### DIFF
--- a/crates/jstz_api/tests/wpt.rs
+++ b/crates/jstz_api/tests/wpt.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use boa_engine::{
     js_string, object::FunctionObjectBuilder, property::PropertyDescriptor,
     value::TryFromJs, Context, JsArgs, JsData, JsNativeError, JsObject, JsResult,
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
nice!

PR #807, opened 2025-02-07
feat(wpt): enable blob tests

huancheng-trili:
# Context

Part of JSTZ-303.

# Description

Enable test suites for Blob ([interface](https://w3c.github.io/FileAPI/#dfn-Blob), [class](https://nodejs.org/docs/v22.13.1/api/buffer.html#class-blob)).

# Manually testing the PR

```sh
```


ryutamago (approved):
🛳️

PR #808, opened 2025-02-07
feat(wpt): enable tests for queuing strategies

huancheng-trili:
# Context

Part of JSTZ-303.
[JSTZ-303](https://linear.app/tezos/issue/JSTZ-303/record-all-relevant-wpt-test-suites)

# Description

Enable test suites for `CountQueuingStrategy` ([interface](https://streams.spec.whatwg.org/#countqueuingstrategy), [class](https://nodejs.org/docs/v22.13.1/api/globals.html#class-countqueuingstrategy)) and `ByteLengthQueuingStrategy` ([interface](https://streams.spec.whatwg.org/#bytelengthqueuingstrategy), [class](https://nodejs.org/docs/v22.13.1/api/globals.html#class-bytelengthqueuingstrategy)).

# Manually testing the PR

```sh
env UPDATE_EXPECT=1 cargo test --test wpt
```

PR #809, opened 2025-02-07
ci: skip wpt but keep test coverage for jstz_api

huancheng-trili:
# Context

Part of JSTZ-303.

# Description

Skip wpt in CI in a different way so that test coverage for jstz_api can still be calculated properly.

I'm making some code changes to jstz_api in #808, but codecov complains that the test coverage is 0 even if I created a test for the changed lines. The problem lies in this `--exclude-from-test` flag. Since the goal is to prevent llvm-cov from running wpt, I replaced that flag with a feature flag in jstz_api that ignores wpt so that other tests in the crate can contribute to test coverage.

# Manually testing the PR

CI runs properly.

Test coverage indeed changed with the [test commit](https://app.codecov.io/github/jstz-dev/jstz/commit/4fc0af07209047e5e7afc5b650fffc8dde71d702/indirect-changes).

### DIFF
--- a/crates/jstz_api/Cargo.toml
+++ b/crates/jstz_api/Cargo.toml
@@ -36,3 +36,6 @@ expect-test.workspace = true
 jstz_wpt = { path = "../jstz_wpt" }
 tezos-smart-rollup-mock.workspace = true
 tokio.workspace = true
+
+[features]
+skip-wpt = []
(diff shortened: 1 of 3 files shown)


ryutamago (approved):
🟢

PR #811, opened 2025-02-10
feat(jstz_api): implement todo to return JsError

huancheng-trili:
# Context

Part of JSTZ-303.

# Description

Implement `todo!` to return JsError for methods that are not yet implemented.

When methods calling `std::todo!` get hit, they simply panic, and this breaks wpt. To ensure that tests fail gracefully, these methods need to return JsError with some error messages. I think overall todos in this crate should return errors anyway rather than panicking since panicking means users will never be able to catch and report such errors.

# Manually testing the PR

```sh
$ cd crates/jstz_api && cargo test todo
running 1 test
test todo::tests::todo ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

### DIFF
--- a/crates/jstz_api/src/todo.rs
+++ b/crates/jstz_api/src/todo.rs
@@ -10,25 +10,54 @@ pub enum Todo {
 
 impl Finalize for Todo {
     fn finalize(&self) {
-        todo!()
+        std::todo!()
     }
 }
 
 #[allow(unused_variables)]
 unsafe impl Trace for Todo {
-    custom_trace!(this, mark, todo!());
+    custom_trace!(this, mark, std::todo!());
 }
 
 #[allow(unused_variables)]
 impl IntoJs for Todo {
     fn into_js(self, context: &mut Context) -> JsValue {
-        todo!()
+        std::todo!()
     }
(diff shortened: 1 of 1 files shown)


crates/jstz_api/src/todo.rs
```diff
@@ -10,25 +10,54 @@ pub enum Todo {
...
-        todo!()
+        std::todo!()
     }
 }
 
 #[allow(unused_variables)]
 impl TryFromJs for Todo {
     fn try_from_js(value: &JsValue, context: &mut Context) -> JsResult<Self> {
-        todo!()
+        std::todo!()
```
ryutamago, line 32: aren't we sujpposed to use the macro below to reutrn JsResultErr here?

huancheng-trili: Not for now. It's true that we can use the new `todo` macro here as well, but I want to keep the changes simple for now and leave this `Todo` struct as is so that we don't get other unexpected changes.

ryutamago (approved):

PR #812, opened 2025-02-10
feat(jstz_engine): support compile native kernel

zcabter:
# Context
Adds support for compile native kernel for jstz engine. 

This bundles jstz engine as an inputless kernel that runs natively, allowing trace of syscalls with strace (linux) or dtrace or ktrace (mac). 

# Description

# Manually testing the PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -910,6 +910,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-utils"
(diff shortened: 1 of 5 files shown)


huancheng-trili (commented):
Why do we need this?

zcabter:
> What will this be used for?
Syscall trace

huancheng-trili (approved):

PR #813, opened 2025-02-11
Leounoki jstz 300/request noop

ryutamago:
# Context

[task link](https://linear.app/tezos/issue/JSTZ-300/add-support-for-no-op)

This PR decouples the transfer and the execturion of sf via the `/-/noop` url path.
`run tezos://KT1.../-/noop` now enables just handling the transfer (if any) without executing the sf

# Description

check the URL path and skip the execution if the `noop` is in the path.

# Manually testing the PR
added unit tests to check the noop  behaviour
cargo test -p jstz-proto

### DIFF
--- a/crates/jstz_proto/src/api/smart_function.rs
+++ b/crates/jstz_proto/src/api/smart_function.rs
@@ -18,7 +18,7 @@ use jstz_crypto::smart_function_hash::SmartFunctionHash;
 use crate::{
     context::account::{Account, Address, Amount, ParsedCode},
     executor::{
-        smart_function::{headers, HostScript, Script},
+        smart_function::{headers, run::NOOP_PATH, HostScript, Script},
         JSTZ_HOST,
     },
     operation::OperationHash,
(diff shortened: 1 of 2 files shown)


huancheng-trili:
Where does this `/-/noop` idea come from? I'm just not very sure about doing this through a path. I think it should be fine in most cases, but this means users won't be able to handle this path in smart functions and having some "reserved paths" seems a bit weird.

ryutamago:
@huancheng-trili 
>users won't be able to handle this path in smart functions and having some "reserved paths" seems a bit weird.
i think it's fine to have a path reserved and decide the rules ourselves similar to how it's done with the transfer headers. wdyt? @zcabter

huancheng-trili:
Transfer header is a different thing. It has the `X-` prefix, which marks it as an application-specific header, and therefore it would be commonly respected by users.

zcabter:
We need reserved paths for a few special behaviours like this and the callback for fa-deposits. I think reserving `/-/` is ok and has the benefit of being 1 character but open for suggestions

huancheng-trili:
If you both think it's okay and you want to proceed with it, you should deal with the review yourselves. I don't have a better solution at the moment, but I don't fully agree with this and therefore I don't think I'm the right person to review it.

crates/jstz_proto/src/api/smart_function.rs
```diff
@@ -27,6 +27,9 @@ use crate::{
 
 use boa_gc::{empty_trace, Finalize, GcRefMut, Trace};
 
+/// skips the execution of the smart function for noop requests
+pub const NOOP_PATH: &str = "/-/noop";
+
```
zcabter, line 32: Nit: Move to `executor::smart_function`

zcabter (approved):

PR #821, opened 2025-02-14
feat(jstz_engine): add `JsValue` conversion traits

zcabter:
# Context
Adds conversion traits to `JsValue`. These traits provides the capability to convert between JsValue and Rust types. This will be useful when implement JsObject in the short term, and make it easy to interop with Js types overall. 

Closes: [JSTZ-211](https://linear.app/tezos/issue/JSTZ-211/implement-jsvalue), [JSTZ-224](https://linear.app/tezos/issue/JSTZ-224/implement-tryfromjs-for-strings),

# Description
* Add `trait TryIntoJs` - Fallible conversion between Rust native types and JsValue
* Add `trait TryFromJs` - Fallible conversion between JsValue and Rust native/JsTypes
* Implements runtime checkers for JsTypes

# Manually testing the PR
`cargo nextest run -p jstz_engine value`


crates/jstz_engine/src/gc/root.rs
```diff
@@ -262,6 +262,8 @@ impl<'a, T: Trace> Rooted<'a, T> {
         unsafe { self.as_raw_handle_mut() }
     }
 
+    // BUG: This breaks the invariant of popping things in the stack --
+    // stack needs to become doubly-linked
```
johnyob, line 266:
```
// FIXME (https://linear.app/tezos/issue/JSTZ-317): Explicitly unrooting in a different order to which 
// values were rooted in results in use-after-free errors!
```

crates/jstz_engine/src/value/conversions/try_from_js.rs
```diff
@@ -0,0 +1,129 @@
...
+        cx: &'cx mut Context<S>,
+    ) -> Option<Self::Output>
+    where
+        S: InCompartment<C> + CanAccess + CanAlloc;
+}
+
+impl<'a, 'cx: 'a, C: Compartment> TryFromJs<'a, 'cx, C> for i32 {
+    type Output = i32;
+
+    type Config = ();
```
johnyob, line 31: mozjs permits us to have more than 1 kind of conversion behaviour for `i32`s (and other int types) -- namely:
underflows/overflows, enforcing range (throwing an exception if not in range), and clamping. 

We should probably support these behaviours (since there are many other ways to obtain an `i32` from `JsValue` than just the `Number` variant in `JsValue`)

crates/jstz_engine/src/value/conversions/try_from_js.rs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        if !value.is_i32() {
+            return None;
+        }
+        Some(unsafe { value.as_raw_ptr().to_int32() })
+    }
+}
+
+impl<'a, 'cx: 'a, C: Compartment> TryFromJs<'a, 'cx, C> for f64 {
+    type Output = f64;
```
johnyob, line 49: For `f64`, we should favour `ToNumber`. Perhaps having some 'strict' (i.e. only convert if actually a f64) behaviour that we can enforce with `Config`

crates/jstz_engine/src/value/conversions/try_from_js.rs
```diff
@@ -0,0 +1,129 @@
...
+    where
+        S: InCompartment<C> + CanAccess + CanAlloc,
+    {
+        if !value.is_string() {
+            return None;
+        }
+        let js_string: JsString<'a, C> =
+            unsafe { JsString::from_raw(value.as_raw_ptr().to_string()) };
+        letroot!(rooted = js_string; [cx]);
+        rooted.to_std_string(cx).ok()
```
johnyob, line 87: Morally the underlying value we obtain here should already be rooted -- this `letroot!` is unnecessary

crates/jstz_engine/src/value/conversions/try_into_js.rs
```diff
@@ -0,0 +1,40 @@
...
+        S: InCompartment<C> + CanAccess + CanAlloc;
+}
+
+impl<'cx, C: Compartment> TryIntoJs<'cx, C> for &str {
+    fn try_into_js<S>(&self, cx: &'cx mut Context<S>) -> Option<JsValue<'cx, C>>
+    where
+        S: InCompartment<C> + CanAccess + CanAlloc,
+    {
+        let js_string = JsString::new(self, cx);
+        Some(unsafe { JsValue::from_raw(StringValue(&*js_string.as_raw_ptr())) })
```
johnyob, line 24: Why not add `JsValue::string(str: JsString<'a, C>, cx: &'cx mut Context<S>)`?

zcabter:
Closing due to hard pivot to Deno https://github.com/jstz-dev/jstz/pull/846

PR #836, opened 2025-02-25
feat(examples): example of calling from a web application

timothymcmackin:
# Context

We have examples of calling smart functions from a CLI but our target market is web developers, so it would be good to have an example of calling Jstz from a web application.

# Description

This is a Vite web app that calls a smart function if you fill in info about the smart function and your account info. I've included a note in the readme about how this is not production ready because you have to put your key in plaintext and there's no signer yet.

# Manually testing the PR

Try the instructions in the readme.


examples/call-from-web/README.md
```diff
@@ -0,0 +1,24 @@
+# Call a Jstz smart function from a web application
```
huancheng-trili, line 1: Not meant to be part of this PR but I guess we need to show these readme files in the website as well. Now it seems that users can only find them in the repo.

examples/call-from-web/README.md
```diff
@@ -0,0 +1,24 @@
...
+This is a [Vite](https://vite.dev/) application that shows how you can call a Jstz smart function from a web application.
+It lets you enter information about your smart function and account, call the smart function, and see the response from the smart function.
+
+**This application does not secure private keys and therefore should not be used in production.**
+
+Follow these steps to use it:
+
+1. Start the Jstz sandbox, create a Jstz account, and deploy a smart function as described in the Jstz [Quick start](https://jstz-dev.github.io/jstz/quick_start.html).
+
+2. Clone this repository and run `npm i`.
```
huancheng-trili, line 12: I think it would be better if we make it clear that `npm i` should be run at the root directory of this repository. Initially I ran it in this example directory and got some errors regarding missing dependencies from `npm run dev`. Even better, we can give some sample commands so that users can simply copy and paste.

examples/call-from-web/README.md
```diff
@@ -0,0 +1,24 @@
+# Call a Jstz smart function from a web application
+
+This is a [Vite](https://vite.dev/) application that shows how you can call a Jstz smart function from a web application.
+It lets you enter information about your smart function and account, call the smart function, and see the response from the smart function.
+
+**This application does not secure private keys and therefore should not be used in production.**
+
+Follow these steps to use it:
+
+1. Start the Jstz sandbox, create a Jstz account, and deploy a smart function as described in the Jstz [Quick start](https://jstz-dev.github.io/jstz/quick_start.html).
```
huancheng-trili, line 10: I'm not sure if this is what you expect, but I feel that you kind of assume that users will indeed deploy the smart function in the quick start guide. It's fine, but if they deploy some other toy functions, the page should be able to handle them as well. I deployed a dummy function and the page could not display the response because the response wasn't in the expected format. See the other comment.

examples/call-from-web/src/main.jsx
```diff
@@ -0,0 +1,146 @@
...
+    const jstzClient = new Jstz({
+      timeout: 6000,
+    });
+
+    const runFunctionRequest = buildRequest(
+      smartFunctionAddress,
+      message,
+      smartFunctionPath,
+    );
+    const nonce = await jstzClient.accounts.getNonce(accountAddress);
```
huancheng-trili, line 52: One important thing here is that the account must be "in the records" already. I created a new account with `jstz create` and then used the address directly and the page failed here because the nonce of this account didn't exist. I had to make a transfer to make this work. This won't be a problem if users do follow the quick start guide first. It's kind of difficult to explain to users that they need to sort of "materialise" accounts first instead of only creating them with jstz cli, but perhaps we need to put this as a prerequisite in the readme.

examples/call-from-web/src/main.jsx
```diff
@@ -0,0 +1,146 @@
...
+      signature: signature,
+    });
+    const {
+      result: {
+        inner: { body },
+      },
+    } = await response;
+
+    const returnedMessage = body
+      ? JSON.parse(decoder.decode(new Uint8Array(body)))
```
huancheng-trili, line 74: This assumes that the body is a byte array. I guess this is also fine with the quick start guide. I'm just wondering if we should instead display the body itself without processing it because the body can be anything. My toy smart function was just
```js
export default (() => new Response('hello world'))
```
and the page failed to handle it, which was a surprise to me at that time because I expected the page to simply display "hello world".

examples/call-from-web/README.md
timothymcmackin: One option is to create a repo of examples like https://github.com/trilitech/tutorial-applications. I could also create a doc page that describes each example and links to it.

examples/call-from-web/README.md
timothymcmackin: Clarified.

examples/call-from-web/README.md
timothymcmackin: Yes, I think to make this as simple as possible the web app should call one function. I added a "counter" smart function and updated the web application to call that function.

examples/call-from-web/src/main.jsx
timothymcmackin: Got it, thanks for the info. Is there a technical term like "revealed" that we can use, like we use "revealed" for layer 1 accounts? I updated the instructions and added error-checking to the app.

huancheng-trili: Yes, "revealed" sounds right here.

timothymcmackin: Updated to "revealed."

examples/call-from-web/README.md
timothymcmackin: I see now. I've got the Counter example building and I added some dependencies to call-from-web but when I copy it outside of the jstz folder and run `npm i; npm run build` I get this error. What dependency is missing here?

```
error during build:
[vite:esbuild-transpile] Transform failed with 1 error:
assets/index-!~{001}~.js:18199:27: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
file: assets/index-!~{001}~.js:18199:27

Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
18197|
18198|  URL = globalThis.URL;
18199|  const __vite__wasmModule = await __vite__initWasm({ "./jstz_sdk_bg.js": { "__wbindgen_object_drop_ref": __wbindgen_object_drop_ref,
   |                             ^
18200|  "__wbindgen_boolean_get": __wbindgen_boolean_get,
18201|  "__wbindgen_is_bigint": __wbindgen_is_bigint,
```

huancheng-trili: Adding this to the vite config should work:
```
  build: {
    target: 'esnext'
  }
```
[ref](https://stackoverflow.com/a/75839023)

huancheng-trili (approved):

PR #837, opened 2025-02-27
chore: benchmark kernel

huancheng-trili:
# Context

Completes JSTZ-30.
[JSTZ-30](https://linear.app/tezos/issue/JSTZ-30)

# Description

Took some code for benchmarking jstz_kernel from [tezos](https://gitlab.com/tezos/tezos/-/tree/master/src/riscv/jstz).

# Manually testing the PR

```sh
# build the crate first to get the bench binary to generate an inbox file
cargo build -p jstz_tps_bench
./target/debug/bench generate --transfers 10 --inbox-file ./crates/jstz_tps_bench/src/kernel/inbox.json
# build the crate again with the feature flag so that the kernel executable compiles with the inbox file
cargo build -p jstz_tps_bench --features static-inbox
./target/debug/kernel --timings > /tmp/benchmark-log
# show the result
./target/debug/bench results --inbox-file ./crates/jstz_tps_bench/src/kernel/inbox.json --log-file /tmp/benchmark-log --expected-transfers 10
```


crates/jstz_bench/Cargo.lock
johnyob, line 1: Shouldn't we be using the root `Cargo.lock` file here?

crates/jstz_bench/fa2.js
johnyob, line 1: We should probably comment that this minified smart function is from our fa2 example, built using: https://gitlab.com/tezos/tezos/-/blob/master/src/riscv/scripts/get-fa2.sh

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 resolver = "2"
 members = [
   "crates/jstz_api",
+  "crates/jstz_bench",
```
johnyob, line 5: super nit: but maybe suggest `jstz_tx_bench` or `jstz_tps_bench` -- since this crate isn't a general benchmarking crate (and won't likely evolve into one), but one for a specific scenario (FA2 transfers)

johnyob (approved):
Overall lgtm 🚀 Just some minor comments / nits

Cargo.toml
huancheng-trili: I changed this to `jstz_tps_bench`.

crates/jstz_bench/fa2.js
huancheng-trili: I added a readme in the crate with some instructions and this note.

PR #838, opened 2025-02-27
feat(docs): basic topic on accounts

timothymcmackin:
This is a pretty basic topic but I thought it was needed. Just basic info on accounts that we can add to as the functionality expands.

### DIFF
--- a/docs/accounts.md
+++ b/docs/accounts.md
@@ -1,0 +1,81 @@
+# Accounts
+
+Jstz works with two kinds of accounts:
+
+- User accounts start with `tz1` and store tez (XTZ).
+
+- Smart function accounts start with `KT1` and store tez and the immutable code of the smart function.
+
+## Working with user accounts
+
+User accounts are stored in the local file `~/.jstz/config.json`, including the alias, address, public key, and secret key for each account.
+
+::: warning
+
+You must keep the secret keys for the accounts secure.
+Jstz does not yet support storing keys in wallets, on hardware devices, or with remote signers.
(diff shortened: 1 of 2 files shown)


docs/accounts.md
```diff
@@ -0,0 +1,80 @@
...
+For example to get the code of a smart function in the local sandbox, run this command, with the address or alias as `<ALIAS_OR_ADDRESS>`:
+
+```bash
+jstz account -a <ALIAS_OR_ADDRESS> -n dev
+```
+
+To call a smart function from the command line, use the `jstz run` command and pass the URL to the smart function as `<ADDRESS>`, as in this example from the [Quick start](/quick_start):
+
+```bash
+jstz run tezos://<ADDRESS>/ --data '{"message":"Give me tez now."}' -n dev
```
huancheng-trili, line 77: Since you bring up smart functions here, I think it's worth mentioning that users can also call smart functions by alias.

```sh
# Quick aliasing
jstz deploy --name <ALIAS> <source file>

# Aliasing after deployment
jstz account alias <ALIAS> <KT1 address>

# Calling function
jstz run tezos://<ALIAS>/
```

But calling by alias only works with the CLI tool. It does not work with the JS SDK.

huancheng-trili (approved):
Just one comment

docs/accounts.md
timothymcmackin: Ah, good, I added that with a note that it works only with the CLI client and not the SDK.

PR #840, opened 2025-02-28
fix(crypto): check length of smart function addresses

huancheng-trili:
# Context

Part of JSTZ-338.
[JSTZ-338](https://linear.app/tezos/issue/JSTZ-338/jstz-account-alias-panics-when-address-is-too-short)

# Description

Check the length of input strings in `from_base58` to avoid panics.

This bug leads to something like this:
```
$ jstz account alias a b                      
thread 'main' panicked at crates/jstz_crypto/src/smart_function_hash.rs:60:20:
byte index 3 is out of bounds of b
```

# Manually testing the PR

Updated the corresponding unit test.


ryutamago (approved):
🙏

PR #842, opened 2025-03-03
ci: benchmark after build

huancheng-trili:
# Context

Part of JSTZ-31.
[JSTZ-31](https://linear.app/tezos/issue/JSTZ-31)

# Description

Runs benchmarking as part of each build. If a build is for a PR, benchmarking will be run against the base branch as well for comparison. If benchmarking does not work in the base branch, it will be skipped. Of course it's possible to cache the result somewhere so that we don't need to run it for the base branch every time, but since it requires setting up a database and nix cache should help avoid building the base branch from scratch, I think it's okay to leave it like this. This result display can be further extended to something like attaching a comment to a PR so that people are easily aware of changes that lead to performance regression.

`jstz_bench` in flake.nix looks a bit awkward because of how the benchmark program needs to be built. I want to keep parameters, e.g. number of smart function calls, all in the same place. With this implementation, everything is wrapped inside `run.sh` and that looks better to me.

Note that [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) is not used here because it's not supposed to be used for PRs. We can use that for the main branch, but that's not the point of this PR.

# Manually testing the PR

The [workflow run](https://github.com/jstz-dev/jstz/actions/runs/13628594079/job/38091300947?pr=842) succeeded.
```
Current branch: 10 FA2 transfers took 49.467928ms @ 202.151 TPS
```

### DIFF
--- a/nix/crates.nix
+++ b/nix/crates.nix
@@ -89,7 +89,7 @@
   #     });
 
   workspace = craneLib.cargoBuild commonWorkspace;
-in {
+in let
   packages = {
     # A list of all the crates in the workspace
     # When adding a new crate, add it to this list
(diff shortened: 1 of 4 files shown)


nix/crates.nix
```diff
@@ -93,6 +93,21 @@ in {
     # When adding a new crate, add it to this list
     # in alphabetical order.
     jstz_api = crate "jstz_api";
+    jstz_bench = craneLib.buildPackage (commonWorkspace
+      // rec {
+        # build the crate first to get the bench binary to generate an inbox file
+        benchmark_cli = crate "jstz_bench";
```
johnyob, line 99: A better way to do this would be to split this into one (or two) nix flake packages and a nix flake app (aka your run.sh). You can then use `nix run tps_bench` in the CI

huancheng-trili: Finally made it work with `nix run .#tps_bench`!

johnyob: It's not how I would've done -- I think it would've been cleaner to:
1. make the `jstz_bench` crate configurable by feature + add a parameter for the static inbox. 
2. Create a third derivation for the static inbox content (which would use `passthru` to pass the `transfer_count` parameter)
3. Used `apps` to run the command you write to `run.sh`

However, I don't want to block this PR any further (and tbh the build process is messy, so its not surprising that the Nix code reflects this a bit). So I'll approve the changes 😃

johnyob (approved):

PR #845, opened 2025-03-03
chore(nix): update nixpkgs and octez

zcabter:
# Context
Required for [cross compiling Deno](https://github.com/jstz-dev/jstz/pull/846) to `riscv64-unknown-musl-linux` 

Deno was failing to cross compile within the jstz project (outside of jstz was fine, even with the same nix riscv toolchain) due to MacOS flags always being included in CFLAG for cross compile builds and missing iconv dependency that should have been bundled with musl. After some painful debugging, I discussed the issue with Ole and we came to a realization that the outdated nixpkgs could be the issue.

# Description
This PR updates the `nixpkgs` which in turn required an `octez` upgrade. 

When I tried updating nixpkgs, octez-21 started to break so I went ahead an upgraded octez and all other flake dependencies as well.

# Manually testing the PR
`nix --accept-flake-config --log-format raw --log-lines 70 -L flake check -j auto` + Green CI


flake.nix
```diff
@@ -91,20 +91,47 @@
...
+                };
+              in ''
+                mkdir -p ${dir}/.cargo
+                cat >> ${dir}/.cargo/config.toml << EOF
+                [source.crates-io]
+                replace-with = "vendored-sources"
+
+                [source.github]
+                replace-with = "vendored-sources"
+
```
zcabter, line 112: Remove this

PR #849, opened 2025-03-04
feat(jstz_proto): implement refund in response

ryutamago:
# Context

[task](https://linear.app/tezos/issue/JSTZ-330/handle-response-refund-for-fetch)

# Description

handle refund in response.

# Manually testing the PR

* unit test

### DIFF
--- a/crates/jstz_proto/src/executor/smart_function.rs
+++ b/crates/jstz_proto/src/executor/smart_function.rs
@@ -62,12 +62,16 @@ pub mod headers {
 
 // Applies on_fullfilled or on_rejected based on either an error was raised or not.
 // If the value is a promise, then we apply the on_fulfilled and on_rejected to the promise.
-fn try_apply_to_value_or_promise(
+pub fn try_apply_to_value_or_promise<F1, F2>(
     value_or_promise: JsResult<JsValue>,
-    on_fulfilled: fn(&JsValue, &mut Context) -> JsResult<()>,
-    on_rejected: fn(&mut Context) -> JsResult<()>,
+    on_fulfilled: F1,
+    on_rejected: F2,
     context: &mut Context,
-) -> JsResult<JsValue> {
+) -> JsResult<JsValue>
+where
+    F1: Fn(&JsValue, &mut Context) -> JsResult<()> + 'static,
+    F2: Fn(&mut Context) -> JsResult<()> + 'static,
+{
     match value_or_promise {
         Ok(value) => match value.as_promise() {
             Some(promise) => {
@@ -367,6 +371,7 @@ impl Script {
 }
 
 pub const TRANSFER_HEADER_KEY: &str = "X-JSTZ-TRANSFER";
+pub const TRANSFERRED_HEADER_KEY: &str = "X-JSTZ-TRANSFERRED";
 
 pub struct HostScript;
 
@@ -476,7 +481,7 @@ impl HostScript {
         headers: &impl Deref<Target = Headers>,
         src: &impl Addressable,
         dst: &impl Addressable,
-    ) -> JsResult<()> {
+    ) -> JsResult<Option<NonZeroU64>> {
         if let Some(amt) = Self::extract_transfer_amount(headers)? {
             let transfer_result = runtime::with_js_hrt_and_tx(|hrt, tx| {
                 Account::transfer(hrt, tx, src, dst, amt.into()).map_err(|e| {
@@ -487,10 +492,10 @@ impl HostScript {
                 })
             });
 
-            return transfer_result;
+            return transfer_result.map(|_| Some(amt));
         }
 
-        Ok(())
+        Ok(None)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -470,27 +475,32 @@ impl HostScript {
...
+    ) -> JsResult<Option<NonZeroU64>> {
+        let amt = match Self::extract_transfer_amount(headers)? {
+            Some(a) => a,
+            None => return Ok(None),
+        };
+        runtime::with_js_hrt_and_tx(|hrt, tx| {
+            Account::transfer(hrt, tx, src, dst, amt.into())
+                .and_then(|_| {
+                    headers.remove(TRANSFER_HEADER_KEY)?;
+                    headers.append(TRANSFERRED_HEADER_KEY, &amt.to_string())?;
```
huancheng-trili, line 493: Should we make sure that this key cannot be set by any smart function? Otherwise, this can be abused in scams, e.g. returning a fake refund header and fooling downstream smart functions that depend on it.

ryutamago: good catch thanks, reject any heaaders that include amount

ryutamago:
@zcabter @huancheng-trili 
For now we only have a `transfer` logic so i placed all the relevant logic (including verifying the headers) inside the `HostScript::handle_transfer` 
However I think it's better to have a separate sf `hook` module that gets called both pre and post sf call if we are starting to add more logic in the future.

crates/jstz_cli/src/main.rs
```diff
@@ -78,6 +78,9 @@ enum Command {
         /// The JSON data in the request body.
         #[arg(name = "data", short, long, default_value = None, value_hint = clap::ValueHint::FilePath)]
         json_data: Option<String>,
+        /// The amount in XTZ to transfer.
+        #[arg(short, long, default_value_t = 0)]
+        amount: u64,
```
zcabter, line 83: Should this be `Option<u64>`?

crates/jstz_cli/src/run.rs
```diff
@@ -64,11 +64,12 @@ pub struct RunArgs {
     url: String,
     http_method: String,
     gas_limit: u32,
+    /// The amount in XTZ to transfer.
+    amount: u64,
```
zcabter, line 68: Same here, `Option<u64>` indicates that running a smart function doesn't require a transfer.

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -470,27 +475,43 @@ impl HostScript {
         Ok(amount)
     }
 
-    /// Transfer xtz from `src` to `dst` if the `TRANSFER_HEADER_KEY` header is present
-    /// and the transfer amount > 0 in request
+    fn verify_headers(headers: &impl Deref<Target = Headers>) -> JsResult<()> {
```
zcabter, line 478: Does `&Headers` not work as in argument?

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -470,27 +475,43 @@ impl HostScript {
...
+        if headers.contains_key(X_JSTZ_AMOUNT) {
+            return Err(JsError::from_native(
+                JsNativeError::typ().with_message("Invalid header X-JSTZ-AMOUNT"),
```
zcabter, line 481:
```suggestion
                JsNativeError::typ().with_message("X-JSTZ-AMOUNT header should not be present"),
```

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -470,27 +475,43 @@ impl HostScript {
...
+            ));
+        }
+        Ok(())
+    }
+
+    /// Transfer xtz from `src` to `dst` if the `X_JSTZ_TRANSFER` header is present & amount > 0
+    /// On success, `X_JSTZ_TRANSFER` is set to `X_JSTZ_AMOUNT`
+    /// Rejects if `X_JSTZ_AMOUNT` is already present in the headers
```
zcabter, line 489:
```suggestion
    /// Rejects if `X_JSTZ_AMOUNT` is already present in the headers or transfer failed
```

crates/jstz_proto/src/executor/smart_function.rs
huancheng-trili:
```suggestion
                JsNativeError::error().with_message("Invalid header X-JSTZ-AMOUNT"),
```

I don't think this should be a type error and there's nothing like `ValueError` in JavaScript, so a general error is probably better.

crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -470,27 +475,43 @@ impl HostScript {
...
     pub fn handle_transfer(
-        headers: &impl Deref<Target = Headers>,
+        headers: &mut impl std::ops::DerefMut<Target = Headers>,
```
zcabter, line 491: Is &mut Headers not allowed here?

zcabter:
Only nit comments, lgtm otherwise

crates/jstz_cli/src/main.rs
ryutamago: done

crates/jstz_cli/src/run.rs
ryutamago: done

crates/jstz_proto/src/executor/smart_function.rs
ryutamago: good catch, we can simplify it now

zcabter (approved):
Letting huan finish the review

huancheng-trili (approved):

PR #854, opened 2025-03-05
chore: use `nix fmt` for fmt and fmt-check targets in Makefile

johnyob:
# Context

Following #837, the pre-commit hook fails (since `fa2.js` is not formatted). 

There are two issues here:
1. The configured ignore `crates/jstz_tps_bench/fa2.js` should be `.prettierignore` instead of the global ignores (for treefmt)
2. The `make fmt` targets don't use Nix

# Description

This PR updates `make fmt` and `make fmt-check` to use `nix fmt`. 
Additionally it moves the ignore for `crates/jstz_tps_bench/fa2.js` to `.prettierignore` 
(and a comment instructing that future changes should be made in `.rustfmt.toml`/`.prettierignore`
 for language-specific ignores)

# Manually testing the PR

Test your pre-commit hook works by committing something :)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,4 +13,6 @@
 # Lock files
 **/*.lock
 
+# Ignore generated files:
 crates/jstz_node/openapi.json
+crates/jstz_tps_bench/fa2.js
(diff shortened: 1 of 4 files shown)


zcabter (approved):
lgmt

PR #856, opened 2025-03-07
ci: build static x64 cli binary

huancheng-trili:
# Context

Completes JSTZ-345.
[JSTZ-345](https://linear.app/tezos/issue/JSTZ-345/build-static-cli-binary-for-x86-64)

# Description

Build CLI for `x86_64-unknown-linux-musl` in order to get static x64 cli binaries.

I tried to use `pkgsMusl` in nix, but nix tried to rebuild a whole bunch of irrelevant packages for musl, which took hours, and in the end there were some error in those in irrelevant packages, and again mozjs does not have a prebuilt for `x86_64-unknown-linux-musl`, so I'm building this as how it's done for aarch64-linux.

# Manually testing the PR

[Test workflow run](https://github.com/jstz-dev/jstz/actions/runs/13699721004/job/38310216001)

Also tested the built binary in a clean x64 environment.

### DIFF
--- a/.github/workflows/cli-npm.yaml
+++ b/.github/workflows/cli-npm.yaml
@@ -17,13 +17,30 @@ jobs:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
   linux-amd64:
     name: Build CLI for Linux AMD64
-    runs-on: [x86_64, linux, nix]
+    runs-on: ubuntu-24.04
+    # need to build in an alpine container so that cargo can build with musl
+    # target: x86_64-unknown-linux-musl
+    container:
+      image: alpine:3.21
     steps:
-      - uses: jstz-dev/jstz/.github/actions/build-cli@main
+      - name: Install dependencies
+        shell: sh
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-npm.yaml
```diff
@@ -17,21 +17,38 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    # need to build in an alpine container so that cargo can build with musl
+    # target: x86_64-unknown-linux-musl
+    container:
+      image: alpine:3.21
     steps:
-      - uses: jstz-dev/jstz/.github/actions/build-cli@main
+      - name: Install dependencies
+        shell: sh
+        run: apk add make musl-dev libcrypto3 openssl-dev clang gcc musl-dev libstdc++ libffi-dev g++ openssl-libs-static bash curl
```
zcabter, line 28: Hmm we only need `musl` when cross-compiling the kernel to `RISCV`. Since we don't have that yet, we probably don't need the `musl` dependency here?

zcabter (approved):

zcabter:
LGTM

PR #860, opened 2025-03-10
chore: replace old sandbox with jstzd

huancheng-trili:
# Context

Completes JSTZ-341.
[JSTZ-341](huanchengchang/jstz-341)

# Description

Replaces the old sandbox daemon with jstzd and removes relevant obsolete things, e.g. the `USE_JSTZD` flag. Note that some things in the sandbox config struct are yet not cleaned up, e.g. `sandbox`. It's kind of messy and therefore I don't feel like removing everything in one PR.

Note that the way of checking if a sandbox is running is changed from checking `sandbox` in the config to calling the jstzd server. This means that `jstz sandbox stop` can kill both a sandbox running in the host environment and a sandbox running in a container. I think this is fine as users are not allowed to run more than one sandbox now anyway.

# Manually testing the PR

Ran `jstz sandbox start/stop` locally.


huancheng-trili:
Oh yes because this assumes that `jstzd` is in your `PATH` and I didn't want to add this dependency on jstzd binary to the nix cli derivation. Whether or not we should add this is another story, but in any case, I'm going to fix this pretty much useless error message for this.

ryutamago (approved):
Nice, verified jstzd works locally with cli!🧹

PR #862, opened 2025-03-10
feat(runtime): add JstzRuntime

zcabter:
# Context
Adds `JstzRuntime`, the top level struct that will manage the `JsRuntime` state. 
This PR sets up for the implementation of runtime APIs. Check out https://github.com/jstz-dev/jstz/pull/863, https://github.com/jstz-dev/jstz/pull/868/filesfor an example.

# Description
*  For `JstzRuntime` to interact with protocol state, we pre-seed the initialization with a `Protocol` that allows us to expose instances of `HostRuntime` and `Transaction` to extension logic. 
* ~`deno_core` exposes the ability to take snapshots of the running instance which we can use to quickly create isolated, sandboxed environments for each Smart Function evaluation but I couldn't got that to work yet 😞 ~ Removed snapshot function
* Currently, `JstzRuntime` can only execute plain, non ECMAScript module Javascript. Supporting default handler will be done in a follow up PR. See [JSTZ-366](https://linear.app/tezos/issue/JSTZ-366/load-init-and-run-default-handler) 
* Introduces `init_mock_host()` util which initializes the MockHost and sets & exposes a vec buffer for the debug msg sink. This allows us to inspect` debug_msg` without reading from stdout. e 
* The runtime fns return `Option`s for now. Refactoring to a proper error type can be done later

# Manually testing the PR
Will be tested in subsequent PRs that make use of the structure

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -304,12 +304,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "base32"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "022dfe9eb35f19ebbcb51e0b40a5ab759f46ad60cadf7297e0bd085afb50e076"
-
-[[package]]
 name = "base58"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


huancheng-trili (approved):
Since this PR is the base for the coming ones, I assume that the low test coverage is acceptable.

PR #863, opened 2025-03-10
feat(runtime): add support for console

zcabter:
# Context
Adds `console` support. It is implemented using the existing `deno_console` extension.

# Description
* To add console support, we register an extension that provides the "print" capability, in our case `debug_msg`, to `deno_console` extension. `HostRuntime` is exposed through borrowing an instance of it from the pre-seeded `op_state`
* `extension!` creates a new Rust module with `name` that includes the `Extension` type and logic needed to instantiate it. To register the console, we create a hard dependency on `deno_console` then attach an instance of console to `globalThis` in js. We need to make the `extension` aware of the js files and execute the initializer by settings `esm` and `esm_entry_point` - without this step, the extension will not be registered to the global object.  Lastly, we register the extensions by including it in the `extensions` field of `RuntimeOptions` on init

# Manually testing the PR [WIP]
`cargo nextest run -p jstz_runtime console`


huancheng-trili:
Can you explain in the description how this works? Does module registration now needs a JS module as well?

And does `import("console")` work?

zcabter:
> Can you explain in the description how this works? Does module registration now needs a JS module as well?
I've updated the description. 

> And does `import("console")` work?
Its callable but it won't resolve because we don't support async at the top level yet. Will be done as part of module loading

crates/jstz_runtime/src/jstz_console/console.js
johnyob, line 1: I propose we follow Deno's naming for JS files to include a loading priority. I think this would be `02_console.js`

zcabter: Hmm I don't quite understand the benefit for us since these files are referenced where the extension is defined.
Extension has a dependency field that enforces extension dependencies at runtime.

zcabter: I'll close this first for the sake of progressing but happy to revisit and update the js file names as needed.

huancheng-trili (approved):

PR #868, opened 2025-03-12
feat(proto-runtime): add kv

zcabter:
# Context

# Description

# Manually testing the PR


crates/jstz_runtime/src/jstz_kv/kv.js
```diff
@@ -0,0 +1,9 @@
+import { KV } from "ext:core/ops";
+
+Object.freeze(KV);
+Object.defineProperty(globalThis, "KV", {
```
huancheng-trili, line 4: Do we want to use `KV` instead of `Kv`? We will need to update the docs if we want `KV`.

johnyob: I propose we keep `Kv`. It is also consistent with Deno's naming for `KV` :)

https://docs.deno.com/api/deno/~/Deno.Kv

huancheng-trili (approved):

crates/jstz_runtime/src/jstz_kv/kv.js
```diff
@@ -0,0 +1,9 @@
+import { Kv } from "ext:core/ops";
+
+Object.freeze(Kv);
+Object.defineProperty(globalThis, "Kv", {
+  value: Kv,
+  enumerable: false,
+  configurable: true,
```
alanmarkoTrilitech, line 7: why is it set configurable? is there any advantage in allowing the the attributes of Kv to be modified?

crates/jstz_runtime/src/jstz_kv/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+impl Kv {
+    #[static_method]
+    #[serde]
+    fn get(op_state: &mut OpState, #[string] key: &str) -> Option<serde_json::Value> {
+        let Protocol { host, tx, kv } = op_state.borrow_mut::<Protocol>();
+        kv.get(host, tx, key).map(|v| v.0.clone())
+    }
+
+    #[static_method]
+    fn set(
```
alanmarkoTrilitech, line 20: why is set/get not #[fast] too?

alanmarkoTrilitech (approved):
LGTM! added some questions

crates/jstz_runtime/src/jstz_kv/mod.rs
zcabter: These cannot be annotated with `#[fast]` because it contains `#[serde]` arguments which cannot take the fast optimized route

crates/jstz_runtime/src/jstz_kv/kv.js
zcabter: It doesn't make much difference here, updated it to false. I've also set `enumerable` to `true` so that it appears as a property on the global object

PR #871, opened 2025-03-12
chore: remove jstz_engine

johnyob:
# Context

We're switching to Deno and `jstz_runtime`. This PR removes `jstz_engine`

# Description

Deletes all associated Nix / Rust files for `jstz_engine`. 

Additionally removes any dependencies from the workspace that are not needed.

# Manually testing the PR

N/A


zcabter (approved):
😢

PR #873, opened 2025-03-13
feat(jstz_core): implement reveal data

ryutamago:
# Context

Part of [adding support for large smart function](https://linear.app/tezos/project/large-smart-function-deployment-e1f8c66e4e03/overview)

[task url](https://linear.app/tezos/issue/JSTZ-358/implement-reveal-data)

# Description

Added `RevealData` struct that introduces apis to
1. reveal data greater than 4KB via the reveal data channel.
2. help encode the data and chunk them into preimages of size 4KB. 

The max size of data supported is now 10MB.

# Manually testing the PR

Added uni tests:
```
cargo test --package jstz_core   
```

### DIFF
--- a/crates/jstz_core/src/reveal_data.rs
+++ b/crates/jstz_core/src/reveal_data.rs
@@ -1,0 +1,194 @@
+use crate::error::Result;
+use crate::{host::HostRuntime, BinEncodable};
+use derive_more::From;
+use serde::{Deserialize, Serialize};
+use serde_big_array::BigArray;
+use std::fmt;
+use tezos_smart_rollup::{
+    core_unsafe::PREIMAGE_HASH_SIZE,
+    dac::{
+        self, PreimageHash as DacPreimageHash, PreimageHashError, V0SliceContentPage,
+        MAX_PAGE_SIZE,
+    },
+};
+use thiserror::Error;
+
+// Maximum number of DAC levels to support, can reveal up to 59MB of data.
+const MAX_DAC_LEVELS: usize = 3;
+// Support `MAX_DAC_LEVELS` levels of hashes pages, + the bottom layer of content.
+const MAX_REVEAL_BUFFER_SIZE: usize = MAX_PAGE_SIZE * (MAX_DAC_LEVELS + 1);
+/// maximum size of the reveal data in bytes (10MB)
+pub const MAX_REVEAL_SIZE: usize = 10 * 1024 * 1024;
+
+/// A 33-byte hash corresponding to a preimage
+type RawPreimageHash = [u8; PREIMAGE_HASH_SIZE];
+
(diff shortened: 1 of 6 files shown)


crates/jstz_core/src/reveal_data.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+impl fmt::Display for PreimageHash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", hex::encode(self.as_ref()))
+    }
+}
+
+#[derive(Display, Debug, Error, From)]
+pub enum Error {
```
huancheng-trili, line 65: I think we can use [thiserror](https://docs.rs/thiserror/latest/thiserror/) here to provide more user-friendly error messages. Since you already wrote the docstrings, you can easily use them as the error messages (and then you might want to add some simple tests for the error messages.)

crates/jstz_core/src/reveal_data.rs
```diff
@@ -0,0 +1,193 @@
...
+    #[test]
+    fn encode_and_prepare_preimages_fails_if_size_exceeds() {
+        let preimage = TestData(vec![0; MAX_REVEAL_SIZE + 1]);
+        let err = RevealData::encode_and_prepare_preimages(&preimage, |_, _| {})
+            .expect_err("should fail");
+        assert!(matches!(err, Error::RevealDataError { .. }));
+    }
+
+    #[test]
+    fn test_encode_and_decode_pwith_rdc() {
```
huancheng-trili, line 163:
```suggestion
    fn test_encode_and_decode_with_rdc() {
```

crates/jstz_core/src/reveal_data.rs
ryutamago: nice, done!

ryutamago: i don't have much description to add here

huancheng-trili: Then what's the description in `RevealDataError { description: String }` for?

ryutamago: it is an error thrown when the data is being revealed wihtin `reveal`

ryutamago: > let's add description to the error message. I assume that's what description is for

could you elaborate on this more? the `description` here is an error thrown when the data is being revealed. this will be the error that happens wihtin `reveal_loop`, which is how we reveal the data.

huancheng-trili: No. `#[error("Errors encountered while revealing data: {description}")]` as you have implemented.

huancheng-trili (approved):

PR #875, opened 2025-03-18
feat(runtime): wpt for jstz_runtime

huancheng-trili:
# Context

Completes JSTZ-381.
[JSTZ-381](https://linear.app/tezos/issue/JSTZ-381/make-wpt-work-with-jstz-runtime)

# Description

Creates a WPT runner with the new runtime based on the same setup for the old WPT runner.

Note that even if the new runner seems to be faster than the old one and can finish all tests much sooner, I'm still keeping the `0-5` filter for subset tests just to avoid having more than 100k lines in the report file. We can relax this later on when we are sure that it really runs fast enough. Most tests fail now because of the missing APIs and this might contribute to the short test execution time.

# Manually testing the PR

```sh
cd crates/jstz_runtime && cargo test --test wpt
```


crates/jstz_runtime/tests/wpt.rs
```diff
@@ -0,0 +1,364 @@
...
+    // Somehow each `execute_script` call has some strange side effect such that the global
+    // test suite object is completed prematurely before all test cases are registered.
+    // Therefore, instead of executing each piece of test scripts separately, we need to
+    // collect them and run them all in one `execute_script` call.
+    let mut source = String::new();
+    for item in &bundle.items {
+        match item {
+            BundleItem::TestHarnessReport => {
+                // Register test callback
+                source += "add_result_callback(globalThis.test_result_callback); add_completion_callback(globalThis.test_completion_callback);";
```
ryutamago, line 269: q: sorry if i'm lacking wpt knoweldge but what does the each vairant of the bundle item mean and look like? 

and where does these `add_result_callback` and `add_completion_callback` come from for `TestHarnessReport`? Do we not have to make a test report for every tests anyway ?

ryutamago (approved):
left some nits/questions but the flow makes sense to me! thanks for the detailed explanation :)

crates/jstz_runtime/tests/test_harness_api.js
```diff
@@ -0,0 +1,29 @@
...
+
+Object.defineProperty(globalThis, "self", {
+  value: globalThis,
+  enumerable: true,
+  configurable: true,
+  writable: true,
+});
+
+Object.defineProperty(globalThis, "test_result_callback", {
+  value: globalThis.Deno.core.ops.test_result_callback,
```
ryutamago, line 18: nit:
can we not import the following functions like
```js
import { test_result_callback } from "ext:core/ops";
```

crates/jstz_runtime/tests/wpt.rs
```diff
@@ -0,0 +1,364 @@
...
+) -> Result<(), JsErrorBox> {
+    let report: &mut TestHarnessReport = op_state.borrow_mut::<TestHarnessReport>();
+    report
+        .set_harness_result(result)
+        .map_err(|e| JsErrorBox::generic(e.to_string()))
+}
+
+deno_core::extension!(
+    test_harness_api,
+    ops = [test_completion_callback, test_result_callback],
```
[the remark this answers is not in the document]
huancheng-trili: By the test script itself. If you look into the actual wpt repository, tests are usually defined by calling a `test` function that wraps the test content. This test function does quite a few things, one of which is calling these callback functions when a test completes.

crates/jstz_runtime/tests/wpt.rs
huancheng-trili: They are defined in the test scripts. A bundle is essentially a test script that represents a test suite. Each test script is like a comprehensive script that includes all helper functions (including `add_result_callback`) and the test content. A `TestHarnessReport` is unique to each test suite and each test suite might contain multiple tests.

crates/jstz_runtime/tests/wpt.rs
```diff
@@ -0,0 +1,364 @@
...
+use tezos_smart_rollup_mock::MockHost;
+
+/// Enum of possible test statuses
+///
+/// More information:
+///  - [wpt documentation][wpt]
+///
+/// [wpt]: https://web-platform-tests.org/writing-tests/testharness-api.html#Test.statuses
+#[derive(Debug, From, Into)]
+pub struct TestStatus(WptSubtestStatus);
```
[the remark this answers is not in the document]
huancheng-trili: I think `WptSubtestStatus` makes sense here because sub-test is one word.

PR #876, opened 2025-03-19
feat(jstz_proto): add reveal large payload op

ryutamago:
# Context

Part of [large sf deployment](https://linear.app/tezos/project/large-smart-function-deployment-e1f8c66e4e03/issues)
[task link](https://linear.app/tezos/issue/JSTZ-360/support-reveallargepayloadoperation)

# Description

* add a new operation `RevealLargePayLoadOperation` 
* refactored and modified the `execute_external_operation` to handle the `RevealLargePayLoadOperation` 

# Manually testing the PR

add unit test 
```
cargo test -p jstz_proto
```

### DIFF
--- a/crates/jstz_proto/src/executor/mod.rs
+++ b/crates/jstz_proto/src/executor/mod.rs
@@ -1,10 +1,12 @@
-use jstz_core::{host::HostRuntime, kv::Transaction};
+use jstz_core::{host::HostRuntime, kv::Transaction, reveal_data::RevealData};
 use tezos_crypto_rs::hash::ContractKt1Hash;
 
 use crate::{
-    operation::{self, ExternalOperation, SignedOperation},
+    operation::{
+        self, ExternalOperation, Operation, OperationHash, RevealType, SignedOperation,
+    },
     receipt::{self, Receipt},
-    Result,
+    Error, Result,
 };
 
 pub mod deposit;
@@ -13,38 +15,6 @@ pub mod fa_withdraw;
 pub mod smart_function;
 pub mod withdraw;
 pub const JSTZ_HOST: &str = "jstz";
-
-fn execute_operation_inner(
-    hrt: &mut impl HostRuntime,
-    tx: &mut Transaction,
-    signed_operation: SignedOperation,
-    ticketer: &ContractKt1Hash,
-) -> Result<receipt::ReceiptContent> {
-    let operation = signed_operation.verify()?;
-    let operation_hash = operation.hash();
-
-    operation.verify_nonce(hrt, tx)?;
-
-    let source = operation.source();
-
-    match operation.content {
-        operation::Content::DeployFunction(deployment) => {
-            let result = smart_function::deploy::execute(hrt, tx, &source, deployment)?;
-
-            Ok(receipt::ReceiptContent::DeployFunction(result))
-        }
-
-        operation::Content::RunFunction(run) => {
-            let result = match run.uri.host() {
-                Some(JSTZ_HOST) => {
-                    smart_function::jstz_run::execute(hrt, tx, ticketer, &source, run)?
-                }
-                _ => smart_function::run::execute(hrt, tx, &source, run, operation_hash)?,
-            };
-            Ok(receipt::ReceiptContent::RunFunction(result))
-        }
-    }
-}
 
 pub fn execute_external_operation(
     hrt: &mut impl HostRuntime,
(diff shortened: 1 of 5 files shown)


crates/jstz_proto/src/executor/mod.rs
```diff
@@ -14,64 +16,234 @@ pub mod smart_function;
...
+            Ok((op_hash, receipt::ReceiptContent::RunFunction(result)))
+        }
+        operation::Content::RevealLargePayloadOperation(reveal) => {
+            let revealed_op = RevealData::reveal_and_decode::<_, SignedOperation>(
+                hrt,
+                &reveal.root_hash,
+            )?;
+            let revealed_op = verify_signed_op(hrt, tx, revealed_op)?;
+            match (reveal.reveal_type, &revealed_op.content) {
+                (RevealType::DeployFunction, &operation::Content::DeployFunction(_)) => {
```
huancheng-trili, line 79: Is there any plan to have more than one reveal type?

crates/jstz_proto/src/executor/mod.rs
```diff
@@ -14,64 +16,234 @@ pub mod smart_function;
...
+            .unwrap(),
+            SecretKey::from_base58(
+                "edsk39qAm1fiMjgmPkw1EgQYkMzkJezLNewd7PLNHTkr6w9XA2zdfo",
+            )
+            .unwrap(),
+        )
+    }
+
+    fn run_function_content() -> Content {
+        let body = r#""value":1""#.to_string().into_bytes();
```
huancheng-trili, line 146: The quotation marks don't match. Is this intended?

crates/jstz_proto/src/operation.rs
```diff
@@ -80,6 +82,12 @@ impl Operation {
                 )
                 .as_bytes(),
             ),
+            Content::RevealLargePayloadOperation(RevealLargePayloadOperation {
+                root_hash,
+                reveal_type,
+            }) => Blake2b::from(
+                format!("{}{}{:?}{:?}", source, nonce, root_hash, reveal_type).as_bytes(),
```
huancheng-trili, line 89: What does `{:?}` give exactly with `root_hash` and `reveal_type`? If we need `{:?}` because they don't implement Display, perhaps it's better that we implement it as I'm a bit concerned about surprises brought by `{:?}`.

crates/jstz_proto/src/executor/mod.rs
```diff
@@ -14,64 +16,234 @@ pub mod smart_function;
...
+        .and_then(|op| execute_op(hrt, tx, op, ticketer));
+    match result {
+        Ok((hash, content)) => Receipt::new(hash, Ok(content)),
+        Err(e) => Receipt::new(operation_hash, Err(e)),
+    }
+}
 
-            Ok(receipt::ReceiptContent::DeployFunction(result))
-        }
+fn execute_op(
```
huancheng-trili, line 47: Can you add some comments explaining which cases `execute_op`, `execute_external_operation`, and `execute_signed_operation` handle and the main difference between them (and rename `execute_op` if necessary)? These names are getting less obvious now. Yes, it's surely possible to trace them from references, but I think it's important to avoid calling the wrong one in the future.

crates/jstz_proto/src/executor/mod.rs
ryutamago: not for this project but we could easily support otehr types

crates/jstz_proto/src/executor/mod.rs
ryutamago: removd

crates/jstz_proto/src/executor/mod.rs
ryutamago: i renamed the `execute_op` to `execute_verified_op`. 

I think the `execute_external_operation` and `execute_signed_operation` is obvious since it takes differnt operation type.

zcabter: I think we should leave the renaming of functions for a separate PR. Kinda hard to follow the important changes.

crates/jstz_proto/src/executor/mod.rs
huancheng-trili: Um it looks the same. The actual string is still `"value":1"`

crates/jstz_proto/src/operation.rs
ryutamago: done!

crates/jstz_proto/src/executor/mod.rs
ryutamago: ops, i've chagned the differnt part of the codenase that used the same exxact code. applied the fix!

crates/jstz_proto/src/executor/mod.rs
ryutamago: @huancheng-trili i will skip the comment/renaming for now and apply it on the next PR

huancheng-trili (approved):

crates/jstz_proto/src/executor/mod.rs
```diff
@@ -1,53 +1,67 @@
...
-                _ => smart_function::run::execute(hrt, tx, &source, run, operation_hash)?,
+                _ => {
+                    smart_function::run::execute(hrt, tx, &source, run, op_hash.clone())?
+                }
             };
-            Ok(receipt::ReceiptContent::RunFunction(result))
+            Ok((op_hash, receipt::ReceiptContent::RunFunction(result)))
+        }
+        operation::Content::RevealLargePayloadOperation(reveal) => {
+            let revealed_op = RevealData::reveal_and_decode::<_, SignedOperation>(
```
zcabter, line 54: Before revealing a large payload, we need to check that `source` is the authorized sener right?

ryutamago: that's going to be handled at the end

crates/jstz_proto/Cargo.toml
```diff
@@ -24,6 +24,8 @@ jstz_core = { path = "../jstz_core" }
 jstz_crypto = { path = "../jstz_crypto" }
 serde.workspace = true
 serde_json.workspace = true
+strum = "0.26"
+strum_macros = "0.26"
```
zcabter, line 28: Where are these used?

ryutamago: i think cursor added silently, i will remove !

crates/jstz_proto/src/executor/mod.rs
zcabter: Ok, add `TODO` and link to the task next time

zcabter (approved):
LGTM

PR #879, opened 2025-03-20
fix(cli): replace localhost with 127.0.0.1

huancheng-trili:
# Context

Part of JSTZ-355.
[JSTZ-355](https://linear.app/tezos/issue/JSTZ-355/replace-all-localhost-with-127001)

# Description

Replace references to `localhost` in CLI with `127.0.0.1`.

Also refactored a bit to make things more testable.

# Manually testing the PR

Tested running the sandbox locally both in the host environment and in a container.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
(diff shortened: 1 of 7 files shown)


ryutamago (approved):
nice to be able to mock the server🙌

PR #884, opened 2025-03-25
feat(jstz_runtime): added global.selfAddress property

alanmarkoTrilitech:
# Context
https://linear.app/tezos/issue/JSTZ-370/implement-selfaddress
# Description
This PR adds global.selfAddress property to the jstz runtime returning the address of the smart function.

# Manual testing
cargo nextest run -p jstz_runtime

### DIFF
--- a/crates/jstz_runtime/src/jstz_kv/kv.rs
+++ b/crates/jstz_runtime/src/jstz_kv/kv.rs
@@ -16,7 +16,7 @@ use utoipa::ToSchema;
 
 #[derive(Debug)]
 pub struct Kv {
-    prefix: String,
+    pub prefix: String,
 }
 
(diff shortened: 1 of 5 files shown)


zcabter:
Actually, the smart function can simply inspect the host base address to get its self address 😓

alanmarkoTrilitech:
> Actually, the smart function can simply inspect the host base address to get its self address 😓 we probably don't need separate function for it

oh ok. how will calling it look like then?

zcabter:
> > Actually, the smart function can simply inspect the host base address to get its self address 😓 we probably don't need separate function for it
> 
> oh ok. how will calling it look like then?

It should be something like `new Url(req.url).hostname`

zcabter:
Closing as this is not needed anymore.

PR #885, opened 2025-03-26
chore(deps): replace tiny-bip39 with bip39

huancheng-trili:
# Context

Part of JSTZ-403.
[JSTZ-403](https://linear.app/tezos/issue/JSTZ-403/passphrases-and-tz1-generation-are-different-to-umami-and-octez-cli)

# Description

The version of `tiny-bip39` in this repo is out of date and `tiny-bip39` itself lacks some features. Replacing it with `bip39` allows us to do more things easily. It seems that there used to be some issues with `bip39` and `tiny-bip39` was created as a fork to address those issues. It's been a while since `tiny-bip39` was created and added to this repo and I assume that now `bip39` is fine with the upgrades.

# Manually testing the PR

* Manual testing: CLI still works
* Unit testing: added one test to monitor bip39 API

### DIFF
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -1,6 +1,6 @@
 use std::collections::hash_map::Entry;
 
-use bip39::{Language, Mnemonic, MnemonicType};
+use bip39::{Language, Mnemonic};
 use clap::Subcommand;
 use dialoguer::{Confirm, Input};
 use jstz_crypto::smart_function_hash::SmartFunctionHash;
@@ -15,7 +15,10 @@ use crate::{
 };
 
 fn generate_passphrase() -> String {
-    let mnemonic = Mnemonic::new(MnemonicType::Words12, Language::English);
+    // unwrap is okay here because we are using a fixed value for word count and it's always
+    // valid unless the library stops supporting word count 12
+    let mnemonic = Mnemonic::generate_in(Language::English, 12)
+        .expect("generate_in should generate mnemonics");
     mnemonic.to_string()
 }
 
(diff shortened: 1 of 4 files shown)


crates/jstz_cli/src/account.rs
```diff
@@ -351,3 +353,14 @@ pub async fn exec(command: Command) -> Result<()> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn generate_passphrase() {
+        // Just to make sure that generate_passphrase works with the current version of Mnemonic.
+        // If anything changes in the library and fails our logic, the unwrap call will lead to
+        // a panic and we can capture that issue here.
+        assert_ne!(super::generate_passphrase(), "");
```
alanmarkoTrilitech, line 364: maybe  test that two consecutively generated mnemonics are not equal might be good?

crates/jstz_cli/src/account.rs
```diff
@@ -15,7 +15,9 @@ use crate::{
 };
 
 fn generate_passphrase() -> String {
-    let mnemonic = Mnemonic::new(MnemonicType::Words12, Language::English);
+    // unwrap is okay here because we are using a fixed value for word count and it's always
+    // valid unless the library stops supporting word count 12
+    let mnemonic = Mnemonic::generate_in(Language::English, 12).unwrap();
```
alanmarkoTrilitech, line 20: still why not have there something like - .expect("Failed to generate mnemonic.") instead?

huancheng-trili: Updated, though I don't think this matters here

crates/jstz_cli/src/account.rs
alanmarkoTrilitech: yeah, but small improvements compound over time :D

alanmarkoTrilitech (approved):

PR #888, opened 2025-03-28
fix(cli): always deposit through jstzd in sandbox

huancheng-trili:
# Context

Part of JSTZ-399.
[JSTZ-399](https://linear.app/tezos/issue/JSTZ-399/make-sure-that-cli-config-works-when-users-switch-between-native-and)

Previously, for deposit commands, CLI always called octez client to perform corresponding operations. This does not work with the containerised sandbox because it's a bit troublesome keeping track of the octez base directory that contains necessary data, so a `container` flag in the sandbox config was created and CLI goes through the exposed jstzd endpoint to tell octez client in the container to perform such operations. This gets a bit messy and difficult to maintain.

# Description

Make CLI deposit command always go through jstzd in the dev environment to mitigate the issue above.

# Manually testing the PR

Ran the sandbox (both in the host environment and in a container) locally and confirmed that it worked.
Added one not very useful unit test to cover the changed lines.

### DIFF
--- a/crates/jstz_cli/src/bridge/deposit.rs
+++ b/crates/jstz_cli/src/bridge/deposit.rs
@@ -20,9 +20,9 @@ pub async fn exec(
     network: Option<NetworkName>,
 ) -> Result<()> {
     let cfg = Config::load().await?;
-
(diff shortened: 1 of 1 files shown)


ryutamago (approved):
nice!

PR #891, opened 2025-03-31
feat(jstz_node): prepare rlp operation

ryutamago:
# Context

[task url](https://linear.app/tezos/issue/JSTZ-362/hanlde-large-payload-operation-in-jstz-node)

# Description

Jstz node now observe the operation before injecting into the rollup and make the data avaible if it's > 4KB. 

* lifted the code size limit in the cli from 4KB to 10MB
* modify the jstz node to make data available for a large DeployFunction operation (signing with bootstrap1 by default)
* tweaked the protocol to associate the `RevealType` to the corresponding operation content type

In the folloiwng PR, we add authentication to the kernel to ensure the reveal large payload operation comes from jstz node (bottstrap1)

NOTE:
I considered moving the encoding/prepare large operation logic into a dedicated module or struct, but since there’s only one function and no shared state or multiple steps, it’s simpler and clearer as a single function in the same file. If the code grows more complex in the future, we can revisit extracting it then.

# Manually testing the PR

will add a large sf example on the following PR to test manually locally


crates/jstz_node/src/config.rs
```diff
@@ -1,28 +1,59 @@
...
 impl JstzNodeConfig {
+    /// Create a new JstzNodeConfig.
+    ///
+    /// If `injector` is not provided, bootstrap1 account will be used as the injector.
     pub fn new(
         endpoint: &Endpoint,
         rollup_endpoint: &Endpoint,
+        rollup_preimages_dir: &Path,
         kernel_log_file: &Path,
+        injector: Option<KeyPair>,
```
huancheng-trili, line 49: Why does this need to be an option? `unwrap_or_default` here can be dangerous since the keys are exposed. The default keys should only be used in the sandbox. (You passed in `None` in the jstzd tests anyway so it's pretty much the same as passing in the default keys.)

crates/jstz_node/src/services/accounts.rs
```diff
@@ -84,15 +100,10 @@ async fn get_nonce(
...
-    let value = rollup_client.get_value(&key).await?;
-    let account_nonce = match value {
-        Some(value) => match deserialize_account(value.as_slice())? {
-            Account::User(UserAccount { nonce, .. }) => nonce,
-            Account::SmartFunction(SmartFunctionAccount { nonce, .. }) => nonce,
-        },
-        None => Err(ServiceError::NotFound)?,
-    };
+    let account_nonce = get_account_nonce(&rollup_client, &address).await?;
+    if account_nonce.0 == 0 {
```
huancheng-trili, line 104: This and `None => Ok(Nonce(0))` above change the logic, right? When a new account first makes an operation, its nonce value starts from 0. The new logic here means that the account is still not found even after the first operation is performed.

crates/jstz_node/src/services/operations.rs
```diff
@@ -6,44 +12,120 @@ use axum::{
...
+
 pub struct OperationsService;
 
 const OPERATIONS_TAG: &str = "Operations";
 
 type HexEncodedOperationHash = String;
 
+// Given a large operation, encode it into preimages and store them in the rollup's preimages directory
+async fn prepare_rlp_operation(
+    rollup_client: &OctezRollupClient,
```
huancheng-trili, line 37: I think you can create a mock rollup server in unit tests and test these two new functions.

crates/jstz_node/src/services/operations.rs
```diff
@@ -6,44 +12,120 @@ use axum::{
...
+        .map_err(|_| anyhow!("Invalid operation"))?
+        .content()
+        .try_into()
+        .map_err(|_| {
+            ServiceError::BadRequest("Large payload operation not supported".to_string())
+        })?;
+    let save_preimages = |hash: PreimageHash, preimage: Vec<u8>| {
+        let path = rollup_preimages_dir.join(hash.to_string());
+        fs::write(&path, preimage)
+            .unwrap_or_else(|_| panic!("failed to save preimage at: {:?}", path));
```
huancheng-trili, line 53: Why do we need to panic here?

crates/jstz_node/src/services/operations.rs
```diff
@@ -6,44 +12,120 @@ use axum::{
...
+    };
+    let KeyPair(public_key, secret_key) = signer;
+    let root_hash =
+        RevealData::encode_and_prepare_preimages(operation, save_preimages)
+            .map_err(|_| anyhow!("failed to prepare reval large payload operation"))?;
```
huancheng-trili, line 58: Not sure how this will affect the response message, but what about using `.context` here? A bit concerned about swallowing errors.

crates/jstz_proto/src/operation.rs
```diff
@@ -139,6 +143,16 @@ pub enum RevealType {
     DeployFunction,
 }
 
+impl TryFrom<&Content> for RevealType {
+    type Error = Error;
+    fn try_from(value: &Content) -> Result<Self> {
+        match *value {
+            Content::DeployFunction(_) => Ok(RevealType::DeployFunction),
+            _ => Err(Error::RevealNotSupported),
```
huancheng-trili, line 151: Why is `RevealTypeMismatch` changed to `RevealNotSupported`?

crates/jstz_node/src/services/operations.rs
ryutamago: it's a callback function to the dac library so we cannot return Result

crates/jstz_node/src/services/accounts.rs
ryutamago: good point, thanks!

crates/jstz_node/src/config.rs
ryutamago: I prefer to not make it optional either, IIRC we wanted to make it optional? wdyt @zcabter

crates/jstz_node/src/main.rs
```diff
@@ -52,8 +56,17 @@ async fn main() -> anyhow::Result<()> {
...
 
-            jstz_node::run(&args.addr, args.port, rollup_endpoint, args.kernel_log_path)
-                .await
+            jstz_node::run(
+                &args.addr,
+                args.port,
+                rollup_endpoint,
+                args.preimages_dir,
+                args.kernel_log_path,
+                // TODO: make it configurable via CLI args
```
huancheng-trili, line 65: A side note: I'm not sure about this idea of passing in keys as CLI arguments. These keys should be loaded from somewhere secure and should not be obviously visible.

huancheng-trili (commented):
Waiting for the optional injector part.

crates/jstz_cli/src/deploy.rs
```diff
@@ -58,8 +56,8 @@ pub async fn exec(
     let code = read_file_or_input_or_piped(code_op)?
         .ok_or(user_error!("No function code supplied. Please provide a filename or pipe the file contents into stdin."))?;
 
-    if code.bytes().len() > MAX_CODE_LENGTH {
-        bail_user_error!("The data availability layer is not yet available. Smart functions are currently restricted to {MAX_CODE_LENGTH} bytes");
+    if code.bytes().len() > MAX_REVEAL_SIZE {
+        bail_user_error!("The data availability layer is not yet available. Smart functions are currently restricted to {MAX_REVEAL_SIZE} bytes");
```
zcabter, line 60:
```suggestion
        bail_user_error!("Smart functions are currently restricted to {MAX_REVEAL_SIZE} bytes");
```

crates/jstz_proto/src/operation.rs
```diff
@@ -164,6 +178,18 @@ pub enum Content {
     RevealLargePayloadOperation(#[bincode(with_serde)] RevealLargePayloadOperation),
 }
 
+impl Content {
+    pub fn new_reval_large_payload(
```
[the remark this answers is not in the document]
ryutamago: nice catch, thanks

crates/jstz_node/src/services/operations.rs
```diff
@@ -1,49 +1,133 @@
...
+        .map_err(|_| anyhow!("Invalid operation"))?
+        .content()
+        .try_into()
+        .map_err(|_| {
+            ServiceError::BadRequest("Large payload operation not supported".to_string())
+        })?;
+    let save_preimages = |hash: PreimageHash, preimage: Vec<u8>| {
+        let path = rollup_preimages_dir.join(hash.to_string());
+        fs::write(&path, preimage)
+            .unwrap_or_else(|_| panic!("failed to save preimage at: {:?}", path));
```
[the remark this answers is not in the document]
ryutamago: fixed and added tests too

zcabter (approved):
LGTM

PR #893, opened 2025-04-01
refactor(cli): check sandbox status through jstzd

huancheng-trili:
# Context

Part of JSTZ-421.
[JSTZ-421](https://linear.app/tezos/issue/JSTZ-421/clean-up-sandbox-config)
Sandbox config is not really necessary after jstzd is integrated into CLI.

# Description

Make CLI check sandbox status through jstzd instead of by checking the sandbox config. Since jstzd server always listens on port 54321, it's sufficient to check if the sandbox is running properly through the health check endpoint.

# Manually testing the PR

* Manual testing: tested the sandbox locally
* Unit testing: added some test cases for the new helper function.


ryutamago (approved):
LGTM

PR #898, opened 2025-04-03
fix(riscv-sandbox): properly enable supervisor build

zcabter:
# Context
Fixes the issues with `riscv-sandbox` not building with supervisor feature

# Manually testing the PR
`nm $(which riscv-sandbox) | grep linux` should not be empty


alanmarkoTrilitech (approved):
🚀

PR #902, opened 2025-04-04
fix(doc): clarifications to the quickstart and samples

timothymcmackin:
# Context

Per request, some clarifications to the quickstart and comments in the example applications that it uses

# Description

# Manually testing the PR

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -1,6 +1,6 @@
 # 🚀 Quick Start
 
-This guide will instruct you in writing, deploying, and using your first _smart function_ in under 10 minutes.
+This guide will instruct you in writing, deploying, and using your first Jstz _smart function_ in under 10 minutes.
 
 ## Prerequisites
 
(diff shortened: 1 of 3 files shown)


docs/quick_start.md
```diff
@@ -243,18 +244,21 @@ Follow these instructions to deploy the sample smart function to a local sandbox
...
     </details>
 
-    Upon successful deployment, Jstz assigns the smart function a unique `KT1` address, serving as its identifier, similar to an IP address or a smart contract address.
+    Upon successful deployment, Jstz assigns the smart function a unique `KT1` address.
+    This address is its identifier, similar to an IP address or a smart contract address.
 
     In the example above, the smart function was deployed to the address `KT1FZuQ4SDP7ahLRyybtNnNxNnRskBGyAXVw`.
     Now the smart function is accessible through a URL of the format `tezos://KT1FZuQ4SDP7ahLRyybtNnNxNnRskBGyAXVw/`.
 
+    After you deploy the smart function, you cannot stop it or change it.
```
huancheng-trili, line 253: Just a minor comment: I know it's true that one cannot stop it, but I wonder if users would be a bit confused by what "stopping a smart function" means. In a previous paragraph, we tell users that smart functions only run when they are called, which means they are not long running processes like servers. Then will users ever think about stopping a smart function? If they do, does stopping it mean to abort a call? If so, it's a valid question, but it has little to do with deployment, which is what the text here is about.

I'm just thinking if we should mention "stop" here. "You cannot delete it or change it" sounds more natural in this context. About stopping a call... that's a different story to that we probably should document. There is a plan for implementing this feature that allows interrupting or aborting a function call, though I'm not completely sure if that will be user-facing.

docs/quick_start.md
```diff
@@ -103,11 +113,29 @@ const handler = async (request: Request): Promise<Response> => {
...
+
+- A few utility functions that read and write data with the Jstz key-value store.
+  Each time a smart function is called, it runs in a new environment and therefore must store any persistent data in the key-value store.
+  In this case, the smart function records how much tez it has sent to each address and also logs the messages that users send to it.
 
 - A `handler` function.
 
-  A smart function processes an HTTP `Request` object and returns a `Response` object, mirroring the functionality of conventional web server handlers or cloud functions.
+  The handler function is the sole entrypoint of a Jstz smart function; it is what runs when a client calls the smart function.
+  The handler function receives a Jstz HTTP [request](/api/request) object that includes the address of the account that called it in its `Referer` header and an optional request body.
```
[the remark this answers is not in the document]
timothymcmackin: ok, thanks for the clarification; I changed things to "However, smart functions behave more like web applications because they accept requests and return responses in a way similar to ordinary web-based HTTP requests." and removed the HTTP request name from other places.

docs/quick_start.md
timothymcmackin: Makes sense - I changed this to "you cannot delete it or change it" and will keep that in mind.

huancheng-trili: Um have you pushed the updates?

timothymcmackin: Sorry about that, my amend and force-push failed. It's updated now.

huancheng-trili (approved):

PR #909, opened 2025-04-08
feat(wpt): introduce expected_tests

huancheng-trili:
# Context

Part of JSTZ-436.
[JSTZ-436](https://linear.app/tezos/issue/JSTZ-436/collect-wpt-test-cases-run-by-deno)

# Description

Introduce `expected_tests` which is like an allow list for tests. Only tests in this set will be executed.

Imagine that the test suites run by deno is a list of file names. We want to make our WPT runner execute tests in this list. The original test filter is not enough for this because `RegexSet` has a maximum supported size and the list is larger than the acceptable size. Compressing the list into regex patterns like what we use right now will surely work, but it's not that straightforward and has the risk of accidentally including tests that deno does not run.

# Manually testing the PR

Added unit tests.


ryutamago (approved):
🛳️

PR #910, opened 2025-04-08
fix(cargo): enable utils feature in jstz_tps_bench

zcabter:
# Context
`jstz_tps_bench` not compiling due to missing import

# Description
Add utils feature to kernel sdk dependency


ryutamago (approved):
lgtm

PR #911, opened 2025-04-09
feat(runtime): support args in call default handler

zcabter:
# Context

We need to pass arguments (eg. Request) to the default handler

# Description
* Make default handler global so that the returned `Promise` can be translated into `async`
* Add arguments to handler
* Make `init_test_setup!` more flexible by allow bounded variables to be optional

# Manually testing the PR
`cargo nextest run -p jstz_runtime --features skip-wpt`

### DIFF
--- a/crates/jstz_runtime/src/runtime.rs
+++ b/crates/jstz_runtime/src/runtime.rs
@@ -11,7 +11,8 @@ use std::{
 use crate::error::Result;
 use crate::ext::jstz_fetch;
 use crate::ext::jstz_fetch::NotSupportedFetch;
-use deno_core::{error::JsError, *};
+use deno_core::*;
+
 use serde::Deserialize;
 use tokio;
 
@@ -115,7 +116,7 @@ impl JstzRuntime {
 
     /// Executes traditional, non-ECMAScript-module JavaScript code, ignoring
     /// its result
-    pub fn execute(mut self, code: &str) -> Result<()> {
+    pub fn execute(&mut self, code: &str) -> Result<()> {
         self.execute_script("jstz://run", code.to_string())?;
         Ok(())
     }
@@ -156,13 +157,6 @@ impl JstzRuntime {
         }?)
     }
 
-    pub async fn run_event_loop(
-        &mut self,
-        poll_options: PollEventLoopOptions,
-    ) -> Result<()> {
-        Ok(self.runtime.run_event_loop(poll_options).await?)
-    }
-
     /// Loads, instantiates and executes the specified JavaScript module.
     ///
     /// This module is treated as the "main" module. See [`preload_main_module`]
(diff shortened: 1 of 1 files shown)


crates/jstz_runtime/src/lib.rs
```diff
@@ -44,31 +44,124 @@ mod test_utils {
...
-            let mut $runtime = $crate::JstzRuntime::new($crate::JstzRuntimeOptions { protocol, ..Default::default() } );
+            let protocol  = Some($crate::Protocol::new(&mut $host, &mut $tx, $addr));
             #[allow(unused)]
-            let $sink = $crate::test_utils::Sink($sink);
+            let mut $runtime = $crate::JstzRuntime::new($crate::JstzRuntimeOptions {
+                protocol,
+                module_loader: std::rc::Rc::new($mod_loader),
+                ..Default::default()
+
+            });
```
alanmarkoTrilitech, line 164: I think it would be better to just use a macro with params - for example $( host = $host:ident; )? instead of this chain of calls.

crates/jstz_runtime/src/runtime.rs
```diff
@@ -369,4 +347,29 @@ export default handler;
         let result_i64 = result.unwrap().open(scope).integer_value(scope).unwrap();
         assert_eq!(result_i64, 42);
     }
+
+    #[tokio::test]
```
alanmarkoTrilitech, line 351: why is one test with #[tokio::test] and another with just #[test]?

crates/jstz_runtime/src/lib.rs
alanmarkoTrilitech: I don't think we need them to be defined in arbitrary order. It's much easier to read/maintain if it is always in the same order. 

For this approach if you really want to keep it, I think it should also have some tests that test that it actually works correctly with different sets of args. Currently its also a bit confusing that if the full set of params is provided that it doesn't just go in sequence as its written but its jumping.

zcabter: It looks like tt muncher is slowing down compile times to a crawl consequently driving the rust-analyzer into the ground :(

zcabter: Updated to conditional arguments

alanmarkoTrilitech: thanks! it looks good to me now! 🚢

alanmarkoTrilitech (approved):
:shipit:

crates/jstz_runtime/src/runtime.rs
```diff
@@ -177,30 +170,18 @@ impl JstzRuntime {
...
-            result
+        let default_fn = {
+            let scope = &mut self.handle_scope();
+            let default_value = get_default_export(ns, scope);
+            let default_fn = v8::Local::<v8::Function>::try_from(default_value)?;
+            v8::Global::new(scope, default_fn)
         };
-
-        Ok(v8::Global::new(scope, result.unwrap()))
+        let fut = self.call_with_args(&default_fn, args);
```
johnyob, line 182: Might be worth adding a comment that `call_with_args` wraps the `scope` in a `TryCatch` for us and converts any exception to an `Error` (at least I had to check the impl to verify this)

crates/jstz_runtime/src/runtime.rs
```diff
@@ -369,4 +347,29 @@ export default handler;
         let result_i64 = result.unwrap().open(scope).integer_value(scope).unwrap();
         assert_eq!(result_i64, 42);
     }
+
```
johnyob, line 350: We should probably also test that errors are correctly treated in promises and that promises are resolved

zcabter: done

crates/jstz_runtime/src/runtime.rs
zcabter: Done

PR #913, opened 2025-04-09
chore(wpt): update wpt

huancheng-trili:
# Context

Part of JSTZ-437.
[JSTZ-437](https://linear.app/tezos/issue/JSTZ-437/calculate-pass-rate-against-deno-test-set)

Deno runs a specific subset of the entire WPT test suites. We want to run the same subset for jstz.

# Description

Update wpt to [e78446e](https://github.com/web-platform-tests/wpt/tree/e78446e34a1921371658a5df08c71d83f50a2a2f). This is the version pinned by deno at the time v2.1.10 was released.

Note that the lines for comparing WPT results are removed from the test runners. Since the number of tests will grow quite a bit, I don't think it makes much sense to check in large files with rather limited information. We'll figure out how to track changes later on.

# Manually testing the PR

WPT job should still run normally.

### DIFF
--- a/crates/jstz_api/tests/wpt.rs
+++ b/crates/jstz_api/tests/wpt.rs
@@ -8,7 +8,6 @@ use boa_engine::{
 };
 use boa_gc::{Finalize, Trace};
 use derive_more::{From, Into};
-use expect_test::expect_file;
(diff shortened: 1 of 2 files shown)


ryutamago (approved):
💪

PR #915, opened 2025-04-09
rename rlp operation

ryutamago:
just renaming the 
`ReealLargePayloadOperation` to `RevealLargePayload` for consistency

### DIFF
--- a/crates/jstz_node/openapi.json
+++ b/crates/jstz_node/openapi.json
@@ -609,7 +609,7 @@
           {
             "allOf": [
               {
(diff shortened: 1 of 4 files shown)


alanmarkoTrilitech (commented):
@ryutamago is this PR rebased correctly? seems like there are many other changes included

alanmarkoTrilitech (approved):
LGTM! 🎉 

before merging can you please add/execute some manual testing section for the PR so that we know it works still the same way?

PR #918, opened 2025-04-10
feat(runtime): patch deno_fetch with jstz fork

zcabter:
# Context
Dangerous and non-blockchain compliant features from deno_fetch need to be removed or traitfied. This allows us to alter the behaviour of fetch dependning no the environment that we're running in (blockchain sandbox and wpt). Isolating changes allows to potentially upstream them in the future

Closes: [JSTZ-419](https://linear.app/tezos/issue/JSTZ-419/fork-deno-fetch), [JSTZ-420](https://linear.app/tezos/issue/JSTZ-420/removetraitify-non-compliant-functionality-from-deno-fetch)

# Description
Patch `deno_fetch` with our [deno_fetch fork](https://github.com/jstz-dev/deno/tree/ryan/patch-deno-fetch ) and registers fetch with dummy implementation

The fork creates a new `deno_fetch_base` crate which exposes the fetch ops as traits that we can implement in our code going forward. The set of changes made by the fork can be found at https://github.com/jstz-dev/deno/pull/1

This PR also fixes nix and wpt to make CI pass
* Patches vendored deno repo
* Moves existing patches to patches/octez
* Overwrites failed wpt tests messages that change between each run because they use non deterministic data like the current time or a random uuid

# Manually testing the PR


johnyob (approved):
LGTM 🚀 I'm able to build (within the Nix shell). But looks like CI is failing. I tried diagnosing it, but it seems to be an issue with how `crane` is vendoring in the git dependency.

PR #921, opened 2025-04-11
feat: tmp

ryutamago:
# Context

[url](https://linear.app/tezos/issue/JSTZ-470/implement-decimals-for-tez-amount)

# Description

# Manually testing the PR

### DIFF
--- a/crates/jstz_cli/src/bridge/deposit.rs
+++ b/crates/jstz_cli/src/bridge/deposit.rs
@@ -7,7 +7,7 @@ use crate::{
     sandbox::{
         assert_sandbox_running, JSTZD_SERVER_BASE_URL, SANDBOX_BOOTSTRAP_ACCOUNTS,
     },
-    utils::AddressOrAlias,
+    utils::{AddressOrAlias, Tez},
 };
 
 // hardcoding it here instead of importing from jstzd simply to avoid adding jstzd
@@ -17,7 +17,7 @@ const NATIVE_BRIDGE_ADDRESS: &str = "KT1GFiPkkTjd14oHe6MrBPiRh5djzRkVWcni";
 pub async fn exec(
     from: String,
     to: AddressOrAlias,
-    amount: u64,
+    amount: Tez,
     network: Option<NetworkName>,
 ) -> Result<()> {
     let cfg = Config::load().await?;
@@ -41,7 +41,7 @@ pub async fn exec(
     }
     let pkh = to_pkh.to_base58();
     debug!("resolved `to` -> {}", &pkh);
-
+    let amount = amount.to_f64();
     if use_sandbox {
         exec_sandbox(JSTZD_SERVER_BASE_URL, &from, &pkh, amount).await?;
     } else {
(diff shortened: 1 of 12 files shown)


alanmarkoTrilitech:
LGTM, but I would prefer there to be some integration tests that use decimal amounts, 0, negative amounts and similar, rather than manually testing from CLI. Currently it seems to test just the Tez struct.

ryutamago:
@alanmarkoTrilitech 
I think adding integration will require quite a lot more time as we have to mock the server etc and it's out of the scope of this PR. should be in a separate ticket for adding integraion test to cli

alanmarkoTrilitech:
> @alanmarkoTrilitech I think adding integration will require quite a lot more time as we have to mock the server etc and i think it is out of the scope of this PR. I will create a separate ticket for adding integration test to the cli commands

hm...I thought that we already can do tests like this. I guess a linear ticket is fine for now. However, it is currently a bit confusing cause it seems like at some places we used a u64 amount where it means Tez and in other parts outside of cli it used u64 which seemed to meant mutez. So I'm not sure if it guarantees correctness currently.

I've also noticed now that at some places it uses f64 for example in call_contract and exec_sandbox functions where f64 is used mainly for formatting to string. I would suggest getting rid of to_f64 function of Tez, and instead implementing Display trait. I think there might be some issues with doing self.0.to_string().parse::<f64>().unwrap() and then printing it for some floats like 0.999999 or similar, so it would be petter to just use printing of the Decimal or implementing a custom formatter if its not provided.

alanmarkoTrilitech:
> > I would suggest getting rid of to_f64 function of Tez, and instead implementing Display trait. I think there might be some issues with doing self.0.to_string().parse::().unwrap() and then printing it for some floats like 0.999999 or similar, so it would be petter to just use printing of the Decimal or implementing a custom formatter if its not provided.
> 
> is there any issue converting Tez to `0.999999f64`? the self.0 is max 6 decimal places and the Decimals libary convert to string properly.

it is ok that Decimals lib converts to string properly - I think it should be done through it (ignore the "or implementing a custom formatter if its not provided" from my previous comment - it is obviously implemented since you have the to_string function). However currently you convert from Decimal to string, then to f64, and then do &format!("{}", amount). when converting "0.999999" to float, it might get represented as something like 0.9999990...01 or similar in f64, which I think is best to avoid and not even provide a function for conversion to f64 in order to avoid bugs with it in the future.

crates/jstz_cli/src/bridge/deposit.rs
```diff
@@ -75,10 +76,11 @@ async fn exec_sandbox(
     jstzd_server_base_url: &str,
     from: &str,
     to_pkh: &str,
-    amount: u64,
+    amount: Tez,
 ) -> Result<()> {
     // go through jstzd server even when the sandbox is not in a container for simplicity
     let client = reqwest::Client::new();
+    let amount: f64 = amount.to_string().parse().unwrap();
```
alanmarkoTrilitech, line 83: can you please link the ticket for changing this also to Decimals in jstzd to this PR?

alanmarkoTrilitech (approved):

PR #919, opened 2025-04-10
feat(proto): using new jstz schema instead of tezos

alanmarkoTrilitech:
# Context
https://linear.app/tezos/issue/JSTZ-203/change-tezos-schemes-to-jstz-scheme

# Description
This PR changes the URI schema from tezos:// to jstz://.

# Manual testing
cargo nextest run --features skip-wpt --features skip-rollup-tests

PR #924, opened 2025-04-11
fix(doc): use jstz protocol instead of tezos

timothymcmackin:
https://github.com/jstz-dev/jstz/pull/919 changed the protocol from `tezos://` to `jstz://` but this topic was not updated.


zcabter (approved):

alanmarkoTrilitech (approved):
lgtm! thanks for noticing this

PR #927, opened 2025-04-14
add example smart function

ryutamago:
simply adds example large smart function that uses zod. 
This smart function was used in the demo.


alanmarkoTrilitech (approved):
lgtm! 🚢

PR #929, opened 2025-04-14
Fix Operation hash for large payload operation

ryutamago:
# Context
[link](https://linear.app/tezos/issue/JSTZ-458/store-recipt-in-the-original-hash)
Before this fix, when handling RLP operation in the protocol, the protocol used  the RLP operation hash. it should be linked to the orignal operation hash instead as RLP operation is an `intermediate opertion`

# Description

* Extended the RLP operation to include the hash of the original operation
* Modified the protocol to use the orginal operation hash instead of the RLP hash

# Manually testing the PR

add unit test.
```
cargo test -p jstz_proto
```


huancheng-trili:
Please fix your PR title.

PR #932, opened 2025-04-14
chore(ci): test CLI happy path with different machines

huancheng-trili:
# Context

Completes JSTZ-356.
[JSTZ-356](https://linear.app/tezos/issue/JSTZ-356/test-cli-happy-path-with-different-machines)

# Description

Test the built CLI binaries to be released as part of the release pipeline. The test is basically running a few simple commands in a clean environment to make sure that the binaries are statically linked and do not need additional libraries.

The pipeline is updated from
```
build mac         -> upload to release
build linux x64   -> upload to release
build linux arm64 -> upload to release
```
to
```
build mac         ─> test in mac          ─┐
build linux x64   ─> test in linux x64    ─┼─> upload all binaries to release
build linux arm64 ─> test in linux arm64  ─┘
```
# Manually testing the PR

[Test build](https://github.com/jstz-dev/jstz/actions/runs/14449726946)
[Uploaded artefacts](https://github.com/huancheng-trili/test-cli/releases/tag/v1.2.3)

### DIFF
--- a/.github/actions/test-cli/action.yml
+++ b/.github/actions/test-cli/action.yml
@@ -1,0 +1,26 @@
+name: Test basic CLI usability
+inputs:
+  artefact_name:
+    description: Name of the uploaded CLI artefact
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.artefact_name }}
+    - name: Test
+      shell: sh
+      run: |
+        check() {
+          if [[ "$1" != *"$2"* ]]; then
+            echo "check failed: '$1' should contain '$2'"
+            exit 1
+          fi
+        }
+
+        chmod +x ./jstz
+        export PATH=$PATH:$PWD
+        check "$(jstz account create test 2>&1)" "User created with address"
+        check "$(jstz account list 2>&1)" "test:"
+        check "$(jstz login test 2>&1)" "Logged in to account test with address"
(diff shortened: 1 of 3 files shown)


.github/actions/test-cli/action.yml
```diff
@@ -0,0 +1,19 @@
...
+      with:
+        name: ${{ inputs.artefact_name }}
+    - name: Test
+      shell: sh
+      run: |
+        chmod +x ./jstz
+        export PATH=$PATH:$PWD
+        jstz account create test
+        jstz account list
+        jstz login test
```
alanmarkoTrilitech, line 19: I guess this will have to change a bit because of XDG_CONFIG_HOME change. It might be good to also test that these commands are not silently failing, by for example checking some condition on the prefix of the last output or something similar. 

Otherwise it looks good to me.

huancheng-trili: What does `XDG_CONFIG_HOME` have to do with this? The commands will not be able to fail silently because any non-zero error code fails the step.

alanmarkoTrilitech: ah yes, I mistakenly read is modifying permissions of /.jstz folder instead of ./jstz.

Not sure though if currently in the CLI it always returns non-zero code or some of the errors can get suppressed in which case checking the output might be worth doing here.

alanmarkoTrilitech (approved):

.github/actions/test-cli/action.yml
huancheng-trili: I see what you mean. For now any error captured leads to a non-zero exit code. The main point of this check is to make sure that the executable can be executed. What happened previously was that the executable was not packaged properly and failed when it tried to load some dynamic libraries. This check ensures that such situation does not happen. We should be able to capture other errors in integration tests.

PR #935, opened 2025-04-15
fix(crypto): remove keypair_from_passphrase

huancheng-trili:
# Context

Part of JSTZ-484.
[JSTZ-484](https://linear.app/tezos/issue/JSTZ-484/update-account-creation)

In jstz, the term `passphrase` has always been what is referred to as mnemonic in other tools like octez-client and L1 wallets. This is misleading and the way CLI derives keys from passphrases is also problematic.

# Description

Replace `keypair_from_passphrase` with `keypair_from_mnemonic` to align key derivation with octez-client.

# Manually testing the PR

Added unit tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2913,6 +2913,7 @@ version = "0.1.0-alpha.0"
 dependencies = [
  "anyhow",
  "bincode 2.0.0-rc.3",
+ "bip39",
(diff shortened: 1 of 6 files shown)


alanmarkoTrilitech (approved):
lgtm!

PR #938, opened 2025-04-15
fix(cli): generate mnemonics instead of passphrases

huancheng-trili:
# Context

Completes JSTZ-484.
[JSTZ-484](https://linear.app/tezos/issue/JSTZ-484/update-account-creation)

In jstz, the term `passphrase` has always been what is referred to as mnemonic in other tools like octez-client and L1 wallets. This is misleading and the way CLI derives keys from passphrases is also problematic.

# Description

Generate mnemonics rather than passphrases during account creation. Passphrases become completely optional and the automatically generated mnemonics do their best to keep accounts unique.

# Manually testing the PR

Updated relevant tests.


crates/jstz_cli/src/account.rs
```diff
@@ -75,20 +75,15 @@ async fn create_account(alias: String, _passphrase: Option<String>) -> Result<()
...
+    let mnemonic = generate_mnemonic();
     let passphrase: String = Input::new()
-                .with_prompt("Enter the passphrase for the new account (or leave empty to generate a random one)")
-                .allow_empty(true)
-                .interact()?;
+        .with_prompt("Enter the passphrase for the new account or leave empty")
+        .allow_empty(true)
+        .interact()?;
 
-    let passphrase = if passphrase.is_empty() {
```
alanmarkoTrilitech, line 83: hm...it i a bit confusing to have both passphrase and mnemonic. what is the plan with the passphrase? Seems like the passphrase should just be removed as the way its used currently is similar to just adding an extra word to mnemonic.

There should probably be also some message asking user to write down/remember the mnemonic.

alanmarkoTrilitech: ok, thanks! are we tracking it somewhere that we need to decide on this? atleast a linear ticket would be good so that we don't forget

huancheng-trili: https://linear.app/tezos/issue/JSTZ-494/figure-out-how-to-align-differences-in-key-derivation

I'll mention this in the PR for importing accounts.

PR #939, opened 2025-04-15
chore(proto): remove outbox messasge id

zcabter:
# Context
Removes outbox message ID to avoid confusion among end users. 

Outbox message id field was intended to be used by users to track their bridge withdrawal. However, completing this feature requires a  lot more work that we cannot prioritize right now. So instead of allowing this to linger, we remove it and set the withdrawal request fields as part of its receipt.

Closes: [JSTZ-489](https://linear.app/tezos/issue/JSTZ-489/remove-outbox-message-id-temporarily)

# Description
* Remove outbox message id
* Add Fa withdraw request fields into the receipt
* Regenerate the new open api doc
* ticketer and proxy l1 contract were converted into smart function hash because it needs encoding and openapi derivations. This change should be safe since Jstz uses the same hashing scheme for contracts
* Added newtype `Kt1Hash` over `ContractKt1Hash` and make `SmartFunctionHash` wrap `Kt1Hash` instead. This allows open api to be generated with `Kt1Hash` schema instead, avoiding confusion.

# Manually testing the PR
Current tests passes

### DIFF
--- a/crates/jstz_proto/src/executor/fa_withdraw.rs
+++ b/crates/jstz_proto/src/executor/fa_withdraw.rs
@@ -51,7 +51,7 @@ pub struct TicketInfo {
 impl TicketInfo {
     pub(super) fn to_ticket(&self, amount: Amount) -> Result<Ticket> {
         FA2_1Ticket::new(
-            Contract::Originated(self.ticketer.0.clone()),
+            Contract::Originated(self.ticketer.clone().into()),
             MichelsonPair(
                 self.id.into(),
                 MichelsonOption(self.content.clone().map(MichelsonBytes)),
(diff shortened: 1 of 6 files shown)


crates/jstz_proto/src/executor/fa_withdraw.rs
```diff
@@ -27,30 +27,31 @@ use crate::{
...
     pub routing_info: RoutingInfo,
     pub ticket_info: TicketInfo,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema, Encode, Decode)]
 pub struct RoutingInfo {
     pub receiver: Address,
-    pub proxy_l1_contract: ContractKt1Hash,
+    pub proxy_l1_contract: SmartFunctionHash,
```
huancheng-trili, line 40: Is it possible to create an alias to `SmartFunctionHash` for this or L1 contract hash in general and make the alias appear in the API spec as a different type? I understand the point of having to make it serialisable, but I think using `SmartFunctionHash` directly here might lead to confusion, at least I was confused when I first read this.

huancheng-trili (commented):
So nothing actually changed in user experience right? They were supposed to be able to use outbox message ID to track withdrawals, but since this was never fully implemented, they couldn't do that anyway?

crates/jstz_proto/src/executor/fa_withdraw.rs
huancheng-trili: Then I guess we can use `Kt1Hash` here instead of `SmartFunctionHash`?

zcabter: Oops. This has now been fixed

huancheng-trili (approved):

PR #942, opened 2025-04-16
fix(examples): use jstz sdk from npm instead of local

zcabter:
# Context

Ensures the dashboard example does not break when main changes.

# Description
Updates `jstz_sdk` dependency in `examples/dashboard` to be pulled from npm registry

# Manually testing the PR


johnyob (approved):
🚀

PR #944, opened 2025-04-17
feat(crypto): get key pairs from secret keys

huancheng-trili:
# Context

Part of JSTZ-485.
[JSTZ-485](https://linear.app/tezos/issue/JSTZ-485/import-jstz-accounts-with-secret-keys-in-cli)

# Description

Get key pairs from secret key strings. Essentially,
```
"edsk..." --> (PublicKey, SecretKey)
```

# Manually testing the PR

Added one unit test.

### DIFF
--- a/crates/jstz_crypto/src/lib.rs
+++ b/crates/jstz_crypto/src/lib.rs
@@ -15,7 +15,7 @@ use tezos_crypto_rs::{hash::SeedEd25519, CryptoError};
 pub fn keypair_from_mnemonic(
     mnemonic: &str,
     passphrase: &str,
-) -> Result<(SecretKey, PublicKey)> {
+) -> Result<(PublicKey, SecretKey)> {
     let m = Mnemonic::parse_in(Language::English, mnemonic).map_err(|e| {
         CryptoError::InvalidKey {
             reason: format!("failed to parse mnemonic: {e}"),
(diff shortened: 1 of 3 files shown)


crates/jstz_crypto/src/lib.rs
```diff
@@ -26,6 +26,12 @@ pub fn keypair_from_mnemonic(
     Ok((SecretKey::Ed25519(sk), PublicKey::Ed25519(pk.into())))
 }
 
+pub fn keypair_from_secret_key(secret_key_str: &str) -> Result<(SecretKey, PublicKey)> {
+    let seed = SeedEd25519::from_base58_check(secret_key_str)?;
+    let (pk, sk) = seed.keypair()?;
+    Ok((SecretKey::Ed25519(sk), PublicKey::Ed25519(pk.into())))
```
alanmarkoTrilitech, line 32: it doesn't matter much, but since seed.keypair returns it in order - (pk, sk), it could be better if it followed the same style and also returned pair (pk,sk) instead of (sk,pk).

alanmarkoTrilitech (approved):
LGTM! 

if it doesn't require too many changes in other places, it would be good to change the pk,sk order before merging.

crates/jstz_crypto/src/lib.rs
huancheng-trili: Also updated `keypair_from_mnemonic`.

PR #945, opened 2025-04-17
feat(cli): import account from secret key

huancheng-trili:
# Context

Completes JSTZ-485.
[JSTZ-485](https://linear.app/tezos/issue/JSTZ-485/import-jstz-accounts-with-secret-keys-in-cli)

# Description

Allow users to import accounts from secret keys using CLI.

# Manually testing the PR

Added integration tests.


crates/jstz_cli/src/account.rs
```diff
@@ -34,6 +35,16 @@ impl User {
             public_key: pk,
         })
     }
+
+    pub fn from_secret_key(secret_key_str: &str) -> Result<Self> {
+        let (secret_key, public_key) = keypair_from_secret_key(secret_key_str)?;
```
alanmarkoTrilitech, line 40: similarly to previous PR, it would be good to have consistent, pk,sk order. for example bellow pk and address is before sk.

crates/jstz_cli/tests/account.rs
```diff
@@ -1,25 +1,28 @@
-use rexpect::session::spawn_command;
+use rexpect::session::{spawn_command, PtySession};
 use std::process::Command;
 use tempfile::TempDir;
 
-fn jstz_cmd() -> (Command, TempDir) {
-    let tmp_dir = TempDir::new().unwrap();
+fn jstz_cmd<'a, T>(args: T, home_dir: Option<TempDir>) -> (PtySession, TempDir)
```
alanmarkoTrilitech, line 5: should this be separate somewhere else? it seem like it isn't particularly related to account, and could be useful in other places too.

alanmarkoTrilitech (changes requested):
left some comments

crates/jstz_cli/src/account.rs
```diff
@@ -287,6 +331,14 @@ pub enum Command {
         #[arg(value_name = "ALIAS")]
         alias: String,
     },
+    /// Imports a user account from a secret key.
```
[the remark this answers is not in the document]
huancheng-trili: Personally I'm not a fan of these emojis. They look unprofessional.

crates/jstz_cli/tests/account.rs
huancheng-trili: This is purely a helper function for this test suite for now. I'll move this elsewhere when needed.

crates/jstz_cli/src/account.rs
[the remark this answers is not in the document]
alanmarkoTrilitech: We can discuss later whether to keep them or not, however it should be consistent. For now, I think it would be better to add an emoji there.

```suggestion
    /// 📥 Imports a user account from a secret key.
```

PR #949, opened 2025-04-17
refactor(proto): merge logic between `SmartFunction::call` and `smart_function::run::execute`

johnyob:
# Context

This PR merges the logic between `SmartFunction::call` and `smart_function::run::execute` (in theory they should do the same thing anyway). 

# Description

 **Dependencies**: #948 

This PR is split into two commits. 

The first commit handles the merging of the logic between `SmartFunction::call` and `smart_function::run::execute`. In some places, I discovered inconsistencies around how transfers were handled. This was discussed offline and the conclusion was that if a transfer fails, then it should revert the callee's transaction and return `Response.error()`. 

The second commit updates all the tests affected by the change of semantics in transfers. 

# Manually testing the PR

```
cargo nextest run -p jstz_proto
```

### DIFF
--- a/crates/jstz_proto/src/executor/smart_function.rs
+++ b/crates/jstz_proto/src/executor/smart_function.rs
@@ -341,7 +341,7 @@ pub struct HostScript;
 
 impl HostScript {
     pub fn run(
-        self_address: &impl Addressable,
+        source_address: &impl Addressable,
         request: &mut GcRefMut<'_, ErasedObject, Request>,
         context: &mut Context,
     ) -> JsResult<JsValue> {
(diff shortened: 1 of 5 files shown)


crates/jstz_proto/src/executor/smart_function.rs
```diff
@@ -336,54 +268,84 @@ impl Script {
...
-                content_type: None,
-            },
-            options,
-            context,
-        )
-    }
-
     pub fn run(
-        self_address: &SmartFunctionHash,
+        self_address: &impl Addressable,
```
ryutamago, line 344: nit:
`&SmartFunctionHash` as we cannot run user address

zcabter: I believe this is the source address?

zcabter: I've renamed this argument to source address

ryutamago (approved):

PR #951, opened 2025-04-17
fix(docs): comparison with smart contracts

timothymcmackin:
# Context

We got some feedback asking how smart functions were different from smart contracts.

# Description

These are a few ideas about how smart functions are similar to and different from smart contracts.

# Manually testing the PR

### DIFF
--- a/docs/functions/overview.md
+++ b/docs/functions/overview.md
@@ -89,7 +89,24 @@ Smart functions behave much like other serverless JavaScript/TypeScript applicat
 - Smart functions are currently restricted to 3915 bytes.
 - Smart functions can import and use packages, but they can use only certain JavaScript APIs, which limits the packages that they can use.
 - Smart functions run in an isolated context; each time a smart function runs, it is as if it is running for the first time.
-  Therefore, smart functions cannot store persistent data such as persistent global variables.
-  They must use the key-value store as described in [Storing data](/functions/data_storage).
-- Unlike smart contracts, smart functions have only a single entrypoint.
+Therefore, smart functions cannot store persistent data such as persistent global variables.
+They must use the key-value store as described in [Storing data](/functions/data_storage).
 <!-- https://huancheng-trili.github.io/jstz-api-coverage/ -->
+
+## Similarities and differences from smart contracts
+
+Smart functions are similar to Tezos [smart contracts](https://docs.tezos.com/smart-contracts) in many ways, including:
+
+- They are transparent, persistent, and cannot be changed after deployment
+- They have dedicated storage that only they can write to but is readable to outside clients
+- They can accept, store, and transfer tez between accounts
+
+Smart functions have these differences from smart contracts:
+
+- Unlike smart contracts, smart functions have only a single entrypoint
+- Smart contracts are compiled from higher-level languages to the Michelson stack-based language, while Jstz runs the TypeScript code of smart functions directly
+
+Here are some other things to know about smart functions and smart contracts:
+
+- Jstz does not support Tezos smart contracts, only Jstz smart functions
+- Smart functions cannot directly call smart contracts, and smart contracts cannot directly call smart functions


docs/functions/overview.md
```diff
@@ -89,7 +89,24 @@ Smart functions behave much like other serverless JavaScript/TypeScript applicat
...
+Smart functions are similar to Tezos [smart contracts](https://docs.tezos.com/smart-contracts) in many ways, including:
+
+- They are transparent, persistent, and cannot be changed after deployment
+- They have dedicated storage that only they can write to but is readable to outside clients
+- They can accept, store, and transfer tez between accounts
+
+Smart functions have these differences from smart contracts:
+
+- Unlike smart contracts, smart functions have only a single entrypoint
+- Smart contracts are compiled from higher-level languages to the Michelson stack-based language, while Jstz runs the TypeScript code of smart functions directly
```
zcabter, line 107: Actually, Jstz runs Javascript code. Typescript smart functions need to be transpiled into Javascript before deployment

docs/functions/overview.md
```diff
@@ -89,7 +89,22 @@ Smart functions behave much like other serverless JavaScript/TypeScript applicat
 - Smart functions are currently restricted to 3915 bytes.
 - Smart functions can import and use packages, but they can use only certain JavaScript APIs, which limits the packages that they can use.
 - Smart functions run in an isolated context; each time a smart function runs, it is as if it is running for the first time.
-  Therefore, smart functions cannot store persistent data such as persistent global variables.
-  They must use the key-value store as described in [Storing data](/functions/data_storage).
-- Unlike smart contracts, smart functions have only a single entrypoint.
+Therefore, smart functions cannot store persistent data such as persistent global variables.
```
huancheng-trili, line 92: I know this is not added in this PR, but I'd like to check with you if my thought is valid.

> smart functions cannot store persistent data such as persistent global variables

It's true that smart functions cannot keep global variables persistent, i.e. global variables simply get destroyed when a function execution ends. However, smart functions can store persistent data, e.g. by using the key-value store as you mentioned below. This line therefore reads a bit confusing to me. I think we need to rephrase this or clarify what "store persistent data" refers to. What do you think?

docs/functions/overview.md
```diff
@@ -89,7 +89,22 @@ Smart functions behave much like other serverless JavaScript/TypeScript applicat
...
+## Similarities and differences from smart contracts
+
+Smart functions are similar to Tezos [smart contracts](https://docs.tezos.com/smart-contracts) in many ways, including:
+
+- They are transparent, persistent, and cannot be changed after deployment
+- They have dedicated storage that only they can write to but is readable to outside clients
+- They can accept, store, and transfer tez between accounts
+
+One important difference between smart contracts and smart functions is that Tezos smart contracts are often written in a higher-level language and compiled to the Michelson stack-based language.
+By contrast, smart functions are transpiled from TypeScript to JavaScript and then Jstz runs the JavaScript code directly.
```
huancheng-trili, line 105: Smart functions can be written in plain JavaScript as well, so they aren't necessarily transpiled from TypeScript. This line now sort of indicates that smart functions need to be written in TypeScript.

docs/functions/overview.md
```diff
@@ -89,7 +89,22 @@ Smart functions behave much like other serverless JavaScript/TypeScript applicat
...
+
+Here are some other things to know about smart functions and smart contracts:
+
+- Jstz does not support Tezos smart contracts, only Jstz smart functions
```
huancheng-trili, line 109: Not sure if this will help or if our target users understand this (since they probably don't know much about blockchains,) but we might want to mention that smart contracts are for layer 1 and smart functions live in layer 2 on top of jstz, so smart functions are kind of in an isolated world and therefore have no idea of smart contracts, even if smart functions and smart contracts _can_ have same addresses.

We might want to clarify the last bit especially. Users might wonder "hey I have this smart contract address KT1..... Smart function addresses are also KT1.... so they are pretty much the same thing...?" Users might have similar questions for accounts later on, like "I have this tezos address tz1... so I can use it directly in jstz...?" or vice versa, and we might need to stress that these addresses are like in different worlds, even if they can be identical.

What do you think?

docs/functions/overview.md
timothymcmackin: Is this better?

> For this reason, smart functions themselves have no persistent state or memory; they must store all persistent data in the key-value store, as described in [Storing data](/functions/data_storage).

docs/functions/overview.md
timothymcmackin: > By contrast, smart functions are written in JavaScript or TypeScript, transpiled to JavaScript if necessary, and then Jstz runs the JavaScript code directly.

docs/functions/overview.md
timothymcmackin: How about this?

Here are some other things to know about smart functions and smart contracts:

- Despite having many similarities and similar-looking addresses, smart functions and smart contracts run in different environments:

  - Smart functions run only on Jstz, which is a layer that runs on top of Tezos layer 1
  - Smart contracts run directly on Tezos layer 1

- Therefore, smart functions and smart contracts cannot directly interact with each other; smart functions cannot directly call smart contracts, and smart contracts cannot directly call smart functions

huancheng-trili: I think this is much clearer!

docs/functions/overview.md
huancheng-trili: Sorry, one missing bit about ambiguity: "have no persistent state or memory" can be interpreted as

> have no persistent "state or memory"

or

> have no "persistent state" or memory

and the latter is obviously not what we mean. At this moment I don't have a better idea other than adding some more details like `i.e. global variables simply get destroyed when a function execution ends`, though this whole concept is trivial enough and we should keep it succinct.

timothymcmackin: How about this?

> For this reason, smart functions themselves have no persistent state; they must store all persistent data in the key-value store, as described in [Storing data](/functions/data_storage). Any global variables are reset each time they run, and if a smart function is called more than once at the same time, the instances do not have access to each other's state or variables.

huancheng-trili: Much better. Thank you!

huancheng-trili (approved):

PR #952, opened 2025-04-17
feat(docs): building and deploying topics

timothymcmackin:
# Context

We had an internal user hear that Jstz runs your TS code without having to build it so they tried to deploy a smart function without building it with the Jstz dependencies. 

# Description

Make it clear that you have to build smart functions with the Jstz dependencies before deploying them.

This PR also adds instructions for deploying smart functions to the sandbox; are we ready to provide instructions for deploying to any other network?

# Manually testing the PR

### DIFF
--- a/docs/functions/building.md
+++ b/docs/functions/building.md
@@ -1,0 +1,65 @@
+# Building smart functions
+
+Smart functions are written and run in ordinary TypeScript, but you must still build them with the Jstz dependencies before deploying them.
+
+For examples of smart function projects, see the examples folder: https://github.com/jstz-dev/jstz/tree/main/examples.
+
+Follow these steps to create and build a TypeScript project for your smart function:
+
+1. Create a `package.json` file appropriate for a TypeScript project.
+   You can use the `npm init` command, the `yarn init` command, or create a `package.json` file manually to describe the project.
+
+1. Add the `@jstz-dev/jstz` dependency and `esbuild` build dependencies.
+
+1. Add a build script that looks like this, with the name of your smart function source file in place of the variable `<SOURCE_FILE>`:
+
+   ```bash
+   esbuild <SOURCE_FILE> --bundle --format=esm --target=esnext --minify --outfile=dist/index.js
+   ```
+
+   Here is an example of a `package.json` file for a smart function project:
+
+   ```json
+   {
+     "name": "my-smart-function",
+     "authors": "",
+     "version": "0.0.0",
+     "main": "index.ts",
+     "dependencies": {
+       "@jstz-dev/jstz": "^0.0.0"
+     },
+     "devDependencies": {
+       "esbuild": "^0.20.2"
+     },
+     "scripts": {
+       "build": "esbuild index.ts --bundle --format=esm --target=esnext --minify --outfile=dist/index.js"
+     }
+   }
+   ```
+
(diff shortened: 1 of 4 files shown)


docs/functions/deploying.md
```diff
@@ -0,0 +1,34 @@
...
+Deploying a smart function to Jstz is different from deploying most web or JavaScript/TypeScript applications because you cannot change the code of a smart function or delete it after you deploy it.
+When you deploy a smart function, Jstz records code of the smart function in its ledger of transactions and there is no way to change or delete entries from the ledger.
+
+## Deploying to the local sandbox
+
+You can use the local sandbox to test smart functions in a simulated environment before you deploy them to a persistent Jstz network.
+
+1. Build the smart function as described in [Building smart functions](/functions/building).
+
+1. Start the Jstz sandbox by running `jstz sandbox --container start` or by starting the [`jstzd`](/jstzd) daemon directly.
```
huancheng-trili, line 12: Let's remove "or by starting the [`jstzd`](/jstzd) daemon directly" for now. Running jstzd directly requires all octez binaries in users' host environment, which means for users who did not build jstz from source/nix will run into problems.

Actually I'd like to hear your opinion here: should we make the containerised sandbox the default for now? That is, when users run `jstz sandbox start` we spin up a container rather than run the sandbox in the host environment. I think the team keeps the host environment sandbox mainly due to the dev setup, which is not what most actual users will have or be willing to configure. By making the containerised sandbox the default, it will make docker a required dependency, but I think it's also going to be less frustrating for users, e.g. no more errors from forgetting the `--container` flag, no more confusions about "why it doesn't work with --container," etc.

docs/functions/building.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. Install the dependencies by running `npm install` or `yarn install`.
+
+1. Add the code of your smart function to a TypeScript file in the project and make sure that the `build` script builds it.
+   For example, you can use the example smart function in [Smart functions](/functions/overview) or any of the smart functions in the [examples folder](https://github.com/jstz-dev/jstz/tree/main/examples).
+
+1. Build the smart function with the `build` script, either by running `npm run build` or `yarn build`.
+   Jstz builds the smart function with the necessary dependencies and writes the output to `dist/index.ts`.
+
+Now you can deploy the smart function to the Jstz network or the local sandbox.
```
huancheng-trili, line 64: I guess similar things are mentioned elsewhere as well, but do you think we should not mention "Jstz network" for now because we simply don't have that yet? I'm just thinking if users would wonder "hmm what's this network? ...oh there isn't such a thing yet, just the sandbox."

And as your comment in deploying.md, "Deploying to Jstz networks" is still a todo.

timothymcmackin: Got it, removed mention of Jstz networks for now.

docs/functions/deploying.md
huancheng-trili: I think we should only mention the container version for now to keep things simple. I'm really concerned about users seeing "To run the sandbox in a terminal window, run the command `jstz sandbox start`." and happily proceeding because "why not" just to be frustrated by something like "ERROR: program not found". I find this local setup quite unpleasant for now and definitely don't want users to get puzzled.

timothymcmackin: Ok, I kept it simple and included only the container version.

PR #953, opened 2025-04-17
fix(docs): organize architecture section

timothymcmackin:
# Context

I thought it would make sense to put things about the Jstz ecosystem in an "Architecture" section. More  things will come in this section as we provide more tools and features.

# Description

Screenshot 2025-04-17 at 3 42 20 PM

# Manually testing the PR


docs/architecture/architecture.md
huancheng-trili, line 1: This should be renamed to `docs/architecture/overview.md`?

huancheng-trili (commented):
I get the idea, but I'm not sure if `accounts` and `sandbox` should be part of "architecture" because architecture to me feels more about the core components that make jstz function. `sandbox` surely is not part of that, and `accounts` sound more like "concepts" in jstz. And the overview page does not mention any of these (not even asset bridge,) which feels a bit awkward bundling them.

docs/architecture/architecture.md
timothymcmackin: Moved things around -- let me know what you think.

huancheng-trili: Looks okay. Just one more thing about using the jstzd page as the "official" sandbox page: as I mentioned in previous PRs, jstzd itself as a sandbox is still not that usable because it requires octez binaries and some additional configuration and documentation for the latter is still not very clear, not to mention that the jstzd page itself needs polishing. Users who want to use jstzd also needs a bit more understanding of how things work with the blockchain components, e.g. how to make CLI point to the right ports, etc. Because of all of these, I'm concerned about making the jstzd page the "official" sandbox page for now.

We can still have a sandbox page, but I think it's better that we create a new one, move the sandbox section in the CLI page into the page and make it the main content, and downgrade the jstzd page into a section in the sandbox page. What do you think?

timothymcmackin: Ok, I addressed that in this PR. I left the sandbox info on the CLI page because it was just about the commands an those should still be on the CLI page.
https://github.com/jstz-dev/jstz/pull/969

PR #956, opened 2025-04-22
fix(nix): update nix flake

zcabter:
# Context
rust-analyzer has been painfully slow. It seems that updating to a later version improved things

# Description
* Ran `nix flake update` to update dependencies of our flakes - oxalica, numtides, nixpkgs and crane
* As a result, python39 was no longer found. Python39 is a WPT dependency but since we plan to remove WPT from local runs, I've removed python dependency altogether.

# Manually testing the PR
Check that vscode rust lsp is now blazing fast (at least for me!)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "crane": {
       "locked": {
-        "lastModified": 1739936662,
-        "narHash": "sha256-x4syUjNUuRblR07nDPeLDP7DpphaBVbUaSoeZkFbGSk=",
+        "lastModified": 1744386647,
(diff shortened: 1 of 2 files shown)


johnyob:
Tested locally, lgtm 🚀 

> As a result, python39 was no longer found. Python39 is a WPT dependency but since we plan to remove WPT from local runs, I've removed python dependency altogether.

Agree that at the moment we don't need `python39` but I find it concerning that we have no way to run WPT locally (meaning that should a regression occur, we cannot test/debug it locally). Perhaps we should add a chore task to investigate some way to locally run WPT?

johnyob (approved):

zcabter:
Let's put this on the back log until TezDev is over as it will wreck on build machines

PR #958, opened 2025-04-22
feat(cli): force account creation

huancheng-trili:
# Context

Completes JSTZ-486.
[JSTZ-486](https://linear.app/tezos/issue/JSTZ-486/overwrite-accounts-with-cli)

# Description

Add one flag `--force` to CLI account creation. This allows users to overwrite a known account alias. For example,
```sh
$ jstz account create foo
...
User created with address: tz1g3cNkswYeUCL9E5ZCuKhLiLSd1jb2r75d
$ jstz account create foo
✕  ERROR  The account 'foo' already exists. Please choose another name.
$ jstz account create foo --force
...
User created with address: tz1hqQUmReRZcWvKqNnCxdjPiaxja14xB5SQ
```

# Manually testing the PR

Updated the relevant integration test.

### DIFF
--- a/crates/jstz_cli/src/account.rs
+++ b/crates/jstz_cli/src/account.rs
@@ -64,15 +64,9 @@ async fn add_smart_function(alias: String, address: SmartFunctionHash) -> Result
     Ok(())
 }
 
-async fn create_account(alias: String) -> Result<()> {
+async fn create_account(alias: String, force: bool) -> Result<()> {
     let mut cfg = Config::load().await?;
-
-    if cfg.accounts.contains(&alias) {
-        bail_user_error!(
-            "The account '{}' already exists. Please choose another name.",
-            alias
-        );
-    }
+    check_alias_uniqueness(&cfg, &alias, force)?;
 
     let user = _create_account()?;
 
@@ -98,15 +92,19 @@ fn _create_account() -> Result<User> {
     Ok(user)
 }
 
-async fn import_account(alias: String, force: bool) -> Result<()> {
-    let mut cfg = Config::load().await?;
-
-    if cfg.accounts.contains(&alias) && !force {
+fn check_alias_uniqueness(cfg: &Config, alias: &str, force: bool) -> Result<()> {
+    if cfg.accounts.contains(alias) && !force {
         bail_user_error!(
             "The account '{}' already exists. Please choose another name or specify the `--force` flag to overwrite the account.",
             alias
         );
     }
+    Ok(())
+}
+
+async fn import_account(alias: String, force: bool) -> Result<()> {
+    let mut cfg = Config::load().await?;
+    check_alias_uniqueness(&cfg, &alias, force)?;
 
     let secret_key_str: String = Input::new()
         .with_prompt("Enter the secret key of your account")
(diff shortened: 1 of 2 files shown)


crates/jstz_cli/src/account.rs
```diff
@@ -64,10 +64,10 @@ async fn add_smart_function(alias: String, address: SmartFunctionHash) -> Result
     Ok(())
 }
 
-async fn create_account(alias: String) -> Result<()> {
+async fn create_account(alias: String, force: bool) -> Result<()> {
     let mut cfg = Config::load().await?;
 
-    if cfg.accounts.contains(&alias) {
+    if cfg.accounts.contains(&alias) && !force {
```
alanmarkoTrilitech, line 70: I think you can now better combine this with account importing - for example this if seems duplicated and has different messsages.

alanmarkoTrilitech (changes requested):
I am not very sure if this is actually needed - what are the examples when a user would want to do this? It essentially deletes the account and then creates it again and in delete we ask it whether the user is sure and he should confirm it by writing the account name again - I think we should do it here too then.

huancheng-trili:
By specifying `--force`, users have explicitly shown the intent of overwriting things and therefore there is no need for a prompt.

As of why a user would want this, perhaps for testing things? I have no clear idea, but this doesn't affect the existing functionality and therefore should be fine.

crates/jstz_cli/src/account.rs
huancheng-trili: Updated

huancheng-trili:
Not saying that their approach is "more correct," but if octez-client is doing this, I assume that this is not such a big problem. Yes, things _can_ happen and _can_ go wrong.

I'm totally fine closing this PR as it's not a big deal, but in any case I don't thinking throwing out a prompt when `--force` is specified is a good idea. It sounds like this:

* User: CLI, create account A _by force_.
* CLI: Okay, do you really want to create account A? It already exists.
* User: ...so I cannot do it by force actually.

alanmarkoTrilitech:
> > yes, it is a fairly common flag to speed things up, however usually you are not able to cause much damage or can at least revert the changes (although in a more complicated way).
> 
> I don't think its semantics includes reversibility. For example, force push irreversibly changes git history which can delete code forever. I think semantics of force is "Guard rails off, I trust you know what you're doing 🫡 ". I agree that things can go wrong that's why we protect against clumsiness by default but give seasoned users a powerful & efficient workflow.

Force push isn't going to immediately result in deletion of the code forever, there are some ways how to recover it for a few days. With the proposed change, it is possible to lose funds if someone accidentally rewrites his account and loses secret key by using -f flag. One way how to resolve this could be requiring an extra --yes flag on mainnet, or disabling the use of --force on mainnet or something similar. In any case even if we merge this as it is, I think it would be good to be at least aware of it somewhere and maybe creating an issue to discuss later.

zcabter (approved):

alanmarkoTrilitech:
I thought that we agreed on having there a check that prevents using --force on mainnet during weekly meeting.

alanmarkoTrilitech (approved):
lgtm! please create a ticket so that we don't forget retricting it on mainnet once we have profiles

huancheng-trili:
Created [JSTZ-505](https://linear.app/tezos/issue/JSTZ-505/restrict-cli-functionality-based-on-profiles) for adding restrictions and [JSTZ-504](https://linear.app/tezos/issue/JSTZ-504/create-cli-profiles-for-different-networks) for introducing profiles.

PR #962, opened 2025-04-23
feat(cli): overwrite smart function name

### LINKED ISSUE #926: `jstz deploy --name` needs a `-f` argument
We can set a local alias for a smart function by passing `--name` but if you try to use  the same name for a later deployment, the command fails. We need a `-f` argument to overwrite names.

huancheng-trili:
# Context

Completes JSTZ-498.
Closes #926.
[JSTZ-498](https://linear.app/tezos/issue/JSTZ-498/jstz-deploy-name-needs-a-f-argument)

# Description

Add one flag `--force` to `jstz deploy` to overwrite known smart functions. Example:
```
$ jstz deploy index.js --name foo
...
Smart function deployed by test_user at address: KT1...
Run with `jstz run jstz://KT1.../`
$ jstz deploy index.js --name foo
✕  ERROR  The smart function name 'foo' already exists. Please choose another name or specify the `--force` flag to overwrite the name.
$ jstz deploy index.js --name foo --force
...
Smart function deployed by test_user at address: KT1...
Run with `jstz run jstz://KT1.../`
```
# Manually testing the PR

Added one integration test.

### DIFF
--- a/crates/jstz_cli/src/deploy.rs
+++ b/crates/jstz_cli/src/deploy.rs
@@ -21,6 +21,7 @@ pub async fn exec(
     balance: Option<Tez>,
     name: Option<String>,
     network: Option<NetworkName>,
+    force: bool,
 ) -> Result<()> {
     let mut cfg = Config::load().await?;
     // Load sandbox if the selected network is Dev and sandbox is not already loaded
@@ -36,11 +37,11 @@ pub async fn exec(
         styles::command("jstz login")
     ))?;
 
-    // 1. Check if smart function account already exists
+    // 1. Check if the name already exists
     if let Some(name) = &name {
-        if cfg.accounts.contains(name) {
+        if cfg.accounts.contains(name) && !force {
             bail_user_error!(
-                "A user/smart function with the alias '{}' already exists.",
+                "The name '{}' is already used by another smart function or a user account. Please choose another name or specify the `--force` flag to overwrite the name.",
                 name
             );
         }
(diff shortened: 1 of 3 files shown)


crates/jstz_cli/tests/deploy.rs
```diff
@@ -0,0 +1,132 @@
...
+use tempfile::{NamedTempFile, TempDir};
+use tezos_crypto_rs::hash::ContractKt1Hash;
+use utils::jstz_cmd;
+
+#[path = "./utils.rs"]
+mod utils;
+
+fn config(endpoint: &str) -> serde_json::Value {
+    serde_json::json!({
+        "current_alias": "test_user",
```
alanmarkoTrilitech, line 16: might be good idea later to create some simple config creator in order to avoid always having to write these down.

huancheng-trili: Yup. I didn't want to reuse the config in account.rs because generalising that means that test suite needs to be updated for something it does not need, like `config("")` for an empty endpoint, and that brings more trouble than simply copying it for another test suite that needs more configuration. If this becomes duplicated too much, we can fix it.

crates/jstz_cli/src/deploy.rs
```diff
@@ -38,9 +39,9 @@ pub async fn exec(
 
     // 1. Check if smart function account already exists
     if let Some(name) = &name {
-        if cfg.accounts.contains(name) {
+        if cfg.accounts.contains(name) && !force {
             bail_user_error!(
-                "A user/smart function with the alias '{}' already exists.",
+                "The smart function name '{}' already exists. Please choose another name or specify the `--force` flag to overwrite the name.",
```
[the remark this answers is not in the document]
huancheng-trili: Updated this to `The name '{}' is already used by another smart function or a user account.`

PR #969, opened 2025-04-24
fix(docs): recast jstzd topic as sandbox topic

timothymcmackin:
Some clarifications for the docs on the sandbox. Is the info about running `jstzd` still necessary?


docs/sandbox.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Running the sandbox
+
+To start the sandbox, make sure that Docker is installed and then run this command to start the sandbox in a Docker container:
+
+```sh
+jstz sandbox --container start
+```
+
+If you see an error that says that the configuration file is improperly configured, delete the `~/.jstz/` folder and try to start the sandbox again.
```
huancheng-trili, line 32: I know this is just moved from elsewhere, but `~/.jstz/` needs to be updated to `~/.config/jstz/`.

timothymcmackin: Updated, thanks.

huancheng-trili (approved):

PR #970, opened 2025-04-24
fix(cli, proto): validate jstz scheme in run function

zcabter:
# Context
Jstz supports `jstz://` scheme in its URL but this has never been validated which is a bug. This PR fixes that.

# Description
* Validate jstz scheme in cli, `RunFunction` operation and in `SmartFunction.call`
* Add `ProcessSession` to cli test utils to reduce friction

# Manually testing the PR
```
cargo nextest run -p jstz_proto invalid_scheme
cargo nextest run -p jstz_cli invalid_scheme --test
```

### DIFF
--- a/crates/jstz_cli/src/run.rs
+++ b/crates/jstz_cli/src/run.rs
@@ -192,6 +192,8 @@ pub async fn exec(args: RunArgs) -> Result<()> {
         .parse()
         .expect("`url_object` is an invalid URL.");
 
+    validate_scheme(&url)?;
+
     let method = Method::from_str(&args.http_method)
         .map_err(|_| user_error!("Invalid HTTP method: {}", args.http_method))?;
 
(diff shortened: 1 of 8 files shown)


crates/jstz_cli/src/run.rs
```diff
@@ -300,6 +302,13 @@ pub async fn exec(args: RunArgs) -> Result<()> {
     Ok(())
 }
 
+fn validate_scheme(url: &Uri) -> Result<()> {
+    match url.scheme_str() {
+        Some("jstz") => Ok(()),
+        _ => bail!("URL scheme must be 'jstz'"),
+    }
+}
```
ryutamago, line 310: nit:
would be helpful to show the invalid schema
match url.scheme_str() {
    Some("jstz") => Ok(()),
    Some(s) => bail!("Invalid scheme '{}'. URL scheme must be 'jstz'", s),
    None => bail!("Missing scheme. URL must start with 'jstz://'"),
}

ryutamago (approved):
one unblocking comment. 
LGTM 🛳️

crates/jstz_cli/src/run.rs
zcabter: Done

PR #978, opened 2025-04-29
feat(proto): use Arc<Mutex<Transaction>>

alanmarkoTrilitech:
Please go the the `Preview` tab and select the appropriate sub-template:

- [🚧 Web/runtime APIs template](?expand=1&template=api_dev.md)
- [📚 Web/runtime APIs documentation template](?expand=1&template=api_doc.md)
- [⬜️ Blank template](?expand=1&template=default.md)


alanmarkoTrilitech:
> Would it not have been better to implement `Transaction` as
> 
> ```rust
> pub struct Transaction {
>   inner: Arc<Mutex<Inner>>
> }
> 
> pub struct Inner {
>   ...
> }
> ```
> 
> and then perform the locking in the various `Transaction` functions?

well yes, it is similar to what JsTransaction is doing, which could be further extended. This PR is however just the first step that gets rid of the static lifetimes. Later there will likely have to be more changes with how internally the Transaction works and most likely locking just on specific keys rather than on the whole Transaction object.

johnyob:
> well yes, it is similar to what JsTransaction is doing, which could be further extended. This PR is however just the first step that gets rid of the static lifetimes. Later there will likely have to be more changes with how internally the Transaction works and most likely locking just on specific keys rather than on the whole Transaction object.

I'm aware of the future changes. However, I think the approach taken by this PR will likely result in an equivalent amount of work in undo-ing the `Arc<Mutex<Transaction>>` changes, which could be avoided if the above pattern was used.

PR #982, opened 2025-05-01
feat(runtime): implement fetch response resource

ryutamago:
# Context

[task link](https://linear.app/tezos/issue/JSTZ-451/chunk-response-body-on-resourceread)

# Description

Implement fetch response resource. the logic for reading the stream stays the same as the original deno fetch implementation. The related deno PR is [here](https://github.com/jstz-dev/deno/pull/3)

# Manually testing the PR

cargo nextest run -p jstz_proto --features riscv --nocapture


crates/jstz_proto/src/runtime/v2/fetch.rs
```diff
@@ -330,12 +357,12 @@ async fn fetch_send(
         .expect("multiple op_fetch_send ongoing");
 
     let response = request.future.await;
-    let response_rid = state
-        .borrow_mut()
-        .resource_table
-        .add(FetchResponseResource {
-            body: RefCell::new(Some(response.body)),
-        });
+    let jstz_response = JstzResponse::new(response.body.into());
```
zcabter, line 360: IIUC, `Body` goes into `Bytes` which goes into `JstzResponse` to have `BytesStream`. Can we implement `BytesStream` on `Body` directly?

crates/jstz_proto/src/runtime/v2/fetch.rs
```diff
@@ -330,12 +357,12 @@ async fn fetch_send(
...
+    let body_size = jstz_response.body_size();
```
zcabter, line 361: You can implement `len()` on `Body` directly since the both `Vec<u8>` and `JsBuffer` as &[u8] has `len()`

crates/jstz_proto/src/runtime/v2/fetch.rs
```diff
@@ -1544,4 +1573,113 @@ mod test {
...
+            r#"
+            const call = async () => {{
+                const body = [1,2,3,4,5,6,7,8,9,10];
+                const expectedBytes = body.length;
+                let request = new Request("jstz://{func_addr}", {{
+                    method: "POST",
+                    body: new Uint8Array(body),
+                    headers: {{
+                        "content-type": "application/octet-stream",
+                    }}
```
zcabter, line 1622: Out of curiousity, this doesn't need to be an octet-stream right? The `response.body` will return a `ReadableStream` regardless

crates/jstz_proto/src/runtime/v2/fetch.rs
```diff
@@ -1544,4 +1573,113 @@ mod test {
...
+                }})
+                let response = await fetch(request);
+                const CHUNK_SIZE = 3;          
+                let actualBytes = 0;    
+                let count = 0;       
+                const reader = response.body.getReader({{ mode: "byob" }});
```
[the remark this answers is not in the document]
ryutamago: yes, extended the test for non-BYOB case

zcabter (approved):
LGTM, please update the branch

PR #986, opened 2025-05-13
fix(octez): increase operation ttl to to blocks (1m) and allow any origin on cors

zcabter:
# Context

Jstzd has 1 second block times and 8 second ttl which means forged operations have at most 8 seconds to be injected and applied. This window is too small for injecting operations from web wallets where usually a manual sign approval is required between forge and inject.

Web wallets require access to octez rpc to work correctly.

# Description
1. Increase `max_operations_time_to_live` to 60 in protocol parameter
2. Add `--cors-origin=*` arg to octez node run

# Manually testing the PR
Only config changes so existing test should pass. Test coverage reports nill unfortunately.
`make test` 

Tested manually by Bohdan


bohdan-trilitech (approved):
👌

PR #988, opened 2025-05-13
fix: use execute_latest_outbox_message with jstzd

huancheng-trili:
# Context

Realised that the helper script needs to be updated while trying out withdrawals.

# Description

Made the `execute_latest_outbox_message` helper script usable with jstzd. Also updated the script overall with a linter.

# Manually testing the PR

Tested the script with a local sandbox.
Shellcheck gives no warning.

### DIFF
--- a/scripts/execute_latest_outbox_message.sh
+++ b/scripts/execute_latest_outbox_message.sh
@@ -1,17 +1,23 @@
+#!/bin/bash
+set -e
+
 run() {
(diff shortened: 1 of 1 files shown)


zcabter (approved):
Its possible to configure the rollup node to automatically execute outbox messages. https://octez.tezos.com/docs/shell/smart_rollup_node.html#deploying-a-rollup-node 

This script might not be needed anymore although keeping it up to date is welcomed.

PR #989, opened 2025-05-15
refactor: create jstz_utils

huancheng-trili:
# Context

Part of JSTZ-544.
[JSTZ-544](https://linear.app/tezos/issue/JSTZ-544/create-sequencer-mode)

In preparation for #990.

# Description

Created a new crate `jstz_utils` that covers helper functions used across crates. For now only `poll` is moved to this crate because it's needed soon. The new crate is named `jstz_utils` instead of simply `utils` because each crate possibly has its own `utils` module and dealing with the naming conflict is a bit annoying.

# Manually testing the PR

CI passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3118,6 +3118,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "jstz_utils"
+version = "0.1.1-alpha.1"
+dependencies = [
+ "anyhow",
+ "reqwest",
+ "serde_json",
+ "tokio",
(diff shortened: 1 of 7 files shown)


alanmarkoTrilitech (approved):
LGTM!

you are planning to use poll in something related to sequencer right? In any case, it might be useful mentioning it in the PR description.

huancheng-trili:
Yes. This is for #990. I've updated the PR description.

PR #990, opened 2025-05-15
feat(jstz_node): create sequencer mode

huancheng-trili:
# Context

Completes JSTZ-544.
[JSTZ-544](https://linear.app/tezos/issue/JSTZ-544/create-sequencer-mode)

# Description

Added an enum `RunMode` to jstz node. This enum indicates the mode that jstz node will run in. `Default` means running jstz node as is and `Sequencer` means the sequencer behaviour will be activated. This mode is determined by the CLI argument `--mode` which defaults to `default`.

Also added one API route `GET /mode` which returns the node's mode for testing.

# Manually testing the PR

* Unit testing: added one test
* Integration testing: added one test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2968,6 +2968,7 @@ name = "jstz_node"
 version = "0.1.1-alpha.1"
 dependencies = [
  "anyhow",
+ "assert_cmd",
  "axum",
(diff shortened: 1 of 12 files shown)


ryutamago (approved):
LGTM

PR #993, opened 2025-05-15
feat(sequencer): insert operations to queue

huancheng-trili:
# Context

Part of JSTZ-546.
[JSTZ-546](https://linear.app/tezos/issue/JSTZ-546/set-up-operation-queue)

When a signed operation to be injected arrives at the sequencer, the operation should be placed in a queue and wait to be processed. If the queue still has space, the response code would be 200 as usual. If the queue is full, the response code would be 503.

# Description

Made the sequencer insert incoming signed operations into the queue.

# Manually testing the PR

* Unit testing: added tests for both the new error and the API handler.

### DIFF
--- a/crates/jstz_node/src/services/operations.rs
+++ b/crates/jstz_node/src/services/operations.rs
@@ -1,8 +1,12 @@
 use std::fs;
 use std::path;
+use std::sync::Arc;
+use std::sync::RwLock;
 
 use crate::config::KeyPair;
+use crate::sequencer::queue::OperationQueue;
 use crate::services::accounts::get_account_nonce;
+use crate::RunMode;
 
 use super::error::{ServiceError, ServiceResult};
 use super::{AppState, Service};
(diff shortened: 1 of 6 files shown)


crates/jstz_node/src/services/operations.rs
```diff
@@ -281,6 +344,48 @@ mod tests {
         size
     }
 
+    fn inject_operation_request() -> Request<Body> {
+        let dummy_op = SignedOperation::new(
+            Signature::Ed25519(Ed25519Signature::from_base58_check("edsigtbD6jADoivxf1iho6mDYPGiVvXw4Hnurn6VzDLG1boyMmmHEAykSrUJjJpvEsHHjQNvLWfm9PdyMBfJ8CX7jSEkh3yrB6m").unwrap().into()),
```
alanmarkoTrilitech, line 349: this is a second time I see creation of those dummy_ops so probably it should have a separate function somewhere

huancheng-trili: It's a bit difficult to do so because this kind of thing is only for tests and each test case might need different operations with specific details. In any case, for this PR, I realised that I can use some other helper functions defined before. This is now updated.

alanmarkoTrilitech (approved):
LGTM!

PR #1003, opened 2025-05-20
JSTZ-495-4: move host_script and top level fetch to v1

zcabter:
# Context

Depends on https://github.com/jstz-dev/jstz/pull/1002

# Description
Moves `host_script.rs` and `run_top_level_fetch` to v1

# Manually testing the PR
`make build && make test`


ryutamago (approved):
🧹

PR #1004, opened 2025-05-20
feat(sequencer): implement worker skeleton

huancheng-trili:
# Context

Part of JSTZ-551.

[JSTZ-551](https://linear.app/tezos/issue/JSTZ-551/create-worker)

# Description

Created a skeleton of the worker. The worker only gets created in the sequencer and runs in a separate thread and does its job periodically.

# Manually testing the PR

* Unit testing: added test cases

### DIFF
--- a/crates/jstz_node/src/sequencer/worker.rs
+++ b/crates/jstz_node/src/sequencer/worker.rs
@@ -1,0 +1,62 @@
+use std::{
+    sync::mpsc::{channel, Sender, TryRecvError},
+    thread::{self, spawn as spawn_thread, JoinHandle},
+    time::Duration,
+};
+
+pub struct Worker {
+    thread_kill_sig: Sender<()>,
+    inner: Option<JoinHandle<()>>,
+}
+
+impl Drop for Worker {
+    fn drop(&mut self) {
+        let _ = self.thread_kill_sig.send(());
+        if let Some(h) = self.inner.take() {
+            let _ = h.join();
+        }
+    }
+}
+
+pub fn spawn(#[cfg(test)] on_exit: impl FnOnce() + Send + 'static) -> Worker {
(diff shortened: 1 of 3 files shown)


ryutamago (approved):
LGTM.
one non-blocking comment is if the shutdown signal is meant to be used once it's probably better to use `oneshot` instead :)

crates/jstz_node/src/sequencer/worker.rs
```diff
@@ -0,0 +1,68 @@
...
+        Arc, Mutex,
+    },
+    thread::{self, spawn as spawn_thread, JoinHandle},
+    time::Duration,
+};
+
+#[derive(Clone)]
+pub struct Worker {
+    thread_kill_sig: Sender<()>,
+    inner: Arc<Mutex<Option<JoinHandle<()>>>>,
```
alanmarkoTrilitech, line 13: is there any reason why JoinHandle needs to be wrapped in Arc<Mutex<...>> ? Are you planning to share inner across multiple threads or something similar? My understanding is that there will be just one worker and you just want Mutex guarantees on the queue.

huancheng-trili: No. I was trying different ways of testing this and somehow added `Clone` to this struct, which led to `Arc<Mutex>`, and forgot to remove them afterwards. For now none of these are required and I have removed them, keeping `Option` only so that the handle can be consumed in `drop`.

ryutamago: > think it's rather rare that one would want to take down only the worker but keep other things in the sequencer running

I agree! my question was more to do with how we would handle a graceful shutdown in case the main server thread panics but i realized we can do that wihtout make it Clone by just checking the JoinHandle

PR #1005, opened 2025-05-20
feat(sequencer): make worker read from queue

huancheng-trili:
# Context

Part of JSTZ-551.

[JSTZ-551](https://linear.app/tezos/issue/JSTZ-551/create-worker)

# Description

Made the worker read from the operation queue. For now it does nothing to the operations because the runtime is not yet set up.

Also moved `dummy_op` to the sequencer module level so that it can be shared across test modules.

# Manually testing the PR

* Unit testing: added test cases

### DIFF
--- a/crates/jstz_node/src/sequencer/worker.rs
+++ b/crates/jstz_node/src/sequencer/worker.rs
@@ -1,8 +1,15 @@
 use std::{
-    sync::mpsc::{channel, Sender, TryRecvError},
+    sync::{
+        mpsc::{channel, Sender, TryRecvError},
+        Arc, RwLock,
+    },
     thread::{self, spawn as spawn_thread, JoinHandle},
     time::Duration,
 };
+
+use log::warn;
+
+use super::queue::OperationQueue;
 
 pub struct Worker {
     thread_kill_sig: Sender<()>,
@@ -18,12 +25,28 @@ impl Drop for Worker {
     }
 }
 
-pub fn spawn(#[cfg(test)] on_exit: impl FnOnce() + Send + 'static) -> Worker {
+pub fn spawn(
+    queue: Arc<RwLock<OperationQueue>>,
+    #[cfg(test)] on_exit: impl FnOnce() + Send + 'static,
+) -> Worker {
     let (thread_kill_sig, rx) = channel();
     Worker {
         thread_kill_sig,
         inner: Some(spawn_thread(move || loop {
-            thread::sleep(Duration::from_millis(500));
+            let v = {
+                match queue.write() {
+                    Ok(mut q) => q.pop(),
+                    Err(e) => {
+                        warn!("worker failed to read from queue: {e:?}");
+                        None
+                    }
+                }
+            };
+
+            match v {
+                Some(_m) => {}
+                None => thread::sleep(Duration::from_millis(100)),
+            }
 
             match rx.try_recv() {
                 Ok(_) | Err(TryRecvError::Disconnected) => {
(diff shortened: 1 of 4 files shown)


crates/jstz_node/src/sequencer/worker.rs
```diff
@@ -24,12 +28,28 @@ impl Drop for Worker {
...
+                    Err(e) => {
+                        warn!("worker failed to read from queue: {e:?}");
+                        None
+                    }
+                }
+            };
+
+            match v {
+                Some(_m) => {}
+                None => thread::sleep(Duration::from_millis(500)),
```
alanmarkoTrilitech, line 51: wouldn't it be better if the queue just sent a message that it isn't empty anymore, instead of busily spinning?

huancheng-trili: I guess you mean that the queue pretty much behaves like a channel. Yes, that would be better, but that means the queue needs more functionality and I think this is good enough for now. And I think for most queues, polling is how clients are expected to interact with them. I've reduced the waiting period here so that new requests don't need to wait for too long.

alanmarkoTrilitech: I'm not sure that the claim that for most queues polling is how clients are expected to interact with them is right. It seems to be a fairly standard Producer/Consumer channel which could be done with something like`crossbeam_channel`. Wouldn't it be worth looking into it more now before making further changes? 
You mention that 'that means queue needs more functionality' - what specifically do you mean?

huancheng-trili: Yes, the push and pubsub model are also common and I shouldn't have mentioned "for most queues." Still, I think for now it's okay to stick to this pull model mainly because the goal is not to have the best solution but a simple enough functioning implementation.

I mean to make the current implementation of the queue to support notifying consumers, we would need to add more things to this component, e.g. something like `crossbeam_channel` as you mentioned, and make sure that they work as expected. Since the current implementation of the queue is rather naive and will probably be rewritten significantly, I think we can leave these improvements to the next iteration when we are going to create a more solid queue.

alanmarkoTrilitech: I understand that the goal is to have  simple enough functioning implementation, but I think this way is just making it more complicated. For example, right now, you are using a channel already anyway to kill the thread. I think you can just use one channel to handle both kill and sending the operations, since send() and recv() on crossbeam_channel do basically the same thing as push_back/pop_front on the queue.

huancheng-trili: Complicated how? Using one channel for both operations and signals seems more complicated to me because these two different things get mixed up.

alanmarkoTrilitech: well it could simplify your loop if there was some enum distinguishing between Operation and Kill - you wouldn't have to both wait for 100ms and then also check if you haven't received kill signal.

alanmarkoTrilitech (approved):
I think single queue for messages and signals would overal fit this better, but I will approve so that it doesn't block as the kill signal is quite simple so far and it can be done later when improving the queue.

crates/jstz_node/src/sequencer/worker.rs
huancheng-trili: I agree. The ideal implementation in my opinion would be something like what `tokio::select!` enables:
```rs
wait! { // something similar to tokio::select
  queue.next() => { ... },
  kill_sig_ch.recv() => { ... }
}
```

Let's come back to this later on.

ryutamago (approved):

PR #1006, opened 2025-05-20
JSTZ-495-1: Move proto api except smart function to v1

zcabter:
# Context

Full changes refactor can be found here: https://github.com/jstz-dev/jstz/pull/1000

# Description
* Moves `jstz_proto::api` to `jstz_proto::runtime::v1::api` except `smart_function`. Moving smart function has a of spill over to other parts of the codebase so its refactor is done in the next PR.
* Excluded tps bench from makefile tests as it keep surfacing false positive issues with the kernel sdk

# Manually testing the PR
`make build && make test`


crates/jstz_proto/src/executor/smart_function/script.rs
```diff
@@ -43,25 +46,13 @@ pub fn register_web_apis(realm: &Realm, context: &mut Context) {
...
-        api::LedgerApi {
-            address: address.clone(),
-        },
-        context,
-    );
-    realm.register_api(
-        api::SmartFunctionApi {
+        ProtocolApi {
             address: address.clone(),
+            operation_hash: Blake2b::from(b"fake_op_hash".as_ref()),
```
ryutamago, line 55: i think we can pass in the operatoins_hash instad of the in the arg here and call 
`compute_seed`

crates/jstz_proto/src/runtime/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod v1;
+
+pub use v1::*;
```
huancheng-trili, line 3: No strong opinion here, but what about removing `pub use v1::*;` and keeping the `v1` namespace in imports? I guess you are thinking of a one-line change here when we switch from v1 to v2, but keeping `v1` explicitly in imports seems clearer and less error-prone to me. The downside is of course that `v1` in all these files will need to be updated when the switch happens, but I assume that a lot of things will need to be updated at that time anyway.

crates/jstz_proto/src/executor/smart_function/script.rs
zcabter: The bespoke register APIs get removed in https://github.com/jstz-dev/jstz/pull/1007/files#diff-bd20826dfb0e6b376089b6196bfaf92a3bfc2332b0520aa2494b5354dd97e40f and replaced with the `ProtocolApi` and `WebApi` types

crates/jstz_proto/src/runtime/mod.rs
zcabter: We will need to support both runtimes in the interim through feature flags. Having a single place for the feature flags will reduce noise. The goal is for things exposed through `crate::runtime` to be  runtime agnostic (still a bit of work to get there) then we wouldn't need to change the rest of the code base when the runtimes switch. 

To improve clarity of the api that the new runtime needs to support, I've explicit exported them rather than using `*`. You can see the full set of apis in the [last pr](https://github.com/jstz-dev/jstz/pull/1003/files) at crates/jstz_proto/src/runtime/mod.rs

ryutamago (approved):

PR #1007, opened 2025-05-20
JSTZ-495-2: Move proto smart function api to v1

zcabter:
# Context

Full changes refactor can be found here: https://github.com/jstz-dev/jstz/pull/1000

Depends on https://github.com/jstz-dev/jstz/pull/1006

# Description
Moves smart function api, fetch handler and script to `runtime::v1`

# Manually testing the PR
`make build && make test`


crates/jstz_proto/src/executor/mod.rs
```diff
@@ -19,7 +18,7 @@ fn execute_operation_inner(
     hrt: &mut impl HostRuntime,
     tx: &mut Transaction,
     op: Operation,
-    ticketer: &ContractKt1Hash,
+    _ticketer: &ContractKt1Hash,
```
ryutamago, line 21: why is this underscored when it's being used?

ryutamago (approved):
🧹

PR #1010, opened 2025-05-22
feat(octez): add rollup history mode

ryutamago:
# Context

part of the [inbox processing task](https://linear.app/tezos/issue/JSTZ-549/read-rollup-inbox-messages)

[task link](https://linear.app/tezos/issue/JSTZ-573/make-rollup-history-mode-configurable)

Sequencer needs to read inbox messages. One way to do that is via the [rollup node rpc](https://gitlab.com/tezos/tezos/-/blob/master/docs/api/alpha-smart-rollup-node-openapi.json) via /global/block/[level]

# Description

* Extended the octez rollup to take an optional `history_mode` that defaults to `Full` (which is the default mode for the binary anyway) 
* make the default sandbox config inside jstzd to use archive mode

# Manually testing the PR

changed the existing test to ensure that histroy mode is passed in correctly


crates/jstzd/src/config.rs
```diff
@@ -129,6 +130,7 @@ pub(crate) async fn build_config(
         SmartRollupHash::from_base58_check(JSTZ_ROLLUP_ADDRESS).unwrap(),
         ROLLUP_OPERATOR_ACCOUNT_ALIAS.to_string(),
         jstz_rollup_path::kernel_installer_path(),
+        Some(DEFAULT_ROLLUP_HISTORY_MODE),
```
huancheng-trili, line 133: This can simply be `None`?

ryutamago: yea i want the default to be archive for now as this is practicallly our sandbox settings.

ryutamago: oops fixed, thanks! 😅

huancheng-trili (approved):

PR #1013, opened 2025-05-23
fix(runtime): runtime use single threaded platform

zcabter:
# Context
Without single threaded, v8 will try to spawn and manage multiple threads, then errors when trying to switch between them due to the Drop hack.

# Description

# Manually testing the PR
CI should pass now


ryutamago (approved):
🛳️

PR #1014, opened 2025-05-23
feat(sequencer): implement runtime db

huancheng-trili:
# Context

Part of JSTZ-575.
[JSTZ-575](https://linear.app/tezos/issue/JSTZ-575/implement-host)

A new host is needed for the sequencer, at least when the runtime is going to run natively. This host stores data in a data store accessible from the exterior.

# Description

Implemented a sqlite wrapper for the new host and some basic methods that perform queries required by the host. Using a sqlite db should be fine for jstz because fortunately jstz only uses some host methods that do not really depend on the tree-based WASM rollup data store.

Some additional helper functions like `exec_read` are implemented to make reusing connections, e.g. for transactions, easier.

# Manually testing the PR

* Unit testing: added tests

### DIFF
--- a/crates/jstz_node/src/sequencer/db.rs
+++ b/crates/jstz_node/src/sequencer/db.rs
@@ -1,0 +1,312 @@
+#![allow(unused)]
+use std::{fs, path::PathBuf};
+
+use anyhow::Result;
+use r2d2::{Pool, PooledConnection};
+use r2d2_sqlite::SqliteConnectionManager;
+use rusqlite::params;
+use rusqlite::Connection;
+use rusqlite::OptionalExtension;
+
+pub type SqliteConnectionPool = Pool<SqliteConnectionManager>;
+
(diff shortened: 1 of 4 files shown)


crates/jstz_node/src/sequencer/db.rs
```diff
@@ -0,0 +1,302 @@
...
+        assert!(!db.key_exists("foo").unwrap());
+
+        let conn = db.connection().unwrap();
+        insert(&conn, "foo", "bar");
+
+        assert!(db.key_exists("foo").unwrap());
+    }
+
+    #[test]
+    fn count_subkeys() {
```
ryutamago, line 212: just a q but we don't have a nested subkeys in jstz now right?

huancheng-trili: We do. Users determine how they want to store values with KV, which means we can have nested subkeys.

PR #1015, opened 2025-05-27
Add octez rollup port forwarding on 18751 for sandbox container

Frank-en-stein:
### Context
When running `jstz sandbox --container start`, the octez rollup node was not accessible from the host machine because its port was randomly assigned and not forwarded through docker. This prevented developers from directly interacting with the rollup node during local development and testing.

### Description
This PR adds port forwarding for the octez rollup node on port 18751 when running the sandbox in a Docker container. The changes include:

- Added a new constant `SANDBOX_OCTEZ_ROLLUP_RPC_PORT` (18751) in the sandbox constants
- Updated the Docker container configuration to expose and forward port 18751
- Modified the jstzd configuration to explicitly set the octez rollup RPC endpoint to use port 18751 with an option to use config to set rollup port
- Updated tests to include the new rollup port configuration

### Testing Instructions
1. Build the CLI with the changes: `make build-cli`
2. Start the sandbox: `./target/release/jstz sandbox --container start`
3. Verify port forwarding: `docker port jstz-sandbox` should show `18751/tcp -> 0.0.0.0:18751`
4. Test rollup accessibility: `curl http://localhost:18751/global/block/head` should return a response

### Patch Coverage
The changes include test updates to verify the configuration is properly set.


crates/octez/src/async/rollup.rs
```diff
@@ -32,27 +32,27 @@ pub enum RollupDataDir {
     Temp,
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Default)]
 pub struct OctezRollupConfigBuilder {
```
zcabter, line 36: I think we should add `#[serde(default)]` at the struct level. Additionally, instead of deriving `Default`, implement it so that mandatory fields can have their custom defaults.

PR #1017, opened 2025-05-27
feat(core): add global host provider

zcabter:
# Context
Introduce two related structs `GlobalHostProvider` and `Host` 

`GlobalHostProvider` is responsible for creating new instances of `Host` and updating the host runtime instance referenced by each Host every time kernel entry is called. In this setup, `Host` can be thought of as holding a reference to a global cell of `Runtime` that only the `GlobalHostProvider` can update.

This is required because suspended futures can outlive the entry call in which it was created, therefore the host runtime that it points to may no longer be valid on resumption. This setup is better than a naive thread local host runtime because it allows localizing tests while still providing all the benefits of thread local host runtime.

# Description
* Introduce `GlobalHostProvider` and `Host`
* Adds thread local `GlobalHostProvider`
* Update kernel functions to use `Host`

# Manual testing
`make build && make test-unit`
`cargo nextest run -p jstz_core host`


alanmarkoTrilitech:
It would be good to add a bit more context to this. It isn't clear to me how can the suspended Futures outlive the kernel entry call - I thought that they get created within the kernel entry, and even though they get suspended it still stays within the same entry call. More context or a small example would help here. Also some more details about why you chose UnsafeCell and why the JsHostRuntime has to have 'static lifetime would be helpful.

zcabter:
> I thought that they get created within the kernel entry, and even though they get suspended it still stays within the same entry call
No, the futures live on the heap, not the stack of calling function. They will live as long as there is a reference to it. In the final PR, I introduce a thread local pool that will be the scheduler that ultimately manages these futures and outlive the kernel entry call https://github.com/jstz-dev/jstz/pull/1019. 

> details about why you chose UnsafeCell
We need some form of getting an interior mutable reference because Arc prevents getting a mut.
`Cell` cannot get a reference to the value
`RefCell` (and other safe interior mutability containers)  imposes performance penalty from unnecessary runtime checks - we know its always safe to get a `&` or `&mut` of `Runtime` since its a stateless FFI API to the host
`UnsafeCell` lets us get a mutable reference and doesn't imposes unnecessary performance penalty

> why the JsHostRuntime has to have 'static lifetime 
'static lifetime on Runtime is required because we need to hold the reference in global state

alanmarkoTrilitech:
> No, the futures live on the heap, not the stack of calling function. They will live as long as there is a reference to it. In the final PR, I introduce a thread local pool that will be the scheduler that ultimately manages these futures and outlive the kernel entry call https://github.com/jstz-dev/jstz/pull/1019.

I'm slightly confused by why we even need thread_local here - its basically supposed to create a separate copy for each thread, right? Does it apply to tokio futures in a single threaded runtime? 
Kernel entry is kind of like main, right? I imagine you could just create the local pool inside of kernel entry call and it gets dropped once kernel entry finishes.

zcabter:
> Kernel entry is kind of like main, right? I imagine you could just create the local pool inside of kernel entry call and it gets dropped once kernel entry finishes

It doesn't act like `main` in WASM. But it is indeed main in RISCV (I had to double check this with RISCV team as the kernel SDK APIs would need changing to support that). So indeed, this change is not required

PR #1019, opened 2025-05-27
refactor(kernel): asynchronify kernel entry

zcabter:
# Context
Ultimately, the run function needs to be async. We need a local pool executor for single threaded driving of futures. The executor should make as much progress as possible until it cannot, at which point it returns control to the PVM node for more input. 

Part of https://linear.app/tezos/issue/JSTZ-593/asychronify-kernel-entry
Closes https://linear.app/tezos/issue/JSTZ-595/asychronify-top-level-kernel-entry

# Description
* Refactors main kernel entry logic into an `async fn`
* Adds a thread local storage `LocalPool` that will be responsible for driving futures 

# Manually testing the PR
`make build && make test`

### DIFF
--- a/crates/jstz_kernel/src/lib.rs
+++ b/crates/jstz_kernel/src/lib.rs
@@ -1,5 +1,6 @@
 use std::cell::RefCell;
 
+use futures::{executor::LocalPool, task::LocalSpawnExt};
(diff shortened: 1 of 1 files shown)


zcabter:
No longer needed

PR #1020, opened 2025-05-27
refactor(proto): asynchronify proto smart_function::run::execute

zcabter:
# Context
Reduces review burden by asynchronifying `smart_function::run::execute` only

# Description
* Make execute async
* Adds temporary `block_on` because the callers of `execute` are not yet async

# Manually testing the PR
`make build && make test`


alanmarkoTrilitech (approved):
LGTM, what do you plan to replace block_on with with once its completed?

Also, during the discussion with Arthur, he didn't like the async/non-deterministic part, but I don't remember what the conclusion was there. was it  that is fine if used just for the oracles?

zcabter:
> > There's actually no non-determinism but we need to communicate that better.
> 
> how so? As I understand it, before executing the smart function, we don't know what the result of the oracle (which can change arbitrarily) would be as this gets requested after the sf execution is started. You can verify it and get always the same result since the result of the oracle call gets recorded and stays the same. However, I think there is a subtle difference that normally you would get the result of the oracle call before you call a contract, here the call of the contract and call of the oracle get merged into one - making it not possible to determine the outcome of the sf call before its actually called.

The oracle result can't change arbitrarily. Its a lazy value that takes some time to be "initialize". SF can await on the intiailization and when that happens, the function suspends. But once the value arrives, it is stable for the rest of the execution. While its suspended, any keys that it read or written to are locked. Successive Sfs will have to wait for those keys in the queue.

Response to an oracle fetch is a jstz operation, it must come through the inbox therefore from Jstz's and PVM's perspective, has a stable sequence within the inbox. 

What you're probably confused about is the fact the SF's can suspend for an arbitrary length of time, and during this time, it possible for successive SF's to use keys that our SF would eventually use. Let's dissect
1. Arbitrary length of time - In general, the blockchain doesn't known when an operation will arrive, but once included in a block, it will have a stable sequence number.  From the user's perspective, this is non-deterministic - When the user inject an op,  they don't not know if their operation will be included at `i`, `i+n` or `i-n`. In fact, it might not be included at all! This is true for Jstz operations too, and oracle response is no different. However, once included, it is deterministic both from the user's and PVM's perspective
2. Eventual keys are altered during suspension - In standard synchronous blockchain environment, we can achieve suspension w 3 operations. Op 1 uses key X, Op 2, uses key Y, and Op 3 uses key Y. Op 1 and Op 3 are simulating our suspension while Op 2 just happens to pass through.
```
Op 1          | Op 2      | Op 3
start          |               |
use X        |               |
ends         |               |
                  | start      |
                  | use    Y |
                  | ends     |
                  |               | starts
                  |               | use y
                  |               | ends
```
This is completely valid. However, there are some friction points 
1. Op 1 must implement a lock over X. Successive ops that try to use X while the key is locked will fail! They will need to keep trying until the lock is freed
2. To achieve atomicity, if Op 3 fails, then it must manually revert changes made by Op1
3. Lastly, from the user's perspective, when Op 1 succeeds, it doesn't mean the operation is complete. They must also wait for Op3 to complete. 
IMO these is terrible UX and DX if trying to implement a town crier like oracle .

In Jstz, we remove these friction points by suspending operations and locking on keys
```
Op 1          | Op 2      | 
start          |               |
locks X      |               |
suspends  |               |
                  | start      |
                  | locks  Y |
                  | ends     |
resumes   |               |
locks y      |               | 
ends         |               | 
```
1. The smart function does not need to think about locking, it is implicit . If another op tries to use the locked key, it gets queued instead of failing
2. Atomicity is naturally follows the operation that it began with
3. Successful operation really means a successful operation

Note that the SF could just lock all its keys up front, in what case would they not want to? The reason for asynchrony is alawys throughput. Hence, the devs would lock only keys as they go fo higher SF throughput, ultimately a better experience for their users

alanmarkoTrilitech:
thanks for the detailed explanation!

I guess my concern could be better described as lack of predictability of the result before its executed, rather than non-determinism.

In the first "standard synchronous blockchain environment",  the operations are executed separately 1,3 by the user, 2, by the oracle. In the second design, where 1 and 3 are merged as just op1, the successful execution of the  second operation result necessarily in the execution of third operation (or the second part of op1 once since its merged).
Previously the user could choose whether to actually finish and execute op3, but now no such choice exist. 
The loss of the option to opt out of op3 seems likely to give the oracle more power, more incentives to cheat and adds more options for manipulation and MEV extraction. 

Yes, I understand that from the user point of view, blockchain doesn't know when the operation will arrive. However, you can in theory monitor the mempool and decide whether to sign your transaction based on the current state. You can also simulate what will be the outcome of executing a certain set of transactions ordered in some way. Here you lose this possibility, since the full set of transactions only becomes known later. Maybe less of a concern right now, but I can imagine this could also have some issues with interoperability later.

Also it is a bit unclear to me how does the failing of transactions and "If another op tries to use the locked key, it gets queued instead of failing" works. Do you pay gas when you execute half of the transaction and it gets queued?

zcabter:
> Previously the user could choose whether to actually finish and execute op3, but now no such choice exist.

Ah no, this is not true in town crier-like oracles. The response operation is a signed and injected from the town crier operator, the user does not choose. Thus, it is super important to ensure the town crier operators/network is decentralized, deterministic and fair.

> The loss of the option to opt out of op3 seems likely to give the oracle more power, more incentives to cheat and add more options for manipulation and MEV extraction.

This is not the way to solve MEV and mepool-inpection based attacks. These are solved by supporting private operations.

> Yes, I understand that from the user point of view, blockchain doesn't know when the operation will arrive. However, you can in theory monitor the mempool and decide whether to sign your transaction based on the current state. You can also simulate what will be the outcome of executing a certain set of transactions ordered in some way. Here you lose this possibility, since the full set of transactions only becomes known later. Maybe less of a concern right now, but I can imagine this could also have some issues with interoperability later.

Simulating the operation does not guarantee that the order will be the final one.This is true on all blockchains today. A consequence of this can be seen in decentralized exchange implementations, where they cannot promise an exact exchange rate but only an approximate one with a maximum slippage the user is willing to tolerate, which is the reason why MEV exists in the first place.


> Also it is a bit unclear to me how does the failing of transactions and "If another op tries to use the locked key, it gets queued instead of failing" works. Do you pay gas when you execute half of the transaction and it gets queued? 

Yes. A suspended operation will have to hold some state, the remaining gas is one of them.

> I'm not saying that it can't work this way, but it seems quite different from the usual way how its done.

The usual way can already be done. In fact, even town crier oracles can be done in Jstz today - it would just be not as nice of an experience compared to a solution that is deeply integrated into the protocol.

PR #1021, opened 2025-05-27
feat(sequencer): set up old runtime

huancheng-trili:
# Context

Part of JSTZ-576.
[JSTZ-576](https://linear.app/tezos/issue/JSTZ-576/use-boa-runtime)

The sequencer needs a runtime to execute incoming operations.

# Description

Set up a runtime environment to be used by the sequencer. The setup is pretty much the same as the kernel.

# Manually testing the PR

* Unit testing: added tests

### DIFF
--- a/crates/jstz_node/src/sequencer/runtime.rs
+++ b/crates/jstz_node/src/sequencer/runtime.rs
@@ -1,0 +1,197 @@
+#![allow(unused)]
+use anyhow::{anyhow, bail, Context};
+use jstz_core::kv::{Storage, Transaction};
+use jstz_crypto::{
+    hash::Hash, public_key::PublicKey, smart_function_hash::SmartFunctionHash,
+};
+use jstz_proto::{executor::execute_operation, operation::SignedOperation};
+use tezos_smart_rollup::{
+    prelude::{debug_msg, Runtime},
+    storage::path::RefPath,
+};
+
+use super::db::Db;
+
+const TICKETER: RefPath = RefPath::assert_from(b"/ticketer");
+const INJECTOR: RefPath = RefPath::assert_from(b"/injector");
+
+pub fn init_host(db: Db) -> anyhow::Result<impl Runtime> {
+    let mut host = crate::sequencer::host::Host::new(db);
+    let ticketer = SmartFunctionHash::from_base58("KT1HbQepzV1nVGg8QVznG7z4RcHseD5kwqBn")
+        .context("failed to parse ticketer address")?;
+
+    host.store_write_all(
+        &TICKETER,
+        &bincode::encode_to_vec(&ticketer, bincode::config::legacy())
+            .context("failed to encode ticketer")?,
+    )
+    .context("failed to write ticketer to host store")?;
+
+    let injector =
+        PublicKey::from_base58("edpkuBknW28nW72KG6RoHtYW7p12T6GKc7nAbwYX5m8Wd9sDVC9yav")
+            .context("failed to parse injector public key")?;
+
+    host.store_write_all(
+        &INJECTOR,
+        &bincode::encode_to_vec(&injector, bincode::config::legacy())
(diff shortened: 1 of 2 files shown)


crates/jstz_node/src/sequencer/runtime.rs
```diff
@@ -0,0 +1,197 @@
...
+    prelude::{debug_msg, Runtime},
+    storage::path::RefPath,
+};
+
+use super::db::Db;
+
+const TICKETER: RefPath = RefPath::assert_from(b"/ticketer");
+const INJECTOR: RefPath = RefPath::assert_from(b"/injector");
+
+pub fn init_host(db: Db) -> anyhow::Result<impl Runtime> {
```
ryutamago, line 18: are we planning to read the TICKETER and INJECTOR from the kv store in the rollup?

huancheng-trili: What do you mean? This is just how the kernel works now right?

crates/jstz_node/src/sequencer/runtime.rs
```diff
@@ -0,0 +1,197 @@
...
+
+fn read_ticketer(rt: &impl Runtime) -> Option<SmartFunctionHash> {
+    Storage::get(rt, &TICKETER).ok()?
+}
+
+fn read_injector(rt: &impl Runtime) -> Option<PublicKey> {
+    Storage::get(rt, &INJECTOR).ok()?
+}
+
+pub fn process_message(rt: &mut impl Runtime, op: SignedOperation) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
huancheng-trili: Indeed, but since `ExternalOperation` is not yet ready (the queue needs to be updated as well,) I'm using `SignedOperation` here for now.

crates/jstz_node/src/sequencer/runtime.rs
ryutamago: the kernel is initialized with the TICKETER and INJECTOR at the build time. Similarily the sequencer should be initlized with these two values too. 

i meant when we start the sequencer, i think it should fetch from the kernel to ensure that sequencer is using the same values.

huancheng-trili: Um I don't think that's correct. These values are inserted when the kernel is installed. The kernel entrypoint then reads these values from KV at execution time. For the sequencer, these values are inserted to runtime store in `init_host` as initialisation.

Because these values are actually kept in KV, the sequencer does not need to read these values from the rollup. It will compare the state of its store and that of the rollup store and therefore ensure that these values match.

ryutamago: > The kernel entrypoint then reads these values from storage at execution time
yes this is my understanding as well. 
> For the sequencer, these values are inserted to runtime store in init_host as initialisation.
Yes that makes sense. so during the runtime the values are avaiable jsut like how it works in the kernel. 

i should have clarified the questions more clearly, where are we getting these values from? will we have a sequencer's config file or fetch the values from the kernel's kv storage and then pass init the init_host ?.

huancheng-trili: I haven't thought through, but what I have in mind now is that this will probably be hardcoded and built into the executable at build time. This value does not change often and pretty much always gets updated when the rollup kernel is updated, and hence I don't think it needs to be "config-file-configurable".

The main challenge is in protocol upgrades, i.e. when we somehow want to use a new value in a new version, the sequencer needs to cope with this change, e.g. it needs to know when to use the new value, assuming that there is a good way of packing the new value into it. If I recall correctly, protocol upgrade in etherlink is triggered by some special inbox messages. Using this approach would make it much simpler. The rollup kernel always lags behind the sequencer, so there shouldn't be any conflict when it's its turn to pick up the new value.

PR #1024, opened 2025-05-27
feat(core): implicit locking

alanmarkoTrilitech:
# Context
[JSTZ-610](https://linear.app/tezos/issue/JSTZ-610/implement-guarded-for-implicit-locking)
# Description
In order to avoid having to lock/unlock transaction object once we use Arc<Mutex<...>> for sharing it across threads, this PR implements Guarded and GuardedMut. Guarded replaces references returned by Transaction and contains a reference counter of the guard and a pointer to a value from the Transaction object. We can use deref or deref_mut, to get references to the value within an unsafe block, whose safety is guaranteed since the lock stays alive while Guarded is alive. Lock gets dropped automatically, when all Guarded/GuardedMut get dropped. Additionally Transaction is modified to be a wrapper for sharing Arc<Mutex<Inner>> with Inner being the actual implementation. It also holds a non-owning reference to the lock which it can clone when its functions return Guarded/GuardedMut.

This is a temporary solution, as in later stages, Transaction will be rewritten to lock on keys rather than having to keep the lock of the whole object.
# Manual testing
```make```

### DIFF
--- a/crates/jstz_cli/src/repl/debug_api/account.rs
+++ b/crates/jstz_cli/src/repl/debug_api/account.rs
@@ -70,7 +70,7 @@ impl AccountApi {
         let address = try_parse_smart_function_address(account.as_str())?;
 
         runtime::with_js_hrt_and_tx(|hrt, tx| -> JsResult<JsValue> {
-            match Account::function_code(hrt.deref(), tx, &address)?.as_str() {
+            match Account::function_code(hrt.deref(), tx, &address)?.deref() {
                 "" => Ok(JsValue::null()),
                 value => Ok(JsValue::String(value.to_string().into())),
             }
--- a/crates/jstz_core/src/kv/transaction.rs
+++ b/crates/jstz_core/src/kv/transaction.rs
@@ -237,11 +237,18 @@ impl InnerTransaction {
         rt: &impl Runtime,
         key: Key,
     ) -> Result<Option<&SnapshotValue>> {
-        if let Some(&snapshot_idx) = self.lookup_map.get(&key).and_then(|h| h.last()) {
-            return Ok(self.stack[snapshot_idx].lookup(&key));
+        if let Some(&snapshot_idx) =
+            self.lookup_map.get(&key).and_then(|history| history.last())
+        {
+            let snapshot = &self.stack[snapshot_idx];
+
+            return Ok(snapshot.lookup(&key));
         }
+
         if let Some(value) = Storage::get::<V>(rt, &key)? {
+            // TODO: This clone is probably not necessary
             self.current_snapshot_insert(key.clone(), SnapshotValue::new(value))?;
+
             self.current_snapshot_lookup(&key)
         } else {
             Ok(None)
@@ -253,17 +260,18 @@ impl InnerTransaction {
         rt: &impl Runtime,
         key: Key,
     ) -> Result<Option<&mut SnapshotValue>> {
-        if let Some(&snapshot_idx) = self.lookup_map.get(&key).and_then(|h| h.last()) {
-            let snap = &self.stack[snapshot_idx];
-            if snap.lookup(&key).is_some() {
-                self.current_snapshot_insert(
-                    key.clone(),
-                    snap.lookup(&key).unwrap().clone(),
-                )?;
-                return self.current_snapshot_lookup_mut(&key);
+        if let Some(&snapshot_idx) =
+            self.lookup_map.get(&key).and_then(|history| history.last())
+        {
+            let snapshot = &self.stack[snapshot_idx];
+
+            if let Some(value) = snapshot.lookup(&key) {
+                self.current_snapshot_insert(key.clone(), value.clone())?;
+                self.current_snapshot_lookup_mut(&key)
+            } else {
+                Ok(None)
             }
-        }
-        if let Some(value) = Storage::get::<V>(rt, &key)? {
+        } else if let Some(value) = Storage::get::<V>(rt, &key)? {
             self.current_snapshot_insert(key.clone(), SnapshotValue::new(value))?;
             self.current_snapshot_lookup_mut(&key)
         } else {
@@ -271,11 +279,97 @@ impl InnerTransaction {
         }
     }
 
-    fn contains_key(&self, rt: &impl Runtime, key: &Key) -> Result<bool> {
-        if let Some(&idx) = self.lookup_map.get(key).and_then(|h| h.last()) {
-            return Ok(self.stack[idx].contains_key(key));
+    /// Returns `true` if the key-value store contains a key-value pair for the
+    /// specified key.
+    pub fn contains_key(&self, rt: &impl Runtime, key: &Key) -> Result<bool> {
+        if let Some(&context_idx) =
+            self.lookup_map.get(key).and_then(|history| history.last())
+        {
+            let context = &self.stack[context_idx];
+
+            return Ok(context.contains_key(key));
         }
+
         Storage::contains_key(rt, key)
+    }
+
+    fn queue_outbox_message(
+        &mut self,
+        rt: &mut impl Runtime,
+        message: OutboxMessage,
+    ) -> Result<()> {
+        if self.persistent_outbox.len(rt)? + self.snapshot_outbox_len + 1
+            > self.persistent_outbox.max(rt)?
+        {
+            Err(OutboxError::OutboxQueueFull)?;
+        }
+        let current_outbox_queue = self.current_snapshot()?.outbox_queue_mut();
+        current_outbox_queue.queue_message(message);
+        self.snapshot_outbox_len += 1;
+        Ok(())
+    }
+
+    /// Begin a transaction.
+    fn begin(&mut self) {
+        self.stack.push(Snapshot::default())
+    }
+
+    /// Commit a transaction.
+    fn commit(&mut self, rt: &mut impl Runtime) -> Result<()> {
+        let curr_ctxt = self.stack.pop().ok_or(KvError::TransactionStackEmpty)?;
+
+        // Following the `.pop`, `prev_idx` is the index of prev_idx (if it exists)
+        let prev_idx = self.current_snapshot_idx();
+
+        if let Some(prev_ctxt) = self.stack.last_mut() {
+            // TODO: These clones are probably uncessary since the entry of btree will always be occupied.
+            for key in curr_ctxt.remove_edits {
+                self.lookup_map.rollback(&key)?;
+                self.lookup_map.update(key.clone(), prev_idx);
+                prev_ctxt.remove(key);
+            }
+
+            for (key, value) in curr_ctxt.insert_edits {
+                self.lookup_map.rollback(&key)?;
+                self.lookup_map.update(key.clone(), prev_idx);
+                prev_ctxt.insert(key, value);
+            }
+
+            prev_ctxt.outbox_queue.extend(curr_ctxt.outbox_queue);
+        } else {
+            for key in &curr_ctxt.remove_edits {
+                Storage::remove(rt, key)?
+            }
+
+            for (key, value) in curr_ctxt.insert_edits {
+                Storage::insert(rt, &key, value.0.as_ref())?
+            }
+
+            flush(rt, &mut self.persistent_outbox, curr_ctxt.outbox_queue)?;
+            self.snapshot_outbox_len = 0;
+
+            // Update lookup map
+            self.lookup_map.clear()
+        }
+
+        Ok(())
+    }
+
+    /// Rollback a transaction.
+    fn rollback(&mut self) -> Result<()> {
+        let curr_ctxt = self.stack.pop().ok_or(KvError::TransactionStackEmpty)?;
+
+        // SAFETY: The set of keys between removal edits and insertion edits are disjoint, meaning no
+        // `lookup_map` entries will be rolledback more than once
+        for key in &curr_ctxt.remove_edits {
+            self.lookup_map.rollback(key)?;
+        }
+
+        for key in curr_ctxt.insert_edits.keys() {
+            self.lookup_map.rollback(key)?
+        }
+
+        Ok(())
     }
 }
 
@@ -294,7 +388,7 @@ impl Transaction {
     /// Begin a transaction.
     pub fn begin(&self) {
         let rc = self.acquire_guard().expect("mutex poisoned");
-        rc.borrow_mut().stack.push(Snapshot::default());
+        rc.borrow_mut().begin();
     }
 
     /// Commit a transaction.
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

File: crates/jstz_core/src/kv/transaction.rs [hunk 1]
```diff
@@ -240,66 +252,154 @@ impl Transaction {
...
-    pub fn get_mut<V: Value>(
-        &mut self,
+impl Transaction {
+    fn acquire_guard(&self) -> Option<RcGuardInner> {
+        if let Some(rc) = self.guard.borrow().upgrade() {
+            return Some(rc);
+        }
+
+        let guard: GuardInner = Arc::clone(&self.inner).try_lock_arc()?;
+        let rc: RcGuardInner = Rc::new(RefCell::new(guard));
```
zcabter, line 288: Are the type annotations required? Can the compiler infer the types on its own?

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -240,66 +252,154 @@ impl Transaction {
...
+        Storage::contains_key(rt, key)
     }
+}
 
-    /// Returns a mutable reference to the value corresponding to the key in the
-    /// key-value store if it exists.
-    pub fn get_mut<V: Value>(
-        &mut self,
+impl Transaction {
+    fn acquire_guard(&self) -> Option<RcGuardInner> {
```
zcabter, line 282: Should this return Result<RcGuardInner> instead?

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -56,14 +61,33 @@ pub type Key = OwnedPath;
 struct LookupMap(BTreeMap<Key, Vec<usize>>);
 
 #[derive(Debug, Default)]
-pub struct Transaction {
+pub struct Inner {
```
zcabter, line 64: A slightly more descriptive name would be `InnerTransaction`

crates/jstz_cli/src/repl/debug_api/account.rs
```diff
@@ -70,7 +70,7 @@ impl AccountApi {
         let address = try_parse_smart_function_address(account.as_str())?;
 
         runtime::with_js_hrt_and_tx(|hrt, tx| -> JsResult<JsValue> {
-            match Account::function_code(hrt.deref(), tx, &address)? {
+            match Account::function_code(hrt.deref(), tx, &address)?.deref() {
```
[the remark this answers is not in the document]
alanmarkoTrilitech: Yes, but it also doesn't make much deference if ```.deref()``` stays there - the line would have to be changed anyway.

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -471,116 +511,205 @@ impl<'a, V> Entry<'a, V> {
...
+// Guarded<T>/GuardedMut<T> – keeps the mutex guard alive while reference is alive
+
+pub struct Guarded<'a, T: ?Sized + 'a> {
+    value: *const T,
+    pub _guard: RcGuardInner,
+    _marker: PhantomData<&'a ()>,
+}
+
+impl<'a, T: ?Sized + 'a> Guarded<'a, T> {
+    pub fn new<'b>(guard: RcGuardInner, value: &T) -> Guarded<'b, T> {
```
zcabter, line 638: Probably don't need `<'b>`. Return type can be `Guarded<'a, T>`. mb 😓

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -240,66 +253,154 @@ impl Transaction {
...
-            let snapshot = &self.stack[snapshot_idx];
-
-            if let Some(value) = snapshot.lookup(&key) {
-                self.current_snapshot_insert(key.clone(), value.clone())?;
-                self.current_snapshot_lookup_mut(&key)
-            } else {
-                Ok(None)
+        if let Some(&snapshot_idx) = self.lookup_map.get(&key).and_then(|h| h.last()) {
+            let snap = &self.stack[snapshot_idx];
+            if snap.lookup(&key).is_some() {
```
zcabter, line 258: Isn't `if let Some(value)  =  snap.lookup(&key)` more appropriate here?

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -240,66 +253,154 @@ impl Transaction {
...
+                Storage::remove(rt, key)?;
+            }
+            for (key, val) in curr.insert_edits {
+                Storage::insert(rt, &key, val.0.as_ref())?;
+            }
+            flush(rt, &mut inner.persistent_outbox, curr.outbox_queue)?;
+            inner.snapshot_outbox_len = 0;
+            inner.lookup_map.clear();
+        }
+        Ok(())
```
zcabter, line 331: Its much cleaner if the logic for `InnerTransaction` and managing the guard were separated, even as an interim solution. Doing so would make github actually show what really changed, rather than a lot of false positive lines. Additionally, it would make it easier to identify what is crucial when we do the Transaction component. 

I'm not sure if you attempted this and found a lot of issues with but it seems that majority of these false positive lines are simply acquring inner, operating on it, then wrapping the result in guarded

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -56,14 +61,34 @@ pub type Key = OwnedPath;
...
-pub struct Transaction {
+pub struct InnerTransaction {
     // A stack of transactional snapshots
     stack: Vec<Snapshot>,
     lookup_map: LookupMap,
     persistent_outbox: PersistentOutboxQueue,
     snapshot_outbox_len: u32,
 }
 
+type GuardInner = ArcMutexGuard<RawMutex, InnerTransaction>;
```
zcabter, line 72:
```suggestion
type GuardInner = ArcMutexGuard<RawMutex, Pin<Box<InnerTransaction>>>;
```

This should be pinned since `Guarded` depends on the fact that pointers to `InnerTransaction` will alway be valid. Mb for missing that one

zcabter: InnerTransaction also needs a field like `_pin: PhantomPinned` to prevent rust from auto deriving `Unpin`

alanmarkoTrilitech: yeah, unfortunately after briefly trying it out, that seems to be the case...

crates/jstz_core/src/kv/transaction.rs
alanmarkoTrilitech: changed it

crates/jstz_core/src/kv/transaction.rs
alanmarkoTrilitech: updated

crates/jstz_core/src/kv/transaction.rs
alanmarkoTrilitech: ok, that seems to be right

zcabter (approved):
Missing test the Guarded components are acceptable in this case because we'll be refactoring Transaction soon. 

LGTM otherwise

zcabter:
> I just realised that there might be an issue with Clone:
> 
> ```
> impl Clone for Transaction {
>     fn clone(&self) -> Self {
>         Self {
>             inner: self.inner.clone(),
>             guard: RefCell::new(Weak::new()),
>         }
>     }
> }
> ```
> 
> I think it should take a weak pointer to a guard if it exists

Actually, I just remembered why we disallow cloning a Weak ref to guard - its to prevent multiple `mut` references

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -415,6 +375,141 @@ where {
...
+    /// Returns the given key's corresponding entry in the transactional
+    /// snapshot for in-place manipulation.
+    pub fn entry<'a, 'b, V>(
+        &'a mut self,
+        rt: &impl Runtime,
+        key: Key,
+    ) -> Result<Entry<'b, V>>
+    where
+        V: Value,
+        'a: 'b,
```
zcabter, line 473: Its morally incorrect to decouple the lifetime of the `Transaction` and `Entry`
```suggestion
    pub fn entry<'a, V>(
        &'a mut self,
        rt: &impl Runtime,
        key: Key,
    ) -> Result<Entry<'a, V>>
    where
        V: Value
```

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -471,116 +560,178 @@ impl<'a, V> Entry<'a, V> {
...
+
+impl<'a, T: ?Sized + 'a> std::ops::Deref for GuardedMut<'a, T> {
+    type Target = T;
+    fn deref(&self) -> &'a Self::Target {
+        unsafe { &*self.value }
+    }
+}
+
+impl<'a, T: ?Sized + 'a> std::ops::DerefMut for GuardedMut<'a, T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
```
zcabter, line 733: Although the `'a` lifetime should transitively applied to `&mut Self::Target` since `'a` lives at least as long as `GuardedMut`, I think its better to make the relationship explicit.

```suggestion
    fn deref_mut(&mut self) -> &'a mut Self::Target {
```

alanmarkoTrilitech:
> > > > I just realised that there might be an issue with Clone:
> > > > ```
> > > > impl Clone for Transaction {
> > > >     fn clone(&self) -> Self {
> > > >         Self {
> > > >             inner: self.inner.clone(),
> > > >             guard: RefCell::new(Weak::new()),
> > > >         }
> > > >     }
> > > > }
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > I think it should take a weak pointer to a guard if it exists
> > > 
> > > 
> > > Actually, I just remembered why I disallowed cloning a Weak ref to guard - its to prevent multiple `mut` references. What was the issue with not propagating the weak ref on clone?
> > 
> > 
> > There wasn't any issue, as in practice, I think there aren't be many cases when we would want to clone the Transaction while its locked and instead take reference to an existing one. However, if it happens, I think the correct behaviour is for it to use the same Guard, rather than trying to lock again.
> > How does it help with preventing multiple mut references? As in the code would deadlock or fail in that case, so you wouldn't use it? You can still have multiple `mut` references then though, if not locking.
> 
> The main issue with cloning the guard when Transaction is cloned is that the original lifetime is lost and that it would be possible to break Rust's lifetime rules of single mut ref of multiple immut refs
> 
> > You can still have multiple `mut` references
> 
> EDIT: I think the current implementation makes impossible to hold multiple `mut` references. But with your `.clone` change, it will be possible to hold mut and immut refs at the same time.
> 
> If the abstraction is implemented correct, this shouldn't be possible
> 
> ```
> struct TestValue(i32);
> 
> // code that initializes `TestValue` at `path`...
> 
> let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> let mut b = tx.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> b.deref_mut().0 = 1;
> let x = a;
> ```
> 
> However, if you allowed the guard to be cloned, then it is possible to hold an immutable reference that gets mutated elsewhere. The code below will return a lock poisoned error in the current implementation but allowable with `.clone` change.
> 
> ```
> let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> let mut tx2 = tx.clone();
> let mut b = tx2.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> b.deref_mut().0 = 1;
> let x = a;
> ```

you are right about the second part. It is however possible to do
```
let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
let mut b = tx.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
b.deref_mut().0 = 1;
let x = a;
```
regardless of whether it does the cloning of the guard or not.

crates/jstz_core/src/kv/transaction.rs
```diff
@@ -415,6 +375,141 @@ where {
...
+            Some(entry) => {
+                let value = entry.as_ref()?;
+                Ok(Some(Guarded::new(rc.clone(), value)))
+            }
+            None => Ok(None),
+        }
+    }
+
+    pub fn get_mut<'a, V: Value>(
+        &'a self,
```
zcabter, line 430: Bug 🐛 
```suggestion
        &'a mut self,
```

zcabter:
> > > > > I just realised that there might be an issue with Clone:
> > > > > ```
> > > > > impl Clone for Transaction {
> > > > >     fn clone(&self) -> Self {
> > > > >         Self {
> > > > >             inner: self.inner.clone(),
> > > > >             guard: RefCell::new(Weak::new()),
> > > > >         }
> > > > >     }
> > > > > }
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > I think it should take a weak pointer to a guard if it exists
> > > > 
> > > > 
> > > > Actually, I just remembered why I disallowed cloning a Weak ref to guard - its to prevent multiple `mut` references. What was the issue with not propagating the weak ref on clone?
> > > 
> > > 
> > > There wasn't any issue, as in practice, I think there aren't be many cases when we would want to clone the Transaction while its locked and instead take reference to an existing one. However, if it happens, I think the correct behaviour is for it to use the same Guard, rather than trying to lock again.
> > > How does it help with preventing multiple mut references? As in the code would deadlock or fail in that case, so you wouldn't use it? You can still have multiple `mut` references then though, if not locking.
> > 
> > 
> > The main issue with cloning the guard when Transaction is cloned is that the original lifetime is lost and that it would be possible to break Rust's lifetime rules of single mut ref of multiple immut refs
> > > You can still have multiple `mut` references
> > 
> > 
> > EDIT: I think the current implementation makes impossible to hold multiple `mut` references. But with your `.clone` change, it will be possible to hold mut and immut refs at the same time.
> > If the abstraction is implemented correct, this shouldn't be possible
> > ```
> > struct TestValue(i32);
> > 
> > // code that initializes `TestValue` at `path`...
> > 
> > let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> > let mut b = tx.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> > b.deref_mut().0 = 1;
> > let x = a;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, if you allowed the guard to be cloned, then it is possible to hold an immutable reference that gets mutated elsewhere. The code below will return a lock poisoned error in the current implementation but allowable with `.clone` change.
> > ```
> > let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> > let mut tx2 = tx.clone();
> > let mut b = tx2.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> > b.deref_mut().0 = 1;
> > let x = a;
> > ```
> 
> you are right about the second part. It is however possible to do
> 
> ```
> let a = tx.get::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> let mut b = tx.get_mut::<TestValue>(hrt, path.clone()).unwrap().unwrap();
> b.deref_mut().0 = 1;
> let x = a;
> ```
> 
> regardless of whether it does the cloning of the guard or not.

Yes, because there's a bug

PR #1030, opened 2025-05-29
feat(core): remove static lifetime from transactions

alanmarkoTrilitech:
# Context
[JSTZ-375](https://linear.app/tezos/issue/JSTZ-375/replace-andmut-transaction-with-arcmutextransaction)

# Description
This PR remove static lifetime from transaction, which is now possible as the Transaction was moved within Arc<Mutex<Inner>>.

# Testing
```make```


zcabter:
LGTM, please rebase and fix conflicts

zcabter (commented):
We should only clone transaction when we need `Send` it . This will reduce the surface for unintended locking

alanmarkoTrilitech:
Done

PR #1037, opened 2025-06-02
feat(sequencer): read smart function code from db

huancheng-trili:
# Context

Part of JSTZ-601.
[JSTZ-601](https://linear.app/tezos/issue/JSTZ-601/expose-runtime-db-with-jstz-node-api)

Jstz node API needs to return data from runtime db when jstz node runs in sequencer mode.

# Description

Updated the handler of `/accounts/<address>/code` such that it returns data from the runtime db when jstz node runs in sequencer mode.

# Manually testing the PR

* Unit testing: added tests

### DIFF
--- a/crates/jstz_node/src/services/accounts.rs
+++ b/crates/jstz_node/src/services/accounts.rs
@@ -148,11 +148,16 @@ async fn get_nonce(
     )
 )]
 async fn get_code(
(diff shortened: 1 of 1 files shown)


ryutamago (approved):
👏

PR #1040, opened 2025-06-03
feat(sequencer): implement reveal_preimage

huancheng-trili:
# Context

Part of JSTZ-592.
[JSTZ-592](https://linear.app/tezos/issue/JSTZ-592/deal-with-large-payload)

# Description

Implemented host function `reveal_preimage` so that the large payload operations work in the sequencer.

`reveal_preimage` reads from disk the preimage files that are stored in the given rollup preimage directory, which means it can be slow when a lot of preimage files are involved in one operation, but that's the same for the rollup kernel anyway.

# Manually testing the PR

* Unit testing: added tests


ryutamago (approved):
🔥

PR #1043, opened 2025-06-03
feat(proto): balance endpoint in host_script

alanmarkoTrilitech:
# Context
# Description
This PR adds balance endpoint in host script.
# Manual testing
```make```


alanmarkoTrilitech:
Ok, I thought you were showing that I should change the match clause in fetch function of v1 runtime. I'll make the changes in the v2. However, this is a working version in v1. Can't it be merged? I know its not particularly useful as we move to v2, but it also doesn't harm anything and is already implemented.

alanmarkoTrilitech:
can't I just merge it instead as its a working version for v1? are we going to completely get rid of v1 in a very near future?

PR #1045, opened 2025-06-04
feat(proto): add event publisher

zcabter:
# Context
Closes https://linear.app/tezos/issue/JSTZ-529/implement-event-and-oraclerequestevent 

The `EventPublisher` is responsible for publishing events to the kernel debug log. In the form `[JSTZ]<json oracle request>`. Json was chosen as its the most convenient string encoding format which is required by the kernel debug log

# Description
1. Adds `enum Event`, OracleRequestEvent` and `EventPublisher`
2.`ByteString` are essentially a sequence of latin-1 bytes that deno uses for fast serialization. So we de/serialize those as json byte arrays.

# Manually testing the PR
`cargo nextest run -p jstz_proto runtime::v2 --features riscv`

### DIFF
--- a/crates/jstz_proto/src/event.rs
+++ b/crates/jstz_proto/src/event.rs
@@ -7,12 +7,9 @@ use tezos_smart_rollup::prelude::debug_msg;
 
 use crate::runtime::v2::oracle::OracleRequest;
 
-pub const JSTZ_PREFIX: &str = "[JSTZ]";
-
 /// Jstz Events
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
-pub enum Event {
-    OracleRequest(OracleRequest),
+pub trait Event: PartialEq + Serialize {
+    fn tag() -> &'static str;
 }
 
 /// Responsible for publishing events to the kernel debug log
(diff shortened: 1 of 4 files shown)


crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -83,3 +85,122 @@ impl TryFrom<&Url> for Address {
...
+            json!({
+                "method":[80,79,83,84],
+                "url":"http://example.com/foo",
+                "headers":[],
+                "body":{
+                    "Vector":[123,34,109,101,115,115,97,103,101,34,58,34,104,101,108,108,111,34,125]
+                }
+            }),
+            json
+        );
```
zcabter, line 204: Todo: Add decode  to complete roundtrip

crates/jstz_proto/src/event.rs
```diff
@@ -0,0 +1,173 @@
+#![allow(unused)]
+use bincode::{Decode, Encode};
+use jstz_core::host::HostRuntime;
+use nom::{bytes::complete::tag, InputTake};
+use serde::{Deserialize, Serialize};
+use tezos_smart_rollup::prelude::debug_msg;
+
+use crate::runtime::v2::oracle::OracleRequest;
+
+pub const JSTZ_PREFIX: &str = "[JSTZ]";
```
alanmarkoTrilitech, line 10: why [JSTZ], why not [ORACLE], I think it would fit better here

crates/jstz_proto/src/event.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// Jstz Events
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub enum Event {
```
alanmarkoTrilitech, line 14: would be good to have these definitions in a top level jstz_oracle crate, so that oracle node can use them

crates/jstz_proto/src/event.rs
```diff
@@ -0,0 +1,173 @@
...
+
+impl EventPublisher {
+    /// Jstz events are published as single line in the kernel debug log with the
+    /// schema "[JSTZ]<json payload>
+    pub(crate) fn publish_event<R>(&mut self, rt: &R, event: &Event) -> Result<()>
+    where
+        R: HostRuntime,
+    {
+        let json = serde_json::to_string(event).map_err(EncodeError::from)?;
+        debug_msg!(rt, "{JSTZ_PREFIX}{json}");
```
alanmarkoTrilitech, line 30: I thought that the request would be stored somewhere in rollup storage and only the request ID gets leaked in kernel debug log. no? at least that is how its written in the design doc

crates/jstz_proto/src/event.rs
```diff
@@ -0,0 +1,173 @@
...
+                body: Some(Body::Vector(
+                    serde_json::to_vec(&json!({ "message": "hello"})).unwrap(),
+                )),
+            },
+        });
+        let mut publisher = EventPublisher::default();
+        publisher.publish_event(&mut host, &event);
+        let head_line = sink.lines().first().unwrap().clone();
+        let decoded = decode_line(&head_line).unwrap();
+        assert_eq!(event, decoded)
```
alanmarkoTrilitech, line 171: testing decoding of something invalid would be useful too

crates/jstz_proto/src/event.rs
zcabter: Since these are protocol events, it makes sense for them to live in `jstz_proto`. We can make oracle node depend on `jstz_proto`

crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -121,11 +123,126 @@ impl TryFrom<&Url> for Address {
...
+    }
```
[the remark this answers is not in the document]
zcabter: Added

crates/jstz_proto/src/event.rs
alanmarkoTrilitech: proto seems like a bit heavy component, but ok let's not complicate it now

crates/jstz_proto/src/runtime/v2/fetch/http.rs
alanmarkoTrilitech: can you please either complete this or add a TODO comment?

zcabter: Done

PR #1046, opened 2025-06-04
feat(sequencer): get subkeys

huancheng-trili:
# Context

Part of JSTZ-601.
[JSTZ-601](https://linear.app/tezos/issue/JSTZ-601/expose-runtime-db-with-jstz-node-api)

Jstz node API needs to return data from runtime db when jstz node runs in sequencer mode.

# Description

Implemented one method `get_subkeys` to return subkeys from the runtime database and fixed the method `count_subkeys`.

I missed out some details in the previous implementation and realised that the query needs to be more complicated. For instance, for query key `/a`,

|key|should select|output|
|---|---|---|
|`/a`|Y|`""`|
|`/a/b`|Y|`"b"`|
|`/ab`|N|-
|`/b/a`|N|-
|`/a/c/d`|Y|`"c"`|

assuming that keys never have a trailing slash (not allowed in a host setup.) This means the query should select
* the target key itself
* keys with the target key as their prefix with one additional level
* keys with the target key as their prefix with two or more additional levels

and extract the immediate children. The previous `count_subkeys` query selects all keys prefixed with the target key, which essentially counts all children in the subtree and therefore is not correct.

Comments added to the method should explain how the query achieves this.

# Manually testing the PR

* Unit testing: added tests

### DIFF
--- a/crates/jstz_node/src/sequencer/db.rs
+++ b/crates/jstz_node/src/sequencer/db.rs
@@ -73,32 +73,76 @@ impl Db {
     }
 
     /// Counts subkeys given a prefix. The prefix itself is included. If the prefix does not exist,
+    /// i.e. it itself does not possess any value AND there is no other key with the prefix,
     /// `None` is returned.
     pub fn count_subkeys(&self, prefix: &str) -> Result<Option<u64>> {
-        // Using glob to find everything that matches `/{prefix}/*`.
-        // Since `jstz_key` is indexed and we only need the count,
-        // performance should be acceptable.
         let client = self.connection()?;
-        let mut glob_prefix = prefix.to_string();
-        if !glob_prefix.ends_with("/") {
-            glob_prefix += "/";
-        }
-        glob_prefix += "*";
 
+        // This is basically `get_subkeys` wrapped by `COUNT` in order to skip unnecessary
+        // return values.
         let mut stmt = client.prepare(
             r#"
-            SELECT 
-                CASE 
-                    WHEN a.val IS NULL THEN NULL 
-                    ELSE a.val + b.val 
-                END AS result
-            FROM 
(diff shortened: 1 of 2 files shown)


crates/jstz_node/src/sequencer/db.rs
```diff
@@ -73,32 +73,76 @@ impl Db {
...
+        // - INSTR(SUBSTR(jstz_key, LENGTH(?2)), '/'): returns the index where the first slash is in
+        //   the input string. This is essentially the length of the immediate subkey.
+        // The nested SUBSTR therefore means "returning a substring of the input (selected key with
+        // prefix removed) from the beginning to the first occurrence of the slash character".
+        let mut stmt = client.prepare(
+            r#"
+            SELECT SUBSTR(jstz_key, LENGTH(?2))
+            FROM jstz_kv
+            WHERE jstz_key = ?1
+                OR jstz_key GLOB ?2
```
ryutamago, line 129: nit:
even if it's functionally correct good to explicitly group it to avoid ambiguity (jstz_key = ?1 OR jstz_key GLOB ?2)
AND NOT...

crates/jstz_node/src/sequencer/db.rs
```diff
@@ -73,32 +73,76 @@ impl Db {
...
+                AND NOT jstz_key GLOB ?3
```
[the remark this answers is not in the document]
huancheng-trili: > GLOB is optimized via indexes iIUC where as LIKE is

I don't understand this sentence. Whatever that means, both use indices and I don't see any extra benefit in using LIKE while GLOB provides more flexibility when needed.

crates/jstz_node/src/sequencer/db.rs
huancheng-trili: Well there is no ambiguity as long as one follows the precedence of operators.

PR #1053, opened 2025-06-04
fix(example): update show-tez

huancheng-trili:
# Context

Closes JSTZ-629.
[JSTZ-629](https://linear.app/tezos/issue/JSTZ-629/update-show-tez)

# Description

Updated the example so that it works with the latest sandbox.

# Manually testing the PR

Ran the example locally.


timothymcmackin (approved):
Works for me locally with the current version of the sandbox docker container.

PR #1057, opened 2025-06-06
refactor(fetch): move fetch into dir mod

zcabter:
# Context
After working through the Oracle work, the fetch.rs file will grow and be refactored in rather complicated ways. Moving fetch into its own directory to allow better readability while logically relating their components. 

# Description

* Moves fetch into a directory mod and split http and errors

# Manually testing the PR
Only moving around of code, no code changes


alanmarkoTrilitech:
We need to coordinate a bit better cause I have also done this as part of working on the balance endpoint in v2.

alanmarkoTrilitech:
ok, well I assumed you wanted me to do it based on your comment in the balance endpoint PR. 

The change seems ok, but shouldn't it compile when uncommenting 
//#[cfg(feature = "riscv")]
pub mod v2;

in v2 mod.rs?

zcabter:
I don't understand, there isn't a "riscv" feature flag in `v2/mod.rs`

alanmarkoTrilitech:
yeah, sorry I meant runtime/mod.rs

zcabter:
I think it will compile. The flag is to prevent its inclusion in the wasm kernel.

zcabter:
Ok, then don't comment it out. I've never tried because it should be behind the feature flag anyway...

alanmarkoTrilitech:
so you are saying that if you comment it out, you don't get any rust analyzer errors and can build it with cargo build? It doesn't do so on my machine even after restarting nix develop and cargo clean. 

It should work though when the flag is removed, right? Otherwise its just a disconnected piece of code that we can't check.

alanmarkoTrilitech:
ah right, it works now. 

LGTM!

alanmarkoTrilitech (approved):

PR #1060, opened 2025-06-06
feat(proto): introduce host script, separate fetch handler, resources

alanmarkoTrilitech:
# Context
Preparation for implementation of the balance endpoint.

# Description
This PR introduces HostScript in v2 runtime and moves load_and_run function into it. It add additionally seaprates fetch_handler and resources so that dependencies are simpler.

# Manual testing
```make```

### CODE REVIEW

**Comment by alanmarkoTrilitech** [comment]
> I presume this PR is simply to prepare for the balance endpoint implementation? Could you add that within the context please?

added

**Comment by alanmarkoTrilitech** [comment]
> LGTM than the tests not being moved.

so they should be moved?

**Comment by zcabter** [comment]
> > LGTM than the tests not being moved.
> 
> so they should be moved?

Yes

File: crates/jstz_proto/src/runtime/v2/fetch/host_script.rs [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+use crate::context::account::Account;
+use crate::runtime::v2::fetch::fetch_handler::ProtoFetchHandler;
+
+pub struct HostScript;
+
+impl HostScript {
+    // - Loads the smart function script at `address`
+    // - Bootstraps a new runtime with new context and module loader
+    // - Runs the smart function
+    pub async fn load_and_run(
```
zcabter, line 25: `load_and_run` should not be part of the `HostScript`. `HostScript` is akin to pre compiles in ethereum, that is a set of pre-defined "smart functions" that users can call. I'd prefer that the public API exposed on HostScript matches, as much as possible, to what we present to users

alanmarkoTrilitech: can you please write down what do you imagine the function signatures to look like?
As I understand it, we have to pass it tx and host, and address,method,url,headers,body are basically args that would normally provided within fetch.

crates/jstz_proto/src/runtime/v2/fetch/sf_runner.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::rc::Rc;
+use url::Url;
+
+use crate::context::account::Account;
+use crate::runtime::v2::fetch::fetch_handler::ProtoFetchHandler;
+
+// - Loads the smart function script at `address`
+// - Bootstraps a new runtime with new context and module loader
+// - Runs the smart function
+pub async fn load_and_run(
```
zcabter, line 22: Hmm, my concern for isolating `load_and_run` into its own module and exposing it as a public fn is that callers may not know about the assumptions it makes ie. tx already began, url is valid sf url and headers pre/post processing of headers occur outside of the function. It also looks like an very important function is not tested but it actually is tested as part of `process_and_dispatch`.

I think its best to move this back to fetch_handler until we have a better way to decouple assumptions.

crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
```diff
@@ -0,0 +1,1195 @@
...
+use url::Url;
+
+use crate::context::account::{Account, Address, AddressKind, Addressable};
+use crate::runtime::v2::fetch::resources::{FetchRequestResource, FetchResponseResource};
+
+use super::http::{Body, Response, SupportedScheme};
+use std::num::NonZeroU64;
+use std::str::FromStr;
+pub struct ProtoFetchHandler;
+use super::sf_runner;
```
zcabter, line 19:
```suggestion
use super::sf_runner;

pub struct ProtoFetchHandler;

```

crates/jstz_proto/src/runtime/v2/fetch/sf_runner.rs
alanmarkoTrilitech: can't it be just made pub(super) and available only within /fetch?

zcabter (commented):
Could you squash the commits please?

PR #1063, opened 2025-06-06
feat(proto): balance endpoint in v2 runtime

alanmarkoTrilitech:
# Context
# Description
This PR adds balance endpoint in v2 runtime.
# Testing
```make```

### DIFF
--- a/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
+++ b/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
@@ -15,8 +15,10 @@ use crate::runtime::v2::fetch::resources::{FetchRequestResource, FetchResponseRe
 use super::host_script::HostScript;
 use super::http::{Body, Response, SupportedScheme};
 
+use crate::executor::smart_function::JSTZ_HOST;
 use std::num::NonZeroU64;
 use std::str::FromStr;
+
 pub struct ProtoFetchHandler;
 
 impl FetchHandler for ProtoFetchHandler {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs [hunk 1]
```diff
@@ -142,51 +144,62 @@ async fn dispatch_run(
...
+                    }
                 }
-            } else {
-                run_result
             }
         }
+        Err(_e) => match url.domain() {
+            Some(JSTZ_HOST) => {
+                HostScript::route(host, tx, from, method, url, headers, data).await
+            }
```
zcabter, line 200: Its more type safe to introduce a `HostName` type that is an enum between `Address` and `JstzHost`. You would then add the parsing logic in `TryFrom<&Url> for HostName`, replacing line 147

alanmarkoTrilitech: done

crates/jstz_proto/src/runtime/v2/fetch/host_script.rs
```diff
@@ -13,12 +13,30 @@ use jstz_runtime::{JstzRuntimeOptions, ProtocolContext};
...
+        headers: Vec<(ByteString, ByteString)>,
+        body: Option<Body>,
+    ) -> Result<Response> {
+        let path = url.path();
+        if path.starts_with("/balances") {
+            return Self::handle_balance(host, tx, from, method, url, headers, body)
+                .await;
+        }
+
+        todo!()
```
zcabter, line 37: I thought this was just here as a placeholder to get early feedback but I now realized you migh have missed cleaning up this line.

I think we can return a 404 not found response here

crates/jstz_proto/src/runtime/v2/fetch/host_script.rs
```diff
@@ -124,8 +142,59 @@ impl HostScript {
...
-    pub async fn handle_balance() {
-        todo!()
+    pub async fn handle_balance(
+        host: &mut impl HostRuntime,
+        tx: &mut Transaction,
+        self_address: Address,
+        method: ByteString,
+        url: Url,
+        _headers: Vec<(ByteString, ByteString)>,
+        _body: Option<Body>,
```
zcabter, line 152: These could be removed

crates/jstz_proto/src/runtime/v2/fetch/host_script.rs
```diff
@@ -124,8 +142,59 @@ impl HostScript {
...
+        let address_str = path
+            .strip_prefix("/balances/")
+            .ok_or_else(|| FetchError::JstzError("Invalid path format".to_string()))?;
+
+        match Self::get_balance(host, tx, address_str, &self_address) {
+            Ok(balance) => Ok(Response {
+                status: 200,
+                status_text: "OK".to_string(),
+                headers: vec![],
+                body: Body::Vector(balance.to_string().as_bytes().to_vec()),
```
zcabter, line 173: Could you add a `From<String> for Body` and `From<&[u8]> for Body` conversion functions please?

crates/jstz_proto/src/runtime/v2/fetch/host_script.rs
```diff
@@ -124,8 +142,59 @@ impl HostScript {
...
+    fn get_balance(
+        host: &mut impl HostRuntime,
+        tx: &mut Transaction,
+        address_str: &str,
+        self_address: &Address,
+    ) -> Result<u64> {
+        let target_address = if address_str == "self" {
+            self_address.clone().into()
+        } else {
+            crate::context::account::Address::from_base58(address_str)
```
zcabter, line 193: If you're using a Jstz type, most of the time there is no need to fully qualify it.

zcabter: Actually, it seem its already imported so the fully qualified part can simply be removed.

crates/jstz_proto/src/runtime/v2/fetch/host_script.rs
```diff
@@ -13,12 +13,33 @@ use jstz_runtime::{JstzRuntimeOptions, ProtocolContext};
...
+        if path.starts_with("/balances") {
+            return Self::handle_balance(host, tx, from, method, url).await;
+        }
+
+        // Return 404 for all other paths
+        Ok(Response {
+            status: 404,
+            status_text: "Not Found".to_string(),
+            headers: vec![],
+            body: Body::Vector("Not Found".as_bytes().to_vec()),
```
zcabter, line 39: This would usually be an empty body

zcabter:
Thanks for the changes. Waiting for the rebase and build to pass

alanmarkoTrilitech:
I've rebased it, can you check again please?

crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
```diff
@@ -203,65 +185,76 @@ pub async fn process_and_dispatch_request(
...
-                } else {
-                    let to: Address = (&url).try_into()?;
-                    let headers = process_headers_and_transfer(
-                        tx,
+    let to = (&url).try_into();
+    match to {
+        Ok(HostName::Address(to)) => {
+            log_request(host, &to, operation_hash.as_ref(), LogEvent::RequestStart);
+            let mut headers =
+                process_headers_and_transfer(tx, host, headers, &from, &to)?;
```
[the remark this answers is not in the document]
alanmarkoTrilitech: the issue with wrapping log request is that it needs host which is borrow inbetween calls to request start and end. A simple solution is to however just separate the code inbetween logging into its own function

PR #1069, opened 2025-06-09
fix(docs): clarify bootstrap accounts and Jstz accounts

timothymcmackin:
I just thought this could use some clarification so people didn't wonder why their accounts didn't have any tez at the start.


huancheng-trili (approved):
I think we can also update the sandbox to make it clear that these are L1 addresses.

PR #1071, opened 2025-06-09
Make sequencer worker async

ryutamago:
# Context

[task link](https://linear.app/tezos/issue/JSTZ-636/make-worker-spawn-async-aware)

In preparation for [supporting deposit](https://linear.app/tezos/issue/JSTZ-637/handle-deposit-operation). The FA deposit is async.

# Description

This PR makes the worker async aware and replace the blocking lock with non blocking lock from tokio. By switching to a non blocking lock, we can now avoid blocking the entire Tokio runtime thread from the worker.

* Make the worker async aware. initialized a single threaded tokio runtime within the new thread
* Replace the std::RwLock with tokio::RwLock

NOTE: In the worker, we can use the `tokio::spawn_blocking` which spawns a standalone thread that tokio is aware of instead of initializing a new tokio runtime but for simplicity we can stick to a new runtime for now

# Manually testing the PR

existing test works

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -14,12 +14,10 @@ use services::{
     operations::OperationsService,
     utils,
 };
-use std::{
-    path::PathBuf,
-    sync::{Arc, RwLock},
-};
+use std::{path::PathBuf, sync::Arc};
 use tempfile::NamedTempFile;
 use tokio::net::TcpListener;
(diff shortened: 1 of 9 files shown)


ryutamago:
@huancheng-trili 
Will work on switching to spwan_blocking after the deposit is done.
[link to task](https://linear.app/tezos/issue/JSTZ-638/use-spawn-blocking).

PR #1075, opened 2025-06-10
feat(ci): publish to AWS for jstz.tezos.com

timothymcmackin:
We have jstz.tezos.com and https://next-jstz-tezos-com.tzstaging.com/ set up for Jstz docs. To make this work we need the secrets from the workflow files to be added to this repo. I can get those secrets in the hands of someone with rights to add secrets to the repo.

For reference, this code is how we publish docs.tezos.com: https://gitlab.com/tezos/docs/-/blob/main/.github/workflows/deploy-production.yml


huancheng-trili (commented):
Um why does the dev build need to be hosted on AWS as well? And do we really need AWS when custom domains can be configured for github pages?

huancheng-trili (approved):

PR #1078, opened 2025-06-11
feat(proto): add v2 run_toplevel_fetch

zcabter:
# Context
Part of https://linear.app/tezos/issue/JSTZ-599/integrate-v2-fetch 
Closes https://linear.app/tezos/issue/JSTZ-650/add-run-toplevel-fetch 

Add  `run_toplevel_fetch` for v2 runtime

# Description
* Removed execute_v2 and related code
* Wraps `process_and_dispatch_request` with `run_toplevel_fetch`
* Extends proto error with `V2Error`
* Add`make test-v2` to run with "v2_runtime" enabled
* V1 - Changed `Response.error` to have status code 400 instead of 500 
* V1 - Introduced `internal_server_error` for internal errors in V1
* Fixed up tests
* Ignore withdrawal tests for v2 as HostScript not yet implemented
* Ignore test that tests strictness of header hygiene for v2 as v2 fetch cleans bad headers instead of failing
* Ignore noop fetch tests for v2 as it is not yet implemented
* Adds tokio runtime to sequencer worker thread

# Manually testing the PR
`make test && make test-v2`

### DIFF
--- a/crates/jstz_api/src/http/response.rs
+++ b/crates/jstz_api/src/http/response.rs
@@ -268,6 +268,21 @@ impl ResponseBuilder {
         //
         // FIXME: jstz only supports status, body and headers here.
         // FIXME: http::Response won't support 0 as a status code
+        let status = 400;
+        let body = Body::null();
+        let headers = Headers::new();
+
+        Ok(Response {
+            response: InnerResponse::builder()
+                .status(status)
+                .body(body)
+                .expect("fixed inputs should never fail"),
+            headers: JsNativeObject::new::<HeadersClass>(headers, context)?,
+            url: None,
+        })
+    }
+
+    pub fn internal_server_error(context: &mut Context) -> JsResult<Response> {
         let status = 500;
         let body = Body::null();
         let headers = Headers::new();
--- a/crates/jstz_node/src/sequencer/runtime.rs
+++ b/crates/jstz_node/src/sequencer/runtime.rs
@@ -61,11 +61,15 @@ fn read_injector(rt: &impl Runtime) -> Option<PublicKey> {
 pub fn process_message(rt: &mut impl Runtime, op: Message) -> anyhow::Result<()> {
     let ticketer = read_ticketer(rt).ok_or(anyhow!("Ticketer not found"))?;
     let injector = read_injector(rt).ok_or(anyhow!("Revealer not found"))?;
+    let tokio = tokio::runtime::Builder::new_current_thread()
+        .build()
+        .unwrap();
     let mut tx = Transaction::default();
     tx.begin();
     let receipt = match op {
-        Message::External(op) => jstz_utils::TOKIO
-            .block_on(execute_operation(rt, &mut tx, op, &ticketer, &injector)),
+        Message::External(op) => {
+            tokio.block_on(execute_operation(rt, &mut tx, op, &ticketer, &injector))
+        }
         Message::Internal(op) => match op {
             InternalOperation::Deposit(op) => deposit::execute(rt, &mut tx, op),
             _ => {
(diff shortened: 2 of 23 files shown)


crates/jstz_node/src/sequencer/runtime.rs
```diff
@@ -61,11 +61,15 @@ fn read_injector(rt: &impl Runtime) -> Option<PublicKey> {
 pub fn process_message(rt: &mut impl Runtime, op: Message) -> anyhow::Result<()> {
     let ticketer = read_ticketer(rt).ok_or(anyhow!("Ticketer not found"))?;
     let injector = read_injector(rt).ok_or(anyhow!("Revealer not found"))?;
+    let tokio = tokio::runtime::Builder::new_current_thread()
+        .build()
+        .unwrap();
```
huancheng-trili, line 66: Is it safe to unwrap? And is there any downside of building a new tokio runtime each time?

huancheng-trili: Then what's the downside of using that global runtime in test utils?

ryutamago (approved):

crates/jstz_api/src/http/response.rs
```diff
@@ -268,6 +268,21 @@ impl ResponseBuilder {
         //
         // FIXME: jstz only supports status, body and headers here.
         // FIXME: http::Response won't support 0 as a status code
+        let status = 400;
```
johnyob, line 271: This technically isn't spec compliant, `Response.error()` should return a "network error" (status 0, empty body, empty headers). Perhaps this is what we want, since the uses of `Response::internal_server_error` are actually for network related issues (i.e. transfer failed)

crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -22,7 +22,12 @@ pub struct Response {
 
 impl Into<http::Response<Option<Vec<u8>>>> for Response {
     fn into(self) -> http::Response<Option<Vec<u8>>> {
-        let mut builder = http::Response::builder().status(self.status);
+        // According to JavaScript documentation, `Response.error()` returns a response with status code 0
+        // and is mainly used for client side network errors. In regular JS, the fetch promise would be rejected.
+        // Within Jstz, the smart function can only return this if it explicitly called `Response.error()`, which
+        // means the intent is closer to a 400 Bad Request.
+        let status = if self.status == 0 { 400 } else { self.status };
```
johnyob, line 29: Similarly, I don't think we should implicitly change the status code here. I agree that the user probably intended something different, but we should be compliant with spec behaviour of `Response.error`

zcabter: `http` won't allow a 0 status code

crates/jstz_node/src/sequencer/runtime.rs
zcabter: @huancheng-trili I remove building the tokio instance each time to building it just once when the worker thread starts. See `crates/jstz_node/src/sequencer/worker.rs` 

When I made the refactor, the large payload test started to fail which was the only failing test. It had something to do with the DAC page encoding but I'm not sure why it didn't fail in the case where tokio was rebuilt each time...very  odd. I re-wrote that test.

crates/jstz_proto/src/runtime/v2/fetch/http.rs
johnyob: Ah, because the spec doesn't consider 0 as a status code. Hmm, I think implicitly rewriting the status code instead of failing here a bit odd -- but I think as long as we make this explicit in docs, then we should be fine. Ironically Deno just fails silently relying on https://github.com/denoland/deno/blob/f54afe03bd5902576a1e8e71d0473a63279a663a/ext/fetch/23_response.js#L180, but this doesn't handle `Response.error`

PR #1079, opened 2025-06-11
feat(oracle): filtered event stream

alanmarkoTrilitech:
# Context
[JSTZ-535](https://linear.app/tezos/issue/JSTZ-535/implement-filteredeventlogstreamt-event)
# Description
This PR adds Filtered even log stream, that for a given regex and a log file, filters out and streams the lines that match a given pattern. In the future we may modify it to emit some Event enums. 

# Manual testing
```make```

### CODE REVIEW

File: crates/jstz_oracle_node/src/filtered_log_stream.rs [hunk 1]
```diff
@@ -0,0 +1,179 @@
...
+                let _ = tx.send(Err(e.into())).await;
+                return;
+            }
+
+            let mut buf = String::new();
+
+            loop {
+                buf.clear();
+                match reader.read_line(&mut buf).await {
+                    Ok(0) => {
```
zcabter, line 42: Could you add a short comment on what `Ok(0)` means?

crates/jstz_oracle_node/src/filtered_log_stream.rs
```diff
@@ -0,0 +1,179 @@
...
+                        while buf.ends_with(['\n', '\r']) {
+                            buf.pop();
+                        }
+                        if pattern.is_match(&buf) {
+                            let _ = tx.send(Ok(buf.clone())).await;
+                        }
+                    }
+                    Err(e) => {
+                        let _ = tx.send(Err(e.into())).await;
+                        return;
```
zcabter, line 55: Why not just use tail file to get the lines then regex match on it? It seems like you're doing the same thing.

crates/jstz_utils/src/tailed_file.rs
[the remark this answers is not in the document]
alanmarkoTrilitech: right, sorry seems like I didn't push some changes properly

zcabter:
The logic looks fine overall but I'm not sure why we're not using `TailFile` for the line parsing logic

crates/jstz_oracle_node/src/filtered_log_stream.rs
```diff
@@ -0,0 +1,167 @@
...
+};
+
+use anyhow::Result;
+use regex::Regex;
+use tokio::sync::mpsc;
+
+use jstz_utils::tailed_file::TailedFile;
+use std::time::Duration;
+
+pub struct FilteredLogStream {
```
zcabter, line 14: Could you add a doc comments on the behaviour of `FilteredLogStream` please? In particular, the main behaviour and what happens when EOF, if opened against an existing file and what it means if the stream returns `None` vs `Some(Ok)` vs `Some(Error)`

crates/jstz_oracle_node/src/filtered_log_stream.rs
```diff
@@ -0,0 +1,167 @@
...
+impl FilteredLogStream {
+    pub async fn new(pattern: Regex, path: PathBuf) -> Result<Self> {
+        let file = TailedFile::init(&path).await?;
+
+        let (tx, rx) = mpsc::channel(1024);
+
+        tokio::spawn(async move {
+            let mut lines = file.lines();
+            loop {
+                match lines.next_line().await {
```
zcabter, line 27: its worth using using a `CancellationToken` where `FilteredLogStream` will cancel it on drop while we use `select!` to await on either a cancellation event on the next line. This would fix the case where the stream is dropped but the task does not drop because no new lines are being sent, causing a file descriptor leak.

crates/jstz_oracle_node/src/filtered_log_stream.rs
alanmarkoTrilitech: added

zcabter (approved):
LGTM, thanks

PR #1080, opened 2025-06-11
feat(proto): parsed code for v2 runtime

zcabter:
# Context
Add compatible `ParsedCode` struct
 
Closes https://linear.app/tezos/issue/JSTZ-651/add-parsedcode

# Description
* Added `ParsedCode` 
* Checks in `ParsedCode` should contain all the checks that could reject a code before it is deployed as a smart function.
* Note that even if code is parsable, it might not necesarily mean that it is a valid smart function. It is also possible for malicious to be run. Specifically
  1. The user could try to modify state through KV or fetch. For this reason, the JstzRuntime instantiated is built w no access to the protocol or related ops - fetch, kv, console
  2. The code could run an infinite loop. We will need gas to solve this.
* Parsing involves compiling the source, instantiating module imports, evaluating the module code and checking that the default handler exists. Note that we do no not allow module imports so existence of imports will return error.
* A bug was found when explicitly entering/exiting sync executions because exit will not be called due to eager error return of `?`. I've introduced an `Entered` struct that ensures isolate is existed upon being dropped. In addition, I renamed `IsolatedFuture` to `AsyncEntered` to follow the same naming convention and refactored it to operate over `&mut JstzRuntime` instead of `&mut OwnedIsolate` to have a similar API with `Entered`

# Manually testing the PR
`cargo nextest run -p jstz_proto parsed_code --features riscv`


alanmarkoTrilitech:
how is "The user could try to modify state through KV or fetch" wrong since we expect that sfs would use these functions?

crates/jstz_proto/src/runtime/v2/parsed_code.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        // 3. Instantiate Module
+        let _ = module.instantiate_module(scope, |_, _, _, resolved_module| {
+            // We do not support dynamic imports so just return the resolved
+            // module as is
+            Some(resolved_module)
+        });
+        check_compile_exception(scope, &module)?;
+
+        // 4. Evaluate Module
```
alanmarkoTrilitech, line 77: the comments could be more descriptive as they are basically repetition of the function name, maybe some quick mention of what it checks/what can fail or something similar would be useful

zcabter:
@alanmarkoTrilitech I've made quite a bit of change after finding out that calling protocol ops without protocol present would panic instead of throwing an error

crates/jstz_proto/src/runtime/v2/parsed_code.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    // Use scope slot (aka annexed memory) to pass import buffer to callback
+    let import_detected = Rc::new(RefCell::new(ImportDetected(false)));
+    scope.set_slot(import_detected.clone());
+
+    let result = module.instantiate_module(&mut scope, resolve_module_callback);
+    if result.is_none() {
+        let has_imports = import_detected.borrow();
+        if has_imports.0 {
+            return Err(ParseError::ImportsNotSupport);
```
alanmarkoTrilitech, line 140: nit: ParseError::ImportsNotSupported seems better name

crates/jstz_runtime/src/runtime.rs
```diff
@@ -161,9 +162,8 @@ impl JstzRuntime {
     /// Executes traditional, non-ECMAScript-module JavaScript code, ignoring
     /// its result
     pub fn execute(&mut self, code: &str) -> Result<()> {
-        unsafe { self.v8_isolate().enter() }
-        self.execute_script("jstz://run", code.to_string())?;
-        unsafe { self.v8_isolate().exit() }
+        let mut this = Entered::new(self);
```
alanmarkoTrilitech, line 165: can you please add to the PR description what was the bug and how the changes here fix it?

crates/jstz_proto/src/runtime/v2/parsed_code.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    #[test]
+    fn parse_with_attempted_protocol_access_fails() {
+        let code = r#"
+            Kv.set("test", "test")
+            let t = Kv.get("test")
+            export default () => 42
+        "#;
+        let error = ParsedCode::parse(code.to_string()).unwrap_err();
```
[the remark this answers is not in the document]
zcabter: This test rejects top level protocol usage

crates/jstz_runtime/src/runtime.rs
zcabter: Done

alanmarkoTrilitech (approved):
LGTM, but please respond to the comment first

PR #1083, opened 2025-06-12
feat(oracle): relay

alanmarkoTrilitech:
# Context
[JSTZ-536](https://linear.app/tezos/issue/JSTZ-536/implement-relay)
# Description
This PR implements relay for the oracle node. Relay is the publisher in the pub-sub mechanism that publishes requests from a stream of OracleRequestEvents

# Testing
```make```

### DIFF
--- a/crates/jstz_oracle_node/src/lib.rs
+++ b/crates/jstz_oracle_node/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg(feature = "v2_runtime")]
 mod filtered_log_stream;
 pub mod relay;
 pub mod request;

### CODE REVIEW

File: crates/jstz_oracle_node/src/relay.rs [hunk 1]
```diff
@@ -0,0 +1,167 @@
...
+                    }
+                }
+            }
+        });
+
+        Ok(Self { tx })
+    }
+
+    pub fn subscribe(&self) -> broadcast::Receiver<OracleRequest> {
+        self.tx.subscribe()
```
zcabter, line 55: Perhaps check receiver count == 1?

crates/jstz_oracle_node/src/relay.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use futures_util::StreamExt;
+use tokio::sync::broadcast;
+
+use crate::filtered_log_stream::FilteredLogStream;
+use crate::request::{request_event_from_log_line, OracleRequest, ORACLE_LINE_REGEX};
+
+use anyhow::Result;
+
+pub struct Relay {
```
zcabter, line 11: Add doc string comment

crates/jstz_oracle_node/src/relay.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        });
+
+        Ok(Self {
+            tx,
+            subscriber_count: std::sync::atomic::AtomicUsize::new(0),
+        })
+    }
+
+    pub fn subscribe(&self) -> Result<broadcast::Receiver<OracleRequest>> {
```
zcabter, line 58: Add doc string comment

crates/jstz_oracle_node/Cargo.toml
```diff
@@ -22,4 +22,13 @@ tempfile.workspace = true
 rand.workspace = true
 futures-util.workspace = true
 tokio-util.workspace = true
-jstz_utils = { path = "../jstz_utils" }
\ No newline at end of file
+jstz_utils = { path = "../jstz_utils" }
+once_cell.workspace = true
+jstz_proto = { path = "../jstz_proto", optional = true}
```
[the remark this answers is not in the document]
alanmarkoTrilitech: there seem to be some ```cargo check``` errors in the code behind v2_runtime that were merged before. Those don't seem to be caused by this change, but if Cargo.toml is changed as you suggest, they get enabled and cause issues due to which I then can't push the code.  

addtionally, there seems to be a failing test - runtime::v2::ledger::test::transfer even on main - when testing with ```make test-v2```.

PR #1085, opened 2025-06-12
fix(sequencer): use tokio RwLock

ryutamago:
# Context

Closes [task link](https://linear.app/tezos/issue/JSTZ-636/make-worker-spawn-async-aware)

The sequencer currently uses a blocking lock inside tokio runtime. Blocking inside a tokio task can block the entire Tokio core thread which could lead to a performance issue when the number of access to the queue increases. This PR fixes it.
# Description

* Replaced the `std::sync::RwLock` with a non blocking lock `tokio::sync::RwLock`

# Manually testing the PR

existing test works

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -14,12 +14,9 @@ use services::{
     operations::OperationsService,
     utils,
 };
-use std::{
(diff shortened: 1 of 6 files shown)


huancheng-trili:
Please rebase.

PR #1088, opened 2025-06-13
feat(runtime): write console messages in logging format

huancheng-trili:
# Context

Part of JSTZ-647.
[JSTZ-647](https://linear.app/tezos/issue/JSTZ-647/log-operation-details-for-v2-runtime)

V2 runtime has no logging basically.

# Description

Updated the console extension for v2 runtime such that log lines are printed in a specific format recognised by the log collector in jstz node. A sample line looks as follows:
```
[JSTZ:SMART_FUNCTION:LOG] {"address":"KT1RJ6PbjHpwc3M5rw5s2Nbmefwbuwbdxton","requestId":"js_types","level":"INFO","text":"false\n"}
```

This is the same as what is implemented for v1 runtime.

Note that since this format is only required when the runtime is built into the kernel, a new feature flag `kernel` is added such that `console` still writes messages in the old format in normal cases, e.g. with repl. That is,
```
[INFO] something
```

# Manually testing the PR

* Unit testing: updated tests

I tried to create new tests for the changes and make the test runner ignore them when the feature flag is not set, but then codecov complains that those new tests "are not covered" because they are ignored. For now I'm "merging" these tests in an awkward way like this. Ideally we need to collect test coverage for different features, but for now I'll keep it like this.

### DIFF
--- a/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
+++ b/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
@@ -280,7 +280,7 @@ async fn load_and_run(
     let mut body = body;
 
     // 0. Prepare Protocol
-    let mut proto = ProtocolContext::new(host, tx, address.clone());
+    let mut proto = ProtocolContext::new(host, tx, address.clone(), String::new());
 
     // 1. Load script
     let script = { load_script(tx, &mut proto.host, &proto.address)? };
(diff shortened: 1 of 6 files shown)


ryutamago (approved):

ryutamago:
lgtm, since it invovles v2 runtime i will let @zcabter do the final review!

PR #1091, opened 2025-06-13
feat(proto): integrate v2 kv

zcabter:
# Context

Closes https://linear.app/tezos/issue/JSTZ-654/add-kv-and-kvvalue

Kv has been moved from `proto` to `runtime` and perhaps should ultimately live in core.. but we can refactor that at a later time.

# Description
*  Exposes KV from runtime 
* Updates KV apis to use core error

# Manually testing the PR
`make build && make test && make build-v2 && make test-v2`


alanmarkoTrilitech (approved):
the PR lgtm. However might be worth it at some point to clearly mark what is part of v2 and v1 to avoid cross dependencies, cause besides this PR it isn't obvious that crates/jstz_runtime/src/ext/jstz_kv/kv.rs is already made for v2 and what other parts were already changed in a similar way.

PR #1095, opened 2025-06-16
fix(sequencer,jstzd): wait for rollup

ryutamago:
# Context

Sequencer assumes that the rollup node is healthy by the time it starts. However jstzd starts the sequencer before the rollup is healthy. This PR fixes that.

# Description

* make jstzd::spawn wait for the rollup to spin up before spinning up the jstz node
* removed the sleep used for temp fix inside inbox::monitor 

# Manually testing the PR

* manually tested by running jstzd 
* existing test works

### DIFF
--- a/crates/jstz_node/src/sequencer/inbox/mod.rs
+++ b/crates/jstz_node/src/sequencer/inbox/mod.rs
@@ -53,6 +53,7 @@ impl WriteDebug for Logger {
 }
 
 /// Spawn a future that monitors the L1 blocks, parses inbox messages and pushes them into the queue.
+/// precondition: the rollup node is healthy.
 pub async fn spawn_monitor<
     #[cfg(test)] Fut: Future<Output = ()> + 'static + Send,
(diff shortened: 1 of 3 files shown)


ryutamago:
@huancheng-trili 
yes the test works locally.

ryutamago:
> It's not that stable, but it's not a new problem. This is definitely the first time with that 500 error though.

ok, i will run the test a few more times just to make sure.

ryutamago:
@huancheng-trili 
> inject handler tries to insert the operation into the queue, the inbox message reader is in the middle of doing so and has the lock, so the handler fails 

That makes sense, I guess it is failing only in the CI as it takes more time to process the external messages and ends up overlapping with the inbox process. 
 
The CI is saying {"error":"database is locked"} though, is this related to the queue being locked? 

Either way i think it's out of scope of this PR - should we merge this first with the delay as a temp fix and work on fixing this bug?

huancheng-trili:
We can keep the delay for now and get this merged. Please clean up the commented lines in the test.

PR #1102, opened 2025-06-16
feat(docs): add biel chatbot and search

timothymcmackin:
We have a chatbot that's been trained on Tezos docs. I added Jstz to its sources and it can now answer questions about Jstz.

Preview: https://timothymcmackin.github.io/previews/jstz/biel-chatbot

As described in [this page](https://docs.tezos.com/overview/chatbot), the chatbot has a button at the right side of the page and also its own search box at the top. Now one search is AI-powered for all TEzos info and the other is ordinary keyword search of only Jstz docs. 

Screenshot 2025-06-16 at 4 29 51 PM

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -3,6 +3,14 @@
 
 import { themes } from "prism-react-renderer";
 
+const BIEL_PROJECT = "w30c27l8fg";
+const BIEL_WARNING =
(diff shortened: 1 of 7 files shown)


timothymcmackin:
Changed it to "Can I call third-party JavaScript libraries in smart functions?" which returns good information.

NicNomadic (approved):

PR #1103, opened 2025-06-18
fix(docs): restore favicon

timothymcmackin:
The docs are missing the favicon. This restores it. I took the icon from jstz.dev.

Preview: https://timothymcmackin.github.io/previews/jstz/favicon/

Screenshot 2025-06-18 at 9 06 08 AM


timothymcmackin:
> I think it's a good opportunity to choose a better favicon, e.g. the jstz-dev github organisation avatar. What do you think?

Makes sense. I changed it to that icon. @josh-reid , what do you think?

Should the icon at the top left, by the Jstz title text, be the same icon?

josh-reid:
> > I think it's a good opportunity to choose a better favicon, e.g. the jstz-dev github organisation avatar. What do you think?
> 
> Makes sense. I changed it to that icon. @josh-reid , what do you think?
> 
> Should the icon at the top left, by the Jstz title text, be the same icon?

Yeah makes sense to me @timothymcmackin I think it's better than the emoji we were using, in the near future we'll probably need to do some light updating to the branding regardless to be in line with the landing page and overall Tezos brand system that we'll be inheriting but this is better for now!

timothymcmackin:
ok, I removed the emoji from the site title as well; there's no easy way to put an svg in there, I don't think. Site is pretty plain, needs some spice from the design team.

Screenshot 2025-06-18 at 12 00 06 PM

PR #1104, opened 2025-06-18
feat(runtime): terminate slow executions in native environment

huancheng-trili:
# Context

Completes JSTZ-664.
[JSTZ-664](https://linear.app/tezos/issue/JSTZ-664/implement-execution-timeout)

The V2 runtime needs to terminate function calls involving infinite loops.

# Description

Updated the v2 runtime such that executions that take more than 500ms are terminated. This solution involves spinning up a timer in a separate thread, which is not going to work in the rollup, so a feature `native` is introduced to hide the relevant code. This is also why the timeout is set to 500ms -- keeping this too long means that the thread pool needs to be larger as the timer thread does not get cancelled when the corresponding execution finishes early. 500ms should be reasonable, given that there is only one processing unit and that we don't want any execution to block the runtime for too long. Terminating executions in this way is possibly the simplest since there is no need to handle transactions or craft receipt with custom errors outside the runtime. One downside is that the error code is 500 and the error message isn't very clear, but I think this is acceptable for now as long as the main issue is addressed.

# Manually testing the PR

* Unit testing: updated the relevant test in jstz_runtime and added one test in jstz_node to demonstrate that it works.

### DIFF
--- a/crates/jstz_runtime/src/runtime.rs
+++ b/crates/jstz_runtime/src/runtime.rs
@@ -1,4 +1,5 @@
 use crate::error::Result;
+use crate::error::RuntimeError;
 use crate::ext::jstz_fetch;
 use crate::ext::jstz_fetch::NotSupportedFetch;
 use deno_core::v8::new_single_threaded_default_platform;
@@ -12,6 +13,14 @@ use jstz_crypto::smart_function_hash::SmartFunctionHash;
 use pin_project::pin_project;
 use std::marker::PhantomData;
 use std::mem::ManuallyDrop;
+#[cfg(feature = "native")]
+use tokio::sync::{
+    mpsc::{channel, Receiver, Sender},
+    oneshot,
+};
+#[cfg(feature = "native")]
+use tokio_util::sync::CancellationToken;
+
 use std::{
     future::Future,
     ops::{Deref, DerefMut},
@@ -28,6 +37,11 @@ use deno_console;
 use deno_url;
 use deno_web::TimersPermission;
 use deno_webidl;
+
+#[cfg(not(test))]
+#[cfg(feature = "native")]
+static RUNTIME_THREADPOOL: std::sync::LazyLock<rayon::ThreadPool> =
+    std::sync::LazyLock::new(|| rayon::ThreadPoolBuilder::new().build().unwrap());
 
 /// Returns the default object of the specified JavaScript namespace (Object).
 ///
@@ -90,6 +104,8 @@ pub struct JstzRuntimeOptions {
     pub module_loader: Rc<dyn ModuleLoader>,
     /// Fetch extension
     pub fetch: Extension,
+    /// Shared execution timeout set at the top level.
+    pub timeout: Option<ExecutionTimeout>,
 }
 
 impl Default for JstzRuntimeOptions {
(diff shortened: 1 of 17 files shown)


crates/jstz_runtime/src/runtime.rs
```diff
@@ -29,6 +37,10 @@ use deno_url;
 use deno_web::TimersPermission;
 use deno_webidl;
 
+#[cfg(feature = "native")]
+static RUNTIME_THREADPOOL: std::sync::LazyLock<rayon::ThreadPool> =
+    std::sync::LazyLock::new(|| rayon::ThreadPoolBuilder::new().build().unwrap());
```
zcabter, line 42: The thread pool limits the depth of fetch calls to the number of threads, so I think we should explicitly set this number instead of relying on number of cores.

crates/jstz_runtime/src/runtime.rs
```diff
@@ -247,16 +259,42 @@ impl JstzRuntime {
             let default_fn = v8::Local::<v8::Function>::try_from(default_value)?;
             v8::Global::new(scope, default_fn)
         };
+
+        // Set up a timer in a different thread to kill executions that take too long.
+        // This is only for the native execution environment.
+        #[cfg(feature = "native")]
+        let flag = {
+            let isolate = self.v8_isolate().thread_safe_handle();
+            set_execution_timeout(isolate)
```
zcabter, line 268: This will spawn a new thread for each `call_default_handler` which means a thread for each `fetch` call, opening up potential attack of starving threads, although somewhat alleviated by the 500ms time slice. 

Instead, we should time execution at the top level and all child fetch calls should share time slices of that timer.  The timer maintains a set of IsolateHandles that get terminated on timeout. Since we only execute one top level operation at a time, for each fetch call, we only need to check if the timer is currently active, and if so, add it to the set of isolate handles

crates/jstz_runtime/src/runtime.rs
```diff
@@ -247,16 +259,42 @@ impl JstzRuntime {
...
+        };
```
zcabter, line 269: Additionally, its worth moving the logic to `new()` to catch infinite loops during module evaluation when code is deployed. I'd add an optional flag in `JstzRuntimeOptions` indicating if a timer should be set or not and for how long

huancheng-trili:
I implemented a different version that initialises the timer at the top level as you suggested. The PR becomes too big right now. I'll clean it up later if this version is okay.

The following is what happens:
* The timer is instantiated at the top level and the clock starts immediately.
* For each fetch call (including the top one,) the instantiated runtime registers its isolate to the timer. The timer has a capacity, and if it's full, the call is rejected immediately.
* When the time is up, the timer closes registration and starts to terminate all isolates registered. New attempts for registration will fail and the corresponding calls will be rejected immediately.
* If the top level fetch call completes before the deadline, the timer will be cancelled.

This implementation suffers from race conditions. It cannot deal with the following situation:
* New fetch call C registers itself to the timer.
* The time is up and the timer terminates executions in C's isolate.
* C actually starts to run the default module.

In step 2, since there is no execution in C's isolate yet, that termination becomes a no-op, and the actual execution later on becomes untracked. I'm not sure what will happen to a child isolate when the execution in the parent isolate is cancelled; I guess it will keep running, which is not good, but then it won't block anything because the parent is cancelled...?

In any case, even the initial naive implementation can have this problem: the timer starts in a new thread, but somehow the main thread is suspended and starts the actual module execution after the timer completes, though the chance is rather rare given that the deadline is long enough. I guess this problem always exists if this termination is not handled at a lower level.

PR #1106, opened 2025-06-18
fix(docs): some fixes for the quickstart

timothymcmackin:
- Simplify prereqs
- Update info about wallet interface
- Correct port that the wallet sample application runs on


docs/quick_start.md
```diff
@@ -277,7 +273,7 @@ Follow these steps to run a sample web application that uses your smart function
    pnpm dev
    ```
 
-1. Open the application in Chrome at `http://localhost:3000`.
+1. Open the application in Chrome at `http://localhost:3201`.
```
NicNomadic, line 276: Wouldn't it be more robust to tell to look at the output of the previous command, and use a placeholder here?

NicNomadic (approved):
LGTM beyond the optional suggestion above

docs/quick_start.md
timothymcmackin: Updated.

PR #931, opened 2025-04-14
Remove emoji

ryutamago:
quick fix on removing the emoji in the logs .

PR #1107, opened 2025-06-18
fix(docs): logs no longer show emojis

timothymcmackin:
Per https://github.com/jstz-dev/jstz/pull/931, logs no longer show emojis. This PR removes them from the docs.


docs/cli.mdx
```diff
@@ -175,10 +175,10 @@ $ jstz run --trace "jstz://${counter}/"
 You should be able to see an output of the counter smart function looking like this:
 
 ```
-[🪵] Counter: null
-[🪵] Counter: 0
-[🪵] Counter: 1
-[🪵] Counter: 2
+[LOG] Counter: null
```
huancheng-trili, line 178: In `main` and the next release, this is `DEBUG`.

huancheng-trili: How did you try main? I need to correct myself: this is going to be `INFO`.

PR #1114, opened 2025-06-20
fix(sequencer): change runtime db journal config

huancheng-trili:
# Context

Completes JSTZ-671. (Maybe)
[JSTZ-671](https://linear.app/tezos/issue/JSTZ-671/fix-runtime-db-connections)

There were some random errors related to the runtime database in #1095.

# Description

Updated the runtime database's config, hoping to make accesses less problematic. WAL journal mode should permit reads during writes instead of terminating all ongoing reads when a write takes place.

# Manually testing the PR

Ran tests with the CI runner multiple times and it seemed fine. Not completely sure how robust this is, but since it cannot be reproduced easily, we see how it goes. At least now the integration test does not fail with the delay in response stream is removed.

### DIFF
--- a/crates/jstz_node/src/sequencer/db.rs
+++ b/crates/jstz_node/src/sequencer/db.rs
@@ -1,5 +1,6 @@
 use std::{fs, path::PathBuf};
 
+use anyhow::Context;
 use anyhow::Result;
 use r2d2::{Pool, PooledConnection};
 use r2d2_sqlite::SqliteConnectionManager;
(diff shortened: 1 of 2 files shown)


crates/jstz_node/src/sequencer/db.rs
```diff
@@ -34,21 +35,36 @@ impl Db {
...
         let pool = SqliteConnectionPool::new(manager)?;
-        Self::create_table(pool.clone())?;
+        Self::setup(pool.clone())?;
 
         Ok(Db { pool })
     }
 
     pub fn connection(&self) -> Result<PooledConnection<SqliteConnectionManager>> {
-        Ok(self.pool.get()?)
+        let conn = self
```
ryutamago, line 50: q: not directly related to this PR but is there a way to reconnect when there is a connection error w the DB?

ryutamago (approved):

crates/jstz_node/src/sequencer/db.rs
huancheng-trili: `busy_timeout` should handle retrying when a connection fails due to locking. Other errors need to be handled by the caller.

PR #1118, opened 2025-06-20
fix(docs):  upgrade to Docusaurus 3.8.1 and enable future-proofing settings

timothymcmackin:
Docusaurus has released some performance-related features and feature flags to prep for Docusaurus 4. We can enable these now to make sure we're ready for version 4.

I've tested the performance features locally and they do not seem to improve local builds much, but supposedly they will improve the speed of Vercel builds (see [Preserving the cache](https://docusaurus.io/blog/releases/3.8#persistent-cache)).

You can verify that the site appears the same by checking the preview: https://timothymcmackin.github.io/previews/jstz/docusaurus-381/

This fix removes some styles for links because in 3.8.1 they override styles for breadcrumbs and the in-page TOC. These links should not be blue but these high-level css rules make them blue. Removing these styles keeps the site looking exactly the same as before.


docs/src/css/custom.css
```diff
@@ -89,14 +89,6 @@ body {
   font-family: "Roboto", sans-serif;
 }
 
-a {
```
huancheng-trili, line 92: Why are these lines removed?

huancheng-trili (approved):

PR #1133, opened 2025-06-24
feat(docs): add docs on switching networks

timothymcmackin:
Basic info on switching networks when we have other networks available.

This example code includes a private key but this key is already known as one of the bootstrap accounts.


docs/architecture/networks.md
```diff
@@ -0,0 +1,127 @@
...
+Like Tezos layer 1 and most other Web3 systems, you can deploy and test smart functions on test networks before deploying to production.
+
+:::note
+
+Jstz does not yet have a stable production network.
+Until it does, consider all networks to be temporary test networks.
+
+:::
+
+The primary network that Jstz developers use is the the local [Sandbox](/sandbox).
```
zcabter, line 14: The local sandbox isn't really a network like how ghost, test or mainnet is. Perhaps there's no need to mention it.

docs/architecture/networks.md
```diff
@@ -0,0 +1,127 @@
...
+         "jstz_node_endpoint": "<JSTZ_NODE>"
+       }
+     }
+   }
+   ```
+
+   This example uses the variable `<NETWORK_NAME>` for the local alias of the new network.
+   You can give the network any alias and then use this name as the value for the `--network` argument in Jstz commands, as in `jstz deploy dst/index.js -n <NETWORK_NAME>`.
+
+1. Bridge some tez from layer 1 on the network to your Jstz account for transaction fees.
```
huancheng-trili, line 46: I don't think this script should be in the documentation. It was meant to demonstrate how the dev wallet would work before the wallet development started. Users are supposed to interact with different networks through the proper apps, e.g. the bridging web app, faucet, etc., deployed for each network.

timothymcmackin: Removed this script and added info that assumes that there are bridges and faucets on these networks.

huancheng-trili: Sorry I forgot to reply. For our own test network on GCP, there is a bridge, though no one can actually use it because there is no way to transfer any tez to anyone's L1 account. There will be a simple L2 faucet as a temporary solution.

huancheng-trili: We talked about implementing something in the CLI tool like `jstz network add ...` to make it easier to add or remove networks without having to touch the config file manually. That needs to wait a bit.

timothymcmackin:
Great, I added the info about the default network.

PR #1135, opened 2025-06-24
feat(oracle): update jstzd to use oracle node, tests added

alanmarkoTrilitech:
# Context
[JSTZ-539](https://linear.app/tezos/issue/JSTZ-539/update-jstzd-to-support-oracle-calls)
# Description
Updates jstzd to support oracle calls and adds tests.

Todo: slightly simplifying the tests
# Manual testing
```make```


alanmarkoTrilitech:
> @alanmarkoTrilitech would you mind fixing the test (that I broke from my change)?

seems like you've fixed it by now since I don't see any test failing locally, though not sure why the build fails here

PR #1136, opened 2025-06-24
feat(orace): support timeout queue

zcabter:
# Context
Closes https://linear.app/tezos/issue/JSTZ-532/implement-requestqueue 

Needed to GC stale requests

# Description
This PR extends the senders map with a timeout value. The field has been renamed to `active_requests` and value has been re-typed to `ReqeustMetadata`. 

We use the fact that BTreeMap is an ordered Map and that RequestId is created in an incremental order. This guarantees that the order of request ids determines the order of timeouts. Hence, the first entry of the BTreeMap will be next request to time out.

Expose `gc_timeout_requests` which will garbage collect timed out requests. Senders which are drop will send a cancellation to receivers which will cause fetch to return timeout

# Manually testing the PR
`make test-v2`

### DIFF
--- a/crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
+++ b/crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
@@ -24,24 +24,19 @@ use crate::{
 static X_JSTZ_ORACLE_GAS_LIMIT: std::sync::LazyLock<ByteString> =
     std::sync::LazyLock::new(|| ByteString::from("x-jstz-oracle-gas-limit"));
 
-#[derive(Debug, Default)]
-pub struct OracleConfig {
-    gas: GasParams,
-}
-
-#[derive(Clone, Debug, Default)]
-struct GasParams {
-    protocol_fee: Gas,
-    oracle_fee: Gas,
-    spam_prevention: Gas,
-}
-
 #[derive(Debug)]
 pub struct Oracle {
     /// Oracle's public key
     public_key: PublicKey,
-    /// Sender channels for in-flight requests
-    senders: BTreeMap<RequestId, Sender<Response>>,
+    /// Holds cached metadata that is checked often
+    ///
+    /// Notes on timeout: The relationship between request id and timeout is such that
+    /// the timeout of the next rid will be at minimum equal to the timeout of the latest
+    /// rid. In addition, since rids are created through an incrementing counter, the
+    /// order of an rid is equivalent to timeout order. This means we can rely on `first_entry`
+    /// of BTreeMap to get the next timeout value (and efficiently delete it) while also having
+    /// an efficient way to delete requests by rid
+    active_requests: BTreeMap<RequestId, RequestMetadata>,
     /// Next request id
     next_request_id: RequestId,
     config: OracleConfig,
@@ -58,7 +53,7 @@ impl Oracle {
             .ok_or(OracleError::PublicKeyNotFound)?;
         Ok(Self {
             public_key,
-            senders: Default::default(),
+            active_requests: Default::default(),
             next_request_id: 0,
             config: config.unwrap_or_default(),
         })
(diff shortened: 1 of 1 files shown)


crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
```diff
@@ -24,24 +24,19 @@ use crate::{
...
-pub struct OracleConfig {
-    gas: GasParams,
-}
-
-#[derive(Clone, Debug, Default)]
-struct GasParams {
-    protocol_fee: Gas,
-    oracle_fee: Gas,
-    spam_prevention: Gas,
-}
```
zcabter, line 37: Moved to line 190 to keep things tidy

alanmarkoTrilitech: hm...I quite liked it when it was at the top, since its short its easy to find. But won't require moving it back.

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
```diff
@@ -24,24 +24,19 @@ use crate::{
...
-    /// Sender channels for in-flight requests
-    senders: BTreeMap<RequestId, Sender<Response>>,
+    /// Holds cached metadata that is checked often
+    ///
+    /// Notes on timeout: The relationship between request id and timeout is such that
+    /// the timeout of the next rid will be at minimum equal to the timeout of the latest
+    /// rid. In addition, since rids are created through an incrementing counter, the
+    /// order of an rid is equivalent to timeout order. This means we can rely on `first_entry`
+    /// of BTreeMap to get the next timeout value (and efficiently delete it) while also having
+    /// an efficient way to delete requests by rid
```
alanmarkoTrilitech, line 38: how is the relationship enforced? As i understand it timeout is something that can be chosen. So you could in theory create a late request which has however smaller timeout than all the requests before it.

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
```diff
@@ -160,6 +163,41 @@ impl Oracle {
...
         Ok(gas_limit)
     }
+
+    /// Triggers the GC for timed out requests
+    pub fn gc_timeout_requests(
+        &mut self,
+        host: &mut impl HostRuntime,
+        current_level: BlockLevel,
+    ) {
+        while let Some(head) = self.active_requests.first_entry() {
```
alanmarkoTrilitech, line 173: when will we call this?

also there is the issue related to earlier comment of why timeouts are sorted properly. Not sure how the assumption is enforced, but in any case a comment here that we have this assumption would be good

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
```diff
@@ -84,21 +79,29 @@ impl Oracle {
...
         };
         let (sender, rx) = channel();
-        if self.senders.contains_key(&request_id) {
+        if self.active_requests.contains_key(&request_id) {
             // protocol error
             return Err(OracleError::BadState("Sender should not yet exist!"));
         }
 
         // Checks have passed, we can do state updates
+        self.incr_request_id();
```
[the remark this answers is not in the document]
zcabter: yes

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
zcabter: No it can't be chosen

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
```diff
@@ -514,7 +554,41 @@ mod test {
...
+            .unwrap();
+        oracle
+            .send_request(&mut host, &mut tx, &caller, req.clone())
+            .unwrap();
+        oracle
+            .send_request(&mut host, &mut tx, &caller, req)
+            .unwrap();
+        assert_eq!(oracle.active_requests.len(), 3);
+        oracle.gc_timeout_requests(&mut host, 0);
+        assert_eq!(oracle.active_requests.len(), 0);
```
alanmarkoTrilitech, line 592: so it got rid of all the requests even though current level is 0?

crates/jstz_proto/src/runtime/v2/oracle/oracle.rs
alanmarkoTrilitech: ok, can you please add a comment in the code that we have the assumption regarding the request/ttl relation

zcabter: Isn't that on line 31?

PR #1145, opened 2025-06-26
chore(jstzd): add balance to built-in bootstrap accounts

huancheng-trili:
# Context

Part of JSTZ-709.
[JSTZ-709](https://linear.app/tezos/issue/JSTZ-709/set-activator-account-properly)

We need a mechanism to rotate bootstrap accounts while making sure that octez node can still activate the protocol successfully.

# Description

Updated the format of `accounts.json`. Now the balance of each bootstrap account is specified in that file as well. This makes it easier to set the balance of individual bootstrap account.

Each entry can actually be transformed into a proper mapping as there are a bit too many fields now, but I'm keeping it simple here temporarily.

# Manually testing the PR

All tests should still work.


ryutamago (approved):
are you planning to update the jstzd integration test? probably better to make it work once these PRs are merged

PR #1156, opened 2025-06-27
fix(sequencer): decode and encode with hex

huancheng-trili:
# Context

I was testing the zod example and found that deploying and running the smart function were very slow.

# Description

Changed the way the sequencer runtime host deals with bytes. Instead of using base58check functions, now it uses hex to encode bytes and decode strings.

# Manually testing the PR

* Unit testing: added a timeout check to an existing test. Without the changes, the test will fail as writing a byte array with 64K elements will take more than 10 seconds.


huancheng-trili:
Yes

PR #1158, opened 2025-06-29
feat(oracle): use global level to calculate ttl

zcabter:
# Context
We need to gc requests when they reach TTL based on the current block level

Part of https://linear.app/tezos/issue/JSTZ-541/handle-oracleresponse-operation-in-kernel 

# Description
* Adds read only current level to the global protocol context with mutator exposed for testing. The write mechanism will be implemented in next PR
* Update protocol context to expose Arc<Mutex<>> over each field rather than on the entire struct
* Set default TTL to 80 levels and calculate the timeout in send request. The block time in jstzd is 1s allowing 80 seconds to respond.
* Update gc to run compare against the global current level

# Manually testing the PR
`cargo nextest run -p jstz_proto oracle  --features v2_runtime`

### DIFF
--- a/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
+++ b/crates/jstz_proto/src/runtime/v2/fetch/fetch_handler.rs
@@ -255,11 +255,11 @@ fn dispatch_oracle(
         .boxed_local());
     }
     let response_rx = {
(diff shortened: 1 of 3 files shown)


alanmarkoTrilitech:
how was 80 seconds chosen?

PR #1159, opened 2025-06-30
feat(oracle): handle oracle response operation

zcabter:
# Context

# Description

# Manually testing the PR

### DIFF
--- a/crates/jstz_proto/src/executor/mod.rs
+++ b/crates/jstz_proto/src/executor/mod.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "v2_runtime")]
+use crate::{operation::OracleResponse, runtime::PROTOCOL_CONTEXT};
+
 use crate::{
     operation::{
         self, Content, InternalOperation, Operation, OperationHash, SignedOperation,
@@ -61,7 +64,17 @@ async fn execute_operation_inner(
             Err(Error::RevealTypeMismatch)
         }
         #[cfg(feature = "v2_runtime")]
-        operation::Content::OracleResponse(_oracle_response) => {
+        operation::Content::OracleResponse(OracleResponse {
+            request_id,
+            response,
+        }) => {
+            let oracle_ctx = PROTOCOL_CONTEXT.get().unwrap().oracle();
+            let mut oracle = oracle_ctx.lock();
+            if &op.public_key != oracle.public_key() {
+                panic!("Oracle keys mismatch")
+            }
+
+            let _ = oracle.respond(hrt, request_id, response);
             todo!()
         }
     }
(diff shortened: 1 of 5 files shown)


zcabter:
TODO: Tests coming

crates/jstz_proto/src/runtime/v2/protocol_context.rs
```diff
@@ -36,6 +42,12 @@ impl ProtocolContext {
         rt: &mut impl HostRuntime,
         current_level: BlockLevel,
     ) -> Result<(), ProtocolContextError> {
+        // Hardcode oracle value pk to injector pk for now
+        let oracle_key = PublicKey::from_base58(
```
alanmarkoTrilitech, line 46: why? shouldn't these be different? I don't see how is injector related to the oracle

crates/jstz_proto/src/executor/mod.rs
```diff
@@ -345,4 +376,231 @@ mod tests {
...
+                .await;
+        assert!(matches!(
+            receipt,
+            Receipt {
+                result: ReceiptResult::Failed(s),
+                ..
+            }
+            if s == "Ed25519 error: signature error".to_string()
+        ));
+    }
```
alanmarkoTrilitech, line 605: it seems like some deduplication could be done here so that the tests are also a bit shorter and easier to read

crates/jstz_proto/src/runtime/v2/protocol_context.rs
zcabter: Added fix me

crates/jstz_proto/src/lib.rs
```diff
@@ -70,15 +70,16 @@ pub mod tests {
     // For example, to inpect the body of a successful RunFunctionReceipt, you can provide the path
     // vec!["result", "inner", "body"]. If you don't really care what the return type is and just
     // want to print field value, you can parameterize with `serde_json::Value`
-    #[allow(unused)]
-    fn inspect_receipt<T: DeserializeOwned>(
+    pub fn inspect_receipt<T: DeserializeOwned>(
         host: &impl HostRuntime,
         op_hash: OperationHash,
-        path_into_receipt: Vec<String>,
+        path_into_receipt: Vec<&'static str>,
```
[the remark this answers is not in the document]
alanmarkoTrilitech: what do you mean by specific path? it's always a specific path, never some vague description of the path :D

alanmarkoTrilitech (approved):
LGTM, except that its unclear why the static string is used

crates/jstz_proto/src/lib.rs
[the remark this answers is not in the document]
zcabter: I guess the right word was "static" path. Like we never create the path dynamically

PR #1164, opened 2025-06-30
fix(docs): new algolia IDs for jstz.tezos.xom

timothymcmackin:
We were issued new Algolia IDs for jstz.tezos.com. This PR updates them so the search will work:

Preview: https://timothymcmackin.github.io/previews/jstz/new-algolia


NicNomadic (approved):
It works fine (although it also seems to be working fine in the current setting...)

PR #1170, opened 2025-06-30
feat(oracle): prevent oracle fetch if dirty transaction

alanmarkoTrilitech:
# Context
[JSTZ-700](https://linear.app/tezos/issue/JSTZ-700/prevent-oracle-fetch-if-dirty-transaction)
# Decription
This PR adds protections preventing running oracle in the case that the transaction is dirty. As a dirty we consider all changes except for changes to smart function accounts.
# Testing
```make```

### DIFF
--- a/crates/jstz_core/src/kv/transaction.rs
+++ b/crates/jstz_core/src/kv/transaction.rs
@@ -58,6 +58,8 @@ pub type Key = OwnedPath;
 #[derive(Debug, Default, Deref, DerefMut)]
 struct LookupMap(BTreeMap<Key, Vec<usize>>);
 
+pub type PendingChangesFilter = dyn Fn(&Key) -> bool;
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/jstz_core/src/kv/transaction.rs [hunk 1]
```diff
@@ -195,6 +197,27 @@ impl InnerTransaction {
...
+                    true
+                }
+            });
+            let has_remove_edits = !snapshot.remove_edits.is_empty();
+            if has_insert_edits || has_remove_edits {
+                return true;
+            }
+        }
+        false
+    }
```
zcabter, line 219: You actually need to check both reads and writes. Reads because when a value is read, it is cached in the tx snapshot. Another operation might mutate it causing the cached value to be stale

zcabter (approved):
LGTM, could you let the big PR go through first?

PR #1186, opened 2025-07-02
fix(docs): update example of signing transactions

timothymcmackin:
This part of the docs links to an example that was removed. The PR updates it to an existing example. (Is it still relevant to show a non-wallet signing example?)


huancheng-trili (approved):
I think it's okay to keep this example since it's still valid and now we even have a warning for users.

PR #1187, opened 2025-07-02
fix(docs): update and clarify install and build instructions

timothymcmackin:
Just to clarify the install instructions. They seem to work for me.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -3,19 +3,19 @@ title: Installing Jstz
 sidebar_label: Installation
 ---
 
-## Download and Install
-
-Jstz has several [components](/architecture/overview#components), but the primary tool that you need to develop on Jstz is the [Jstz CLI](/cli), which is available as the package [`@jstz-dev/cli`](https://www.npmjs.com/package/@jstz-dev/cli).
+Jstz has several [components](/architecture/overview#components), but the primary tool that you need to develop on Jstz is the [Jstz CLI](/cli), which is available as the NPM package [`@jstz-dev/cli`](https://www.npmjs.com/package/@jstz-dev/cli).
 It allows you to start the Jstz sandbox, deploy smart functions to it, and interact with them.
+
+:::note
+
+Jstz is available only on Unix-based systems.
+
+:::
 
 Follow these instructions to install the Jstz CLI:
 
-:::danger
-⚠️ `jstz` is only available on Unix-based systems. ⚠️
-:::
-
-Ensure `docker` is installed on your system. If not, please follow [this guide](https://docs.docker.com/get-docker/).
-Download and install `jstz` via NPM with this command:
+Ensure that [Docker](https://docs.docker.com/get-docker/) is installed on your system.
+Then, download and install `jstz` via NPM with this command:
 
 ```sh
 npm i -g @jstz-dev/cli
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -28,108 +28,89 @@ yarn global add @jstz-dev/cli
...
-```sh
-sudo dnf install clang
-export CC=clang
-```
+   ```sh
+   git clone https://github.com/jstz-dev/jstz.git
+   ```
 
-Linux:
+1. Install LLVM:
```
zcabter, line 48: If we're recommending nix (which I would), then install llvm, rust and octez steps are not required.

docs/installation.md
```diff
@@ -28,108 +28,89 @@ yarn global add @jstz-dev/cli
...
 
-Once Octez has been built, copy the following binaries to `jstz`:
+   ```sh
+   curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+   ```
 
-- `octez-client`
-- `octez-node`
-- `octez-smart-rollup-node`
+1. Download the following Octez binaries to the `jstz` directory:
```
huancheng-trili, line 90: The build setup is actually a bit complicated now. To make this document more organised, I think it needs to be split into 2 parts:

1. For MacOS users: nix is the only way to proceed because of octez binaries. Later on when it becomes less problematic with `brew install`, we can revisit this.
2. For Linux users: two paths here:
  * nix: `nix develop` should handle all dependencies.
  * Manual dependency installation: pretty much the existing flow. Note that the octez binaries need to be in `PATH`. Downloading them to `jstz` doesn't work unless `jstz` is in path, which I don't think any user would do.

And perhaps add a note at the very top of this doc: ~windows users cannot work with jstz~

huancheng-trili: Later on this will be even more complicated for those who don't want to use nix because of dependencies of the riscv kernel. We can come back to that later.

RichAyotte (approved):

docs/installation.md
```diff
@@ -32,70 +32,103 @@ You are now ready to [write your first smart function](/quick_start) 🚀.
...
-   Fedora:
+   You can locate the resulting built artifact at `/target/wasm32-unknown-unknown/release/jstz_kernel.wasm`.
+
+1. Build and start the sandbox by running these commands:
 
    ```sh
-   sudo dnf install clang
-   export CC=clang
+   make build-cli
+   PATH=.:$PATH cargo run --bin jstz -- sandbox start
```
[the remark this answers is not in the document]
timothymcmackin: Removed.

docs/installation.md
huancheng-trili: After checking the octez section for non-nix again, I think we need to add a bit more explanation:

* Users need to download the baker executables as well, e.g. [arm64-octez-baker-PsRiotum](https://gitlab.com/tezos/tezos/-/package_files/211363808/download). All baker executables available need to be downloaded. For example, for octez v23, both `octez-baker-PsRiotum` and `octez-baker-PtSeouLo` need to be downloaded. In the future (probably very soon) we will use the agnostic baker, e.g. [arm64-octez-baker](https://gitlab.com/tezos/tezos/-/package_files/211363338/download), which is going to replace baker executables for each protocol, which is going to be simpler for users.
* For each executable, users need to
  * download the desirable executable according to their machine architecture
  * rename the artefact and remove the architecture prefix, e.g. from `arm64-octez-baker-PsRiotum` to `octez-baker-PsRiotum`
  * make the artefact executable, i.e. `chmod +x octez-baker-PsRiotum`
  * ensure that the renamed executable is in `$PATH`

docs/installation.md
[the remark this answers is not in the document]
timothymcmackin: Got it, removed the `PATH=.:$PATH` because the step about the Octez binaries makes it clear that they myst be in `PATH`.

docs/installation.md
timothymcmackin: Updated with more detailed binary instructions.

docs/installation.md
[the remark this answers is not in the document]
huancheng-trili: The example in that section needs to be running jstzd as well.

docs/installation.md
```diff
@@ -28,108 +28,122 @@ yarn global add @jstz-dev/cli
...
 
-```sh Arch
-pacman -S clang
-export CC=clang
-```
+1. Build and start the sandbox by running these commands:
 
-Nix:
+   ```sh
+   make build-jstzd-kernel
```
huancheng-trili, line 69:
```suggestion
```

Sorry, I read the whole thing again and realised that there is no need for this line given the `make build` in the previous bullet point.

docs/installation.md
```diff
@@ -28,108 +28,122 @@ yarn global add @jstz-dev/cli
...
+   ```
+
+   You can locate the resulting built artifact at `target/wasm32-unknown-unknown/release/jstz_kernel.wasm`.
+
+1. Build and start the sandbox by running these commands:
 
-This will initially run `octez-node` and initialize `octez-client`. Once the client is initialized, the `jstz_kernel` and `jstz_bridge`
-is originated, a `octez-smart-rollup-node` and `jstz-node` is spun up.
+   ```sh
+   make build-jstzd-kernel
```
huancheng-trili, line 146:
```suggestion
```

ditto

timothymcmackin:
> Sorry about not raising all the comments in the first place. I read the whole document again and realised that now there might be a bit of confusion. The document starts with the CLI tool and kind of suggests that it will tell users how to build the CLI tool, but now in the end it tells users to build jstzd. If you also think that this is a bit awkward, you can add one bullet point that tells users how to build the CLI tool:
> 
> ```
> make build-cli
> ```
> 
> but then we kind of have to also tell users how to use the build to interact with the sandbox, which is again a bit awkward. Perhaps this needs to be replanned and we don't have to do it in this PR.

I've gotten a little turned around over these reviews, too. I guess if it's complicated to use the built version of the CLI we can omit the steps for building it.

huancheng-trili:
No. This is because the baker executable for protocol alpha is installed via nix and the default protocol used by jstzd now is alpha. The non-nix linux setup needs that echo line to tell jstzd not to run alpha because the baker executable for alpha is not part of the tezos release and therefore not available in the non-nix setup. As I mentioned earlier, if we change the default protocol to the latest protocol _released_, we can get rid of that awkward echo line and users won't be confused.

huancheng-trili:
Yes we can merge this and add patches later. Thanks!

PR #1190, opened 2025-07-04
feat(runtime): support self on worker global scope

zcabter:
# Context
`self` is a property expected on `WorkerGlobalScope`. 

Generally, we cannot support every property on WorkerGlobalScope but we can support self which is used by libraries like Axios to access Fetch API irregardless of the environment.

See https://html.spec.whatwg.org/multipage/workers.html#dom-workerglobalscope-self

# Description
* Adds `self` to `globalThis` in `jstz_runtime`
* Adds an Axios example

# Manually testing the PR
Write jstzd config
```
cat <<EOF >  ~/jstzd-config.json
{
    "jstz_node": {
        "mode": "sequencer",
        "capacity": 1024
    }
}
EOF
```

Rebuild jstzd and run it
```
make build-jstzd-kernel && cargo run --bin jstzd --features oracle run ~/jstzd-config.json
```
Follow instructions in `examples/axios/README.md`

### DIFF
--- a/crates/jstz_runtime/src/ext/jstz_main/99_main.js
+++ b/crates/jstz_runtime/src/ext/jstz_main/99_main.js
@@ -1,3 +1,9 @@
 import { workerGlobalScope } from "ext:jstz_main/98_global_scope.js";
 
 Object.defineProperties(globalThis, workerGlobalScope);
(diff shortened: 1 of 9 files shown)


timothymcmackin (approved):
Works for me locally in nix.

PR #1194, opened 2025-07-07
refactor: move oracle to jstzd

huancheng-trili:
# Context

# Description

# Manually testing the PR


crates/jstzd/src/task/oracle_node.rs
```diff
@@ -0,0 +1,43 @@
...
+        Ok(Self { inner: oracle })
+    }
+
+    async fn kill(&mut self) -> Result<()> {
+        drop(self.inner.take());
+        Ok(())
+    }
+
+    async fn health_check(&self) -> Result<bool> {
+        Ok(true)
```
alanmarkoTrilitech, line 41: it should check some condition here

huancheng-trili: There isn't anything for health check in oracle node at this moment anyway and the purpose of this PR is not to cover that part.

alanmarkoTrilitech: then adding a todo!() or a comment with linking a linear issue would be good so that we don't forget it

alanmarkoTrilitech:
> Well, if you want to run oracle node, you need to create a main for that. It's a _node_ anyway and therefore I think it makes sense to do so if that is a _node_. Based on what you mentioned, jstz node should be bundled with the rollup node somehow because jstz node does not function on its own either. The _user experience_ does not change because of this as users don't care about what happens under the hood and the sandbox will still function as is. If you mean _developer experience_, that's a different issue and that should be handled by some other solutions, not by bundling things together just for convenience.

we can discuss this during the meeting today

alanmarkoTrilitech:
forgot about this, sorry. I think it's a bit different compared to the rollup node, since that's done by a different team and is much heavier, does't support just jstz, but also etherlink/tezlink, so it's less appropriate to bundle together.  However for Jstz, I don't see why Oracle node couldn't be just an optional extension of the jstz node for simpler setup rather having different separate nodes, one for each smaller feature. 

@zcabter what do you think?

zcabter:
@alanmarkoTrilitech It should be separate to the jstz node because one shouldn't concern themselves with configuring the oracle if they don't intend to operate one. This is also a nice separation of concerns.

Additionally, the oracle node must respond to the canonical sequencer logs but we shouldn't assume that the oracle and sequencer are going to be run by the same entity.

alanmarkoTrilitech (approved):
LGTM, can you please either link a task for the oracle node CLI or if if you decided to work on it mention it here so that we can track it.

huancheng-trili:
[JSTZ-786](https://linear.app/tezos/issue/JSTZ-786/create-executable-for-oracle-node)

PR #1195, opened 2025-07-07
feat(oracle): jsbuffer body serialisation

alanmarkoTrilitech:
# Context
[JSTZ-766](https://linear.app/tezos/issue/JSTZ-766/fix-oracle-request-serialisation)

# Desctiption
Request body was incorrectly serialised/deserialised when provided via JsBuffer. This PR fixes it by properly serialising JsBuffer and deserialising it into Vec<u8>. This enables running an example with Openai API call.

# Manual testing
```make test-v2```
Deploy and run oracle_post_openai.js test (for best results with your own API key, however "Incorrect API key provided" also counts as working).

### CODE REVIEW

File: examples/oracle/oracle_post_openai.js [hunk 1]
```diff
@@ -0,0 +1,20 @@
+const handler = async () => {
```
zcabter, line 1: This shouldn't be in `examples/`. Move to `js_tests` instead

crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -510,4 +543,33 @@ mod test {
         .map(|(k, v)| (k.as_bytes().into(), v.as_bytes().into()));
         assert_eq!(expected, *res);
     }
+
+    #[test]
+    fn body_buffer_roundtrip() {
```
zcabter, line 548: Would be good to also add a response round trip

crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -196,6 +197,38 @@ impl<'s> ToV8<'s> for Body {
...
+
+impl From<BodyDeserializeHelper> for Body {
+    fn from(value: BodyDeserializeHelper) -> Self {
+        let bytes = match value {
+            BodyDeserializeHelper::Vector(b) | BodyDeserializeHelper::Buffer(b) => b,
+        };
+        Body::Vector(bytes)
+    }
+}
+
```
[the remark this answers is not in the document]
alanmarkoTrilitech: the idea here is that you get ```{"Vector":[…]}``` or ```{"Buffer":[…]}``` in the serialised output rather than just the Vec so that we know where it came from, which is useful for logging/testing/debugging.

crates/jstz_proto/src/runtime/v2/fetch/http.rs
alanmarkoTrilitech: sure, added also a response buffer roundtrip test

zcabter: I don't see the change

zcabter: could you also add a bincode round trip?

crates/jstz_proto/src/runtime/v2/fetch/http.rs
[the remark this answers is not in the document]
alanmarkoTrilitech: done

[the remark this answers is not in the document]
alanmarkoTrilitech: ok, it wasn't like this before though. Updated now.

crates/jstz_proto/src/runtime/v2/fetch/http.rs
```diff
@@ -407,13 +427,54 @@ mod test {
         };
         let json = serde_json::to_string(&request.clone()).unwrap();
         assert_eq!(
-            r#"{"status":200,"status_text":"OK","headers":[[[107,49],[118,49]],[[107,50],[118,51]],[[107,50],[118,50]],[[107,49],[118,52]]],"body":{"Vector":[123,34,109,101,115,115,97,103,101,34,58,34,104,101,108,108,111,34,125]}}"#,
```
zcabter, line 410: 👍

zcabter:
If you're going to move `oracle_basic.js` into its own folder, you also need to update the doc and add a `README.`. I suggest just leaving it as is

zcabter (changes requested):
If you're going to move `oracle_basic.js` into its own folder, you also need to update the doc (see `docs/functions/api.md` and add a `README` I suggest just leaving it as is

zcabter (approved):

PR #1207, opened 2025-07-08
feat(kernel): warm up runtime on kernel start

zcabter:
# Context
V8 platform setup takes a few seconds to setup.

Launching a dummy runtime warms the v8 platform up such that future runtime creation will be much faster

# Manually testing the PR
`make riscv-kernel-pvm`


alanmarkoTrilitech (approved):
LGTM!

what does running much faster next time mean? is it less than a second?

PR #1212, opened 2025-07-11
chore(runtime): test api coverage

huancheng-trili:
# Context

Part of JSTZ-771.
[JSTZ-771](https://linear.app/tezos/issue/JSTZ-771/comparison-doc-generated-from-wpt)

# Description

Created one integration test that checks how many node APIs the runtime covers. The logic is basically borrowed from [cloudflare/workers-nodejs-compat-matrix](https://github.com/cloudflare/workers-nodejs-compat-matrix). This test requires a library script and a baseline data file. To minimise the amount of code checked in, both are supposed to be fetched manually beforehand. They are added to the nix test build.

# Manually testing the PR

```sh
cd $(git rev-parse --show-toplevel)/crates/jstz_runtime
./tests/api_coverage/setup.sh
cargo test --test api_coverage
```

### DIFF
--- a/crates/jstz_runtime/README.md
+++ b/crates/jstz_runtime/README.md
@@ -1,3 +1,16 @@
 # Jstz Runtime
 
 To find compatible deno extension versions, look for their version number in the [Cargo.toml](https://github.com/denoland/deno/blob/v2.1.10/Cargo.toml#L74) of `deno_core:v2.1.0`
+
+### API coverage test
+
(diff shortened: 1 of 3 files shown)


zcabter:
Why bother with tracking node support? Jstz will report 0% since we don't support Node.js at all..

Its more useful to just compare Web APIs. IIRC, there was a comparison table that you generated in the past. It would probably be enough to anchor that into our docs somewhere

zcabter:
Could you update `jstz_runtime/README.md` with instructions for running the tests?

zcabter (approved):

PR #1214, opened 2025-07-11
feat(oracle): executable added

alanmarkoTrilitech:
# Context
[JSTZ-786](https://linear.app/tezos/issue/JSTZ-786/create-executable-for-oracle-node)

# Description
After a recent change in #1194, jstz oracle node isn't part of jstz node and thus isn't part of any exacutable (besides jstzd). This PR creates s CLI for the oracle node.

# Manual testing
```make``` , running the CLI

### DIFF
--- a/crates/jstz_oracle_node/src/main.rs
+++ b/crates/jstz_oracle_node/src/main.rs
@@ -6,6 +6,7 @@ use env_logger::Env;
 use jstz_crypto::{public_key::PublicKey, secret_key::SecretKey};
 #[cfg(feature = "v2_runtime")]
 use jstz_oracle_node::node::OracleNode;
+use serde::Deserialize;
 
 const DEFAULT_JSTZ_NODE_ENDPOINT: &str = "http://127.0.0.1:8933";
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: crates/jstz_oracle_node/src/main.rs [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+
+const DEFAULT_LOG_PATH: &str = "logs/kernel.log";
+const DEFAULT_JSTZ_NODE_ENDPOINT: &str = "http://127.0.0.1:8933";
+
+#[derive(Debug, Parser)]
+#[command(name = "jstz-oracle-node")]
+#[command(about = "JSTZ Oracle Node - Provides oracle data for JSTZ rollup")]
+struct Args {
+    /// Path to the log file
+    #[arg(long, default_value = DEFAULT_LOG_PATH)]
```
zcabter, line 18: I'm not sure we should default this. If we do, it should be the same the jstzd default. However, the default for them is to create a temp file so its not known up front.

crates/jstz_oracle_node/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+    if !args.log_path.exists() {
+        anyhow::bail!("Log path does not exist: {:?}", args.log_path);
+    }
+
+    // Parse key file
+    let (public_key, _secret_key) =
+        parse_key_file(args.key_file).context("failed to parse key file")?;
+
+    log::info!("Starting JSTZ Oracle Node");
+    log::info!("Log path: {:?}", args.log_path);
```
zcabter, line 57: Also, it'll be nice to print (and use) the absolute path with [canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html)

crates/jstz_oracle_node/src/main.rs
alanmarkoTrilitech: done

crates/jstz_oracle_node/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+    {
+        let _oracle_node =
+            OracleNode::spawn(args.log_path, public_key, _secret_key, args.node_endpoint)
+                .await
+                .context("Failed to spawn oracle node")?;
+
+        log::info!("Oracle node started successfully");
+
+        // Keep the node running
+        // The node will keep running until this task is dropped
```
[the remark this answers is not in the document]
alanmarkoTrilitech: done

crates/jstz_oracle_node/src/main.rs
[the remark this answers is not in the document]
alanmarkoTrilitech: done

crates/jstz_oracle_node/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// JSTZ node endpoint
+    #[arg(long, default_value = DEFAULT_JSTZ_NODE_ENDPOINT)]
+    node_endpoint: String,
+
+    /// Path to file containing key pair (format: "public_key:secret_key")
+    #[arg(long)]
+    key_file: PathBuf,
+}
+
+fn parse_key_file(path: PathBuf) -> anyhow::Result<(PublicKey, SecretKey)> {
```
zcabter, line 29: Sorry for commenting late but thinking about this again, our convention is use json files. So t think that's what this should be

zcabter (approved):

PR #1218, opened 2025-07-15
chore(nix): bump  rust to 1.88.0

zcabter:
# Context
The latest `riscv-sandbox` crate depends on rust 1.86 and above. We might as well bump it up to the latest stable version.

Part of https://linear.app/tezos/issue/JSTZ-792/expose-riscv-sandbox-with-huge-memory-in-nix
Closes https://linear.app/tezos/issue/JSTZ-793/upgrade-rust-to-1880

# Description
Bumps rust version to 1.88.0  and as a result causes lots of clippy errors that needs to to be fixed. The main clippy error is inlining string interpolation.

# Manually testing the PR
```
direnv allow
make
```
Unfortunately, code coverage tis not happy with clippy changes but no additional tests 😓

### DIFF
--- a/crates/jstz_runtime/.gitignore
+++ b/crates/jstz_runtime/.gitignore
@@ -1,1 +1,4 @@
 tests/deno_report.json
+tests/api_coverage/baseline.js
+tests/api_coverage/utils.js
+tests/api_coverage/output.json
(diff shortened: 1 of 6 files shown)


johnyob (approved):
Tested build and tests locally, lgtm 🚀

PR #1220, opened 2025-07-15
chore(nix): upgrade riscv sandbox to build with huge-memory feature

zcabter:
# Context
Jstz requires the `huge-memory` feature to be enabled on riscv-sandbox

Closes https://linear.app/tezos/issue/JSTZ-792/expose-riscv-sandbox-with-huge-memory-in-nix

# Description
* Updates nix to bring `riscv-sandbox` with `huge-memory` enabled
* Adds manual testing artifacts

# Manually testing the PR
```
direnv allow
make run-manual-test
```


zcabter:
> lgtm, tested locally 🚀
> 
> Slight nit, I find the archive approach a bit odd (why not just use a directory? we'll delete it anyway) -- but its temporary any 🤷

Agreed. I've replaced the archive with just the directory

PR #1227, opened 2025-07-23
chore(core): explicitly add lib atomic as dependency

huancheng-trili:
# Context

Completes JSTZ-811.
[JSTZ-811](https://linear.app/tezos/issue/JSTZ-811/explicitly-add-lib-atomic-as-dependency)

# Description

Added `build.rs` to `jstz_core` to make it explicitly link the atomic library when the build target is riscv64.

# Manually testing the PR

Tested with the following cases:
* CI
* `make riscv-pvm-kernel`
* created a [dummy rust project](https://github.com/huancheng-trili/test-kernel-build) that depended on jstz_kernel and built it successfully

### DIFF
--- a/crates/jstz_core/build.rs
+++ b/crates/jstz_core/build.rs
@@ -1,8 +1,13 @@
-use std::env;
-
 fn main() {
-    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
     if target_arch == "riscv64" {
+        // Note: jstz_core does not really need to link atomic. This is actually for
+        // jstz_runtime and jstz_proto as they depend on deno and v8 needs to link
+        // atomic when the build target is riscv. The actual reason is not clear, but
+        // one theory is that atomic needs to be linked before rusty_v8 starts to be
+        // compiled, which takes place before both crates get compiled, which means
+        // that linking atomic in build.rs of those crates is too late. Since both
+        // crates depend on jstz_core, linking atomic here works.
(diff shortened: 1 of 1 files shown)


johnyob (approved):

zcabter:
It should probably be propagated via `jstz_runtime`?

huancheng-trili:
> why jstz_core? I understood that the linking errors for atomics came from v8?

Indeed the errors are from v8. I'm not sure what happens under the hood actually. Even though I don't think this is how it works, my guess is that the library has to be linked _before_ `jstz_runtime` and `jstz_proto` start to be compiled (though build.rs is already run before the actual compilation.) Since both crates depend on `jstz_core`, linking the library when `jstz_core` is being compiled satisfies this condition. I tried linking the library for `jstz_crypto` (depended on by both crates as well) instead and it also worked. This probably means that this is not the proper way of doing this, but for now it works.

huancheng-trili:
That makes sense. I've added some comments to explain why this is added to `jstz_core`.

zcabter:
> It is indeed coming from rusty_v8. I'll upstream the patch to our fork when the build environment is ready. We can keep this till then
Actually,  it wasn't a problem with `rusty_v8`. It is using wrong linking flag. `cargo::rustc-link-lib` sends the `-l` flag to libs  only but we need to link the binary. Use`cargo:rustc-link-arg=-latomic` instead

huancheng-trili:
`cargo:rustc-link-arg=-latomic` does not work when `jstz_kernel` is being imported as a library.

huancheng-trili:
`cargo:rustc-link-arg=-latomic` does not work when `jstz_kernel` is imported as a foreign library.

PR #1237, opened 2025-07-28
feat(utils): implement event stream

ryutamago:
# Context

Part of: [807](https://linear.app/tezos/issue/JSTZ-807/event-eventstream)
Closes: [808](https://linear.app/tezos/issue/JSTZ-808/implement-eventstream-for-oracle)

# Description

* Introduce `EventStream` in jstz utils that filters the logs from file and decode into `Event` type. It can be extended to handle stream over the network in the future.

* Make oracle relay use `EventStream`.

* Move `FilteredLogStream` from `jstz_oracle_node` to `jstz_utils`.

# Manually testing the PR

* unit tests
```
cargo test --package jstz_utils
```
* manually tested using jstzd and `oracle_basic.js`

### DIFF
--- a/crates/jstz_core/src/event.rs
+++ b/crates/jstz_core/src/event.rs
@@ -144,6 +144,16 @@ mod test {
         }
     }
 
+    fn mock_event() -> MockEvent {
+        MockEvent {
+            id: 1,
+            caller: PublicKeyHash::from_base58("tz1XSYefkGnDLgkUPUmda57jk1QD6kqk2VDb")
+                .unwrap(),
+            gas_limit: 100,
+            url: Url::from_str("http://example.com/foo").unwrap(),
+        }
+    }
+
     #[test]
     fn test_publish_decode_roundtrip() {
         let mut sink = Sink(Vec::new());
@@ -153,13 +163,7 @@ mod test {
                 &mut sink.0,
             )
         });
-        let event = MockEvent {
-            id: 1,
-            caller: PublicKeyHash::from_base58("tz1XSYefkGnDLgkUPUmda57jk1QD6kqk2VDb")
-                .unwrap(),
-            gas_limit: 100,
-            url: Url::from_str("http://example.com/foo").unwrap(),
-        };
+        let event = mock_event();
         EventPublisher::publish_event(&host, &event).unwrap();
         let head_line = sink.lines().first().unwrap().clone();
         assert_eq!(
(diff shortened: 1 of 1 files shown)


zcabter:
Nit: PR title should adhere to `commitlint` format. For one commit PRs, use the commit title

crates/jstz_utils/src/event_stream.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl<'a, E: Event> EventStream<'a, E> {
+    /// Create an event stream from a file.
+    ///
+    /// The file is expected to contain lines matching the given pattern.
+    /// The pattern is a regular expression that will be used to filter the lines.
+    /// The lines that match the pattern will be decoded into event of type `E`.
+    /// The lines that do not match the pattern will be ignored.
+    pub async fn from_file(pattern: Regex, path: PathBuf) -> Result<Self> {
```
zcabter, line 30: `from_file` doesn't need a `pattern` argument. A regex that matches "[Event::tag()] prefix followed by any characters" can be computed here

crates/jstz_utils/src/event_stream.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        match self.log_stream.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok(line))) => {
+                match decode_line::<E>(&line) {
+                    Ok(event) => Poll::Ready(Some(Ok(event))),
+                    Err(_) => Poll::Pending, // Ignore noises.
```
zcabter, line 51: The line passes the regex parse but failed the event parse which is not noise but likely a jstz error. We should `log::warn!()` the error to tell the user something wasn't right

crates/jstz_utils/src/event_stream.rs
zcabter, line 1: `jstz_oracle_node::request::request_event_from_log_line` is no longer used. It should be removed and its tests should be incorporated here

crates/jstz_utils/src/event_stream.rs
```diff
@@ -0,0 +1,153 @@
...
+pub type LogStream<'a> = BoxStream<'a, anyhow::Result<String>>;
+
+/// A stream of Jstz Events, decoded from a log stream.
+pub struct EventStream<'a, E: Event> {
+    log_stream: LogStream<'a>,
+    _marker: PhantomData<E>,
+}
+
+impl<'a, E: Event> EventStream<'a, E> {
+    /// Create an event stream from a file.
```
zcabter, line 24:
```suggestion
    /// Create an [`EventStream`] from a kernel log file.
```

crates/jstz_utils/src/event_stream.rs
```diff
@@ -0,0 +1,153 @@
...
+        Regex::new(r#"^\[MOCK\]\s*(?P<json>\{.*\})\s*$"#).expect("hard-coded regex")
+    }
+
+    #[tokio::test]
+    async fn picks_up_new_request() -> anyhow::Result<()> {
+        let tmp = NamedTempFile::new()?;
+        let path = tmp.path().to_path_buf();
+
+        let stream = EventStream::from_file(regex(), path.clone()).await?;
+        futures_util::pin_mut!(stream);
```
zcabter, line 121: Can be removed

crates/jstz_utils/src/tailed_file.rs
```diff
@@ -22,3 +34,199 @@ impl TailedFile {
...
     }
 }
+
+/// A stream that “tails” a file and yields only the lines that match the supplied regular expression.
+///
+/// * Opens the file in *follow* mode, emits only lines that are appended afterwards.
+/// * Reads until the end of the file, then waits a bit and tries again.
+///
+/// Dropping `FilteredLogStream` calls `cancel()` and closes the file.
+pub struct FilteredLogStream {
```
zcabter, line 44: I think its appropriate to reduce visibility to`(crate)`

zcabter (changes requested):

crates/jstz_utils/src/event_stream.rs
ryutamago: removed it, the event test covers the decoding bit so I don't think we need to add tests

crates/jstz_utils/src/event_stream.rs
```diff
@@ -0,0 +1,174 @@
...
+    Regex::new(&format!(r"^\[{}\].*$", E::tag())).map_err(anyhow::Error::from)
+}
+
+impl<'a, E: Event> EventStream<'a, E> {
+    /// Create an [`EventStream`] from a kernel log file.
+    ///
+    /// The file is expected to contain lines matching the given pattern.
+    /// The pattern is a regular expression that will be used to filter the lines.
+    /// The lines that match the pattern will be decoded into event of type `E`.
+    /// The lines that do not match the pattern will be ignored.
```
zcabter, line 33:
```suggestion
    /// The file is expected to be kernel log file
    ///  Lines that match prefixed with `E::tag()` will be decoded into event of type `E`. Otherwise, they will be ignored
```

PR #1238, opened 2025-07-29
feat(env): benchmark peak memory uasge with heaptrack

zcabter:
# Context
Need a benchmark for peak memory usage

Closes https://linear.app/tezos/issue/JSTZ-805/benchmark-v8-baseline-peak-memory-in-risc-v-sandbox

# Description
Uses `heaptrack` to get peak memory usage from `riscv-sandbox`. The sandbox only allocates memory that was `mmap`'ed by the kernel application. 

We are interested in `peak RSS` which is the peak Resident Set Size ie. physical memory in RAM

# Manually testing the PR
Note that this only works on linux machines. Given that nix is install, 
```
# Comment out octez in our flake.nix for faster set up
nix develop
./scripts/heaptrack.sh
```


alanmarkoTrilitech (approved):
LGTM, haven't test it though

PR #1246, opened 2025-08-01
fix(tps_bench): bench run uses dirname for path resolution

zcabter:
# Context
Allows internal paths to resolve correctly when `jstz_tps_bench/run.sh` from any root path

# Description
Sets base directory of relative paths to the directory of the script. Also add kernel compilation as the first step.

# Manually testing the PR
```
./crates/jstz_tps_bench/run.sh
```
```
cd crates && ./jstz_tps_bench/run.sh
```
```
cd crates/jstz_tps_bench && ./run.sh
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ crates/jstzd/resources/jstz_rollup/jstz_kernel.wasm
 # Heaptrack raw data
 *.zst
 
+*.log
+inbox.json
--- a/crates/jstz_tps_bench/run.sh
+++ b/crates/jstz_tps_bench/run.sh
@@ -6,14 +6,29 @@ inbox_file_path=./inbox.json
 n_transfer=10
 log_file_path=./output.log
 result_path=./result.log
+dir="$(realpath $(dirname "$0"))"
+
+case "${DISABLE_BUILD}" in
+1 | true | yes) ;;
+*)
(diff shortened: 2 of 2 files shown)


huancheng-trili (commented):
Why is this necessary? The readme states that this script is supposed to be run from the root of `jstz_tps_bench`. I would rather leave kernel build out as this means that the benchmark job will build the kernel once more.

zcabter:
> Why is this necessary? The readme states that this script is supposed to be run from the root of `jstz_tps_bench`.

Why wouldn't we want to enable running this from anywhere? 

> I would rather leave kernel build out as this means that the benchmark job will build the kernel once more.

Cargo caches the build. It won't rebuild it if it was built before.

huancheng-trili:
I don't see the point of having to make this runnable from anywhere. Your assumption is that the kernel is built on the same machine where the test will be run, which holds for now but will change in the future.

alanmarkoTrilitech:
I don't see an issue with having $dir there. While yes, readme might say that it should be run from jstz_tps_bench, the script becomes simpler to use if its user doesn't have to remember this. The script already uses its parent directories so isn't sufficiently decoupled at this point from the rest of the codebase for it to make sense to require running it only from its folder. Ideally it should be possible to provide kernel path to the script.

Regarding the kernel build, sure it could be built somewhere else, at this point thought it seems like a fairly small change that could be reverted at any point. I would suggest adding there a flag, e.g. DISABLE_BUILD which could be used later if we want to just run bench without rebuild.

alanmarkoTrilitech (approved):
:ship:

PR #1248, opened 2025-08-01
fix(wpt): remove `setTimeout` and `clearTimeout`

huancheng-trili:
# Context

Part of JSTZ-772.
[JSTZ-772](https://linear.app/tezos/issue/JSTZ-772/fix-wpt)

# Description

Removed `setTimeout` and `clearTimeout` from the runtime used in WPT. Details are in the comments. In short, the custom implementation of these two functions that throws an exception fails the test setup and therefore needs to be overwritten.

# Manually testing the PR

Ran WPT locally. The test still failed but with a different error message.


alanmarkoTrilitech (approved):
LGTM, it manages now to run some FileAPI tests and then fails.

Should there be a task for later adding it back or enabling timer so that its not forgotten?

PR #1249, opened 2025-08-04
feat(bench): implement inbox builder

huancheng-trili:
# Context

Part of JSTZ-841.
[JSTZ-841](https://linear.app/tezos/issue/JSTZ-841/implement-test-framework-for-dispatching-operations)

# Description

Implemented a struct `InboxBuilder` that builds jstz operations into inbox messages.

# Manually testing the PR

* Unit testing: added tests.

### DIFF
--- a/crates/jstz_tps_bench/src/builder.rs
+++ b/crates/jstz_tps_bench/src/builder.rs
@@ -1,0 +1,294 @@
+use http::{HeaderMap, Method, Uri};
+use jstz_core::BinEncodable;
+use jstz_crypto::{
+    hash::Hash, keypair_from_mnemonic, public_key::PublicKey, secret_key::SecretKey,
+    smart_function_hash::SmartFunctionHash,
+};
+use jstz_proto::{
+    context::account::{Address, Nonce},
+    operation::{Content, DeployFunction, Operation, RunFunction, SignedOperation},
+    runtime::ParsedCode,
+};
+use tezos_data_encoding::enc::BinWriter;
+use tezos_smart_rollup::{
+    inbox::{ExternalMessageFrame, InboxMessage},
+    michelson::MichelsonUnit,
+    types::SmartRollupAddress,
+    utils::inbox::file::{InboxFile, Message},
+};
+
+// tag + 20 byte address
+const EXTERNAL_FRAME_SIZE: usize = 21;
+const DEFAULT_GAS_LIMIT: u32 = 100_000;
+const MNEMONIC: &str =
+    "donate kidney style loyal nose core inflict cup symptom speed giant polar";
+
+pub struct Account {
+    nonce: Nonce,
+    sk: SecretKey,
+    pk: PublicKey,
+    pub address: Address,
+}
+
+pub struct InboxBuilder {
+    messages: Vec<Message>,
+    rollup_address: SmartRollupAddress,
+}
+
+impl InboxBuilder {
+    pub fn new(rollup_address: SmartRollupAddress) -> Self {
+        Self {
+            rollup_address,
+            messages: Vec::new(),
+        }
+    }
+
+    pub fn build(self) -> InboxFile {
+        InboxFile(vec![self.messages])
+    }
+
+    pub fn create_accounts(count: usize) -> crate::Result<Vec<Account>> {
+        let mut accounts = vec![];
+        for i in 0..count {
+            let (pk, sk) = keypair_from_mnemonic(MNEMONIC, &i.to_string())?;
+            let account = Account {
+                address: Address::from_base58(&pk.hash())?,
+                sk,
+                pk,
+                nonce: Default::default(),
+            };
+            accounts.push(account);
+        }
+        Ok(accounts)
+    }
+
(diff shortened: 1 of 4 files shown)


crates/jstz_tps_bench/src/builder.rs
```diff
@@ -0,0 +1,294 @@
+use http::{HeaderMap, Method, Uri};
```
alanmarkoTrilitech, line 1: I think it would be better to place this outside of jstz_tps_bench in e.g. jstz_utils as it seems to have more general use-case

crates/jstz_tps_bench/src/builder.rs
```diff
@@ -0,0 +1,294 @@
...
+        method: Method,
+        headers: HeaderMap,
+        body: Option<Vec<u8>>,
+    ) -> crate::Result<()> {
+        let content = Content::RunFunction(RunFunction {
+            uri,
+            method,
+            headers,
+            body,
+            gas_limit: DEFAULT_GAS_LIMIT.try_into()?,
```
alanmarkoTrilitech, line 133: gas_limit could be an Option<u32> arg which if set to None results in DEFAULT_GAS_LIMIT

crates/jstz_tps_bench/src/builder.rs
```diff
@@ -0,0 +1,294 @@
...
+};
+
+// tag + 20 byte address
+const EXTERNAL_FRAME_SIZE: usize = 21;
+const DEFAULT_GAS_LIMIT: u32 = 100_000;
+const MNEMONIC: &str =
+    "donate kidney style loyal nose core inflict cup symptom speed giant polar";
+
+pub struct Account {
+    nonce: Nonce,
```
alanmarkoTrilitech, line 27: don't we have a definition like this somewhere else already? I find it weird to define something like this in this file. If you want to keep it here I would suggest renaming to something else like AccountSecrets, AccountDetails or something like that

crates/jstz_tps_bench/src/builder.rs
huancheng-trili: I think it's fine to keep it here for now until something else in another crate is going to reference this.

crates/jstz_tps_bench/src/builder.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    #[test]
+    fn create_accounts() {
+        let accounts = InboxBuilder::create_accounts(10).unwrap();
+        let addresses = accounts
+            .iter()
+            .map(|v| v.address.clone())
+            .collect::<HashSet<_>>();
+        assert_eq!(addresses.len(), 10);
```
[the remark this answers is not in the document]
huancheng-trili: The hash set ensures that the accounts all have different addresses.

crates/jstz_tps_bench/src/builder.rs
alanmarkoTrilitech: I still think it would be good to move this to jstz_utils now. In this way it becomes more visible and in general proactively separating useful tools like this results in lower code duplication across different parts of the codebase.

huancheng-trili: I see your point, but I'm still not convinced that this needs to be moved now. This thing and the entire crate itself is, in my opinion, pretty much a standalone tool, and we need this inbox builder simply because testing the kernel relies on inbox messages. If there should be any use case where one needs to build inbox messages elsewhere, what can happen is
* they are aware of this piece of logic and move it to jstz_utils -> fine
* they are not aware of this inbox builder and implement something similar
  * if that is caught during code review, bullet point one will be carried out
  * if that is checked in and someone else later realises the duplication, some refactoring will be carried out and this will be moved to jstz_utils

Either way, eventually duplication, if any, will be cleaned up and this builder will be taken care of _if someone needs to reuse this_. On the other hand, if no one ever needs to use this, this will sit in jstz_utils as if being dumped there for no reason, which is not good for organising code either.

Another thing about moving this to utils at this moment is that there is no guarantee that the future reference will not require any sort of customisation. If customisation is needed, the logic moved to utils will be updated and perhaps eventually a copy with some slight touch needs to be added to this crate anyway. I would rather keep this here until this really needs to be shared and take actions accordingly.

huancheng-trili: _Eventually_ someone will find out. If no one finds out and nothing bad ever happens, that won't count as an issue anyway. Even if this is placed in a debatably more visible place, it's not guaranteed that this will be immediately discovered. There will always be some time and effort wasted except in the best case, which is basically rare.

Unless you have a concrete idea of how it is going to be used or there is any trace that this piece of code is going to be used elsewhere soon, I won't assume that this is going to be so useful that it needs to be placed in utils. If you have any, just make it clear.

I have moved the code anyway since I find this kind of debate completely unnecessary.

alanmarkoTrilitech: > I have moved the code anyway since I find this kind of debate completely unnecessary.
nice :D thank you!

PR #1253, opened 2025-08-05
refactor(jstz_node): pass by reference in encode_operation

huancheng-trili:
# Context

Part of JSTZ-844.
[JSTZ-844](https://linear.app/tezos/issue/JSTZ-844/directly-send-inbox-messages-to-jstz-node-for-testing)

# Description

Updated the signature of function `encode_operation` such that injector and value store can be passed in by reference. This allows injector instances and store instances to be reused, which is essential for #1254.

# Manually testing the PR

All existing tests still work.


crates/jstz_node/src/services/utils.rs
```diff
@@ -20,22 +22,23 @@ pub async fn worker_health(State(state): State<AppState>) -> impl IntoResponse {
...
         }
     }
 
-    pub async fn get_value(self, key: String) -> anyhow::Result<Option<Vec<u8>>> {
+    pub async fn get_value(&self, key: String) -> anyhow::Result<Option<Vec<u8>>> {
         Ok(match self {
             Self::Rollup(rollup_client) => rollup_client.get_value(&key).await?,
             Self::Db(db) => {
-                match tokio::task::spawn_blocking(move || db.read_key(&key))
+                let copy = db.clone();
```
alanmarkoTrilitech, line 40: is it necessary to clone here?

alanmarkoTrilitech (approved):
LGTM! Please check whether the clone is necessary before merging

crates/jstz_node/src/services/utils.rs
huancheng-trili: Yes. `db` needs to be moved to the task but it cannot be moved. This is the main reason of changing `StoreWrapper::Db` to `StoreWrapper::Db(Arc<Db>)`.

PR #1254, opened 2025-08-05
feat(jstz_node): inject inbox message through operation api

huancheng-trili:
# Context

Completes JSTZ-844.
[JSTZ-844](https://linear.app/tezos/issue/JSTZ-844/directly-send-inbox-messages-to-jstz-node-for-testing)

# Description

Created an API endpoint in jstz node for injecting operations encoded into inbox messages. The allows feeding in more types of operations without having to go through the rollup node. This is for testing only and sits behind a new feature flag `inject_inbox`. This feature flag should never be used in production.

One major thing that needs to be improved here is that the ticketer address and jstz rollup address used to check the inbox messages are hardcoded now, which means that the inbox messages passed in needs to be generated with the exact same addresses. This problem is not just for the sequencer but overall for inbox message parsing. These two addresses should and will be made configurable soon.

# Manually testing the PR

* Generate inbox messages with the benchmark suite with feature flag v2_runtime
* Run jstz node in sequencer mode with v2 runtime
* Call the new API endpoint with the inbox file, i.e. `curl -XPOST localhost:8933/operations/inbox -H 'content-type: application/json' -d @inbox.json`
* Observe the lines in the debug log file

Operations were executed successfully as expected.


crates/jstz_node/src/services/operations.rs
```diff
@@ -187,13 +201,70 @@ async fn insert_operation_queue(
...
+                insert_operation_queue(&queue, op).await?;
+            }
+            Ok(())
+        }
+        _ => Err(ServiceError::BadRequest(
+            "injecting inbox messages directly is only available in sequencer mode"
+                .to_string(),
+        )),
+    }
+}
```
alanmarkoTrilitech, line 266: would it be possible to create an automated test for this?

PR #1258, opened 2025-08-07
feat(bench): build l1 deposit inbox message

huancheng-trili:
# Context

Part of JSTZ-841.
[JSTZ-841](https://linear.app/tezos/issue/JSTZ-841/implement-test-framework-for-dispatching-operations)

# Description

# Manually testing the PR

### DIFF
--- a/crates/jstz_tps_bench/src/bin/all_operations.rs
+++ b/crates/jstz_tps_bench/src/bin/all_operations.rs
@@ -1,0 +1,40 @@
+use std::path::Path;
+
+use anyhow::Context;
+use jstz_utils::inbox_builder::InboxBuilder;
+
+use clap::Parser;
+use tezos_crypto_rs::hash::ContractKt1Hash;
+use tezos_smart_rollup::types::SmartRollupAddress;
+
+#[derive(Parser, Debug)]
+#[command(version, about, long_about = None)]
+struct Args {
+    /// Target rollup address.
+    #[arg(long)]
+    rollup_address: String,
+
+    /// Jstz ticketer contract address.
+    #[arg(long)]
+    ticketer_address: String,
+
+    /// Path to the output inbox file.
+    #[arg(long, default_value = "inbox.json")]
+    inbox_file: Box<Path>,
+}
+
+/// Generates inbox messages for most of the operations available in Jstz.
+fn main() -> jstz_tps_bench::Result<()> {
+    let args = Args::parse();
+    let rollup_addr = SmartRollupAddress::from_b58check(&args.rollup_address)
+        .context("failed to parse rollup address")?;
+    let ticketer_addr = ContractKt1Hash::from_base58_check(&args.ticketer_address)
+        .context("failed to parse ticketer address")?;
+    let mut builder = InboxBuilder::new(rollup_addr, Some(ticketer_addr));
+    let mut accounts = builder.create_accounts(2)?;
+
+    builder.deposit_from_l1(&mut accounts[0], 1000000)?;
+
+    let inbox = builder.build();
+    inbox.save(&args.inbox_file)
+}
--- a/crates/jstz_utils/src/inbox_builder.rs
+++ b/crates/jstz_utils/src/inbox_builder.rs
@@ -10,11 +10,16 @@ use jstz_proto::{
     runtime::ParsedCode,
 };
 use std::error::Error;
+use tezos_crypto_rs::hash::ContractKt1Hash;
 use tezos_data_encoding::enc::BinWriter;
 use tezos_smart_rollup::{
-    inbox::{ExternalMessageFrame, InboxMessage},
-    michelson::MichelsonUnit,
-    types::SmartRollupAddress,
+    inbox::{ExternalMessageFrame, InboxMessage, InternalInboxMessage, Transfer},
+    michelson::{
+        ticket::{FA2_1Ticket, Ticket},
+        MichelsonContract, MichelsonNat, MichelsonOption, MichelsonOr, MichelsonPair,
+        MichelsonUnit,
+    },
+    types::{Contract, PublicKeyHash, SmartRollupAddress},
     utils::inbox::file::{InboxFile, Message},
 };
 pub type Result<T> = std::result::Result<T, Box<dyn Error>>;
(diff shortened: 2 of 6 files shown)


crates/jstz_tps_bench/src/bin/all_operations.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Jstz ticketer contract address.
+    #[arg(long)]
+    ticketer_address: String,
+
+    /// Path to the output inbox file.
+    #[arg(long, default_value = "inbox.json")]
+    inbox_file: Box<Path>,
+}
+
+fn main() -> jstz_tps_bench::Result<()> {
```
alanmarkoTrilitech, line 28: is there any reason for using this Result type instead of just usual anyhow Result?

crates/jstz_utils/src/inbox_builder.rs
```diff
@@ -24,6 +29,13 @@ const EXTERNAL_FRAME_SIZE: usize = 21;
 const DEFAULT_GAS_LIMIT: u32 = 100_000;
 const MNEMONIC: &str =
     "donate kidney style loyal nose core inflict cup symptom speed giant polar";
+type DepositInboxMsgPayloadType = MichelsonOr<
```
alanmarkoTrilitech, line 32: this type alias. was already defined at multiple places, would be better to have it unified and use one of them

crates/jstz_utils/src/inbox_builder.rs
```diff
@@ -145,6 +162,46 @@ impl InboxBuilder {
...
+    pub fn deposit_from_l1(
+        &mut self,
+        account: &Account,
+        amount_mutez: u64,
+    ) -> Result<()> {
+        match &self.ticketer_address {
+            Some(ticketer) => {
+                let payload: DepositInboxMsgPayloadType =
+                    MichelsonOr::Left(MichelsonPair(
+                        MichelsonContract(Contract::Implicit(
```
alanmarkoTrilitech, line 175: also having a function for defining this type would make it more readable

crates/jstz_tps_bench/src/bin/all_operations.rs
huancheng-trili: Fundamentally it's because of the error type used in the inbox builder module. As of why that module used its own type, I have no idea, but in any case, changing (or perhaps improving) the error type is not part of that refactoring. This new binary simply follows what is already defined. This can be patched later on if we want to do that.

crates/jstz_utils/src/inbox_builder.rs
huancheng-trili: I'm not going to do that for now. Moving that type and all the relevant inbox things is too much and that introduces more dependency issues.

crates/jstz_utils/src/inbox_builder.rs
huancheng-trili: I don't think this is going to help much. Calling that helper function means passing in 3 "not seemingly clear" arguments for this payload and readers will still need to track down what happens. The function itself is not too long and does not involve many steps and therefore I consider this manageable.

alanmarkoTrilitech: I am not arguing that its going to "help much", but it will help a little since the function would have a clear name, args would be named and it overall makes the deposit_from_l1. it takes like 3 mins to do so I don't see why not to do it

crates/jstz_utils/src/inbox_builder.rs
```diff
@@ -145,6 +162,46 @@ impl InboxBuilder {
...
+                        // any user address is okay here since L1 is not really involved
+                        source: PublicKeyHash::from_b58check(
+                            "tz1W8rEphWEjMcD1HsxEhsBFocfMeGsW7Qxg",
+                        )
+                        .unwrap(),
+                        destination: self.rollup_address.clone(),
+                        payload,
+                    }));
+                let mut bytes = Vec::new();
+                msg.serialize(&mut bytes)?;
```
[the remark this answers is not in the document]
alanmarkoTrilitech: its about wrapping it in the Internal, serializing (two lines as you also need to mke that bytes vec) and putting it into Message::Raw()

alanmarkoTrilitech (commented):
Well you replied to 5 out of 5 suggestions that you don't like them or don't want to do them. Its of course fine to disagree and discuss, but I feel like you are a bit too defensive here. I can just comment to that that I still think it would be nice to make some improvement in this area, but don't think its particularly productive trying to persuade you about it or spent too much effort writing arguments for it. Yes, most suggestions won't make a ground-breaking difference, but I thought they would make some small improvement, which in worst case would be just neutral from your point of view. I think it would be much more productive/smoother if you either decided to do at least some of that them (as it takes just few minutes), or made a suggestion to do something slightly different, made a linear ticket to keep it as a task for later, etc.

huancheng-trili:
The thing is, I don't see those as improvements. Reusing logic just because "something identical is already implemented somewhere" is not going to simplify things. A small amount of duplication is not an unforgivable sin, not to mention that sometimes it's rather problematic to move around things. My point has always been that "improvements" should only be carried out when necessary. There is no point implementing changes that are not necessarily going to be used. If later on they become necessary, they will be carried out _then_.

crates/jstz_tps_bench/src/bin/all_operations.rs
alanmarkoTrilitech: can you please link a linear task for this so that we can track it?

crates/jstz_utils/src/inbox_builder.rs
alanmarkoTrilitech: also link a linear task for this one too please

alanmarkoTrilitech:
I think its overall good to make some improvements to a code that was already merged when we touch it as otherwise those small things can compound overtime and lower the quality standard. But as I've written in the previous comment I'm fine with you making a linear ticket to track it as a task for later, if its not a priority of this PR.

crates/jstz_tps_bench/src/bin/all_operations.rs
huancheng-trili: What is to be tracked? There is no obvious reason for changing this error. I don't think looking into whether this should be changed should be a ticket.

alanmarkoTrilitech (approved):
LGTM! 🚀

PR #1263, opened 2025-08-11
feat(jstz_node): add storage_sync_db

ryutamago:
# Context

Part of: [812](https://linear.app/tezos/issue/JSTZ-812/implement-storageupdate-consumer-in-jstz-node)
Next PR: [1264](https://github.com/jstz-dev/jstz/pull/1264)

In the following PR [1264](https://github.com/jstz-dev/jstz/pull/1264), a storage update event consumer is spawned when `storage_sync` is true. This consumer will process the events and update the `storage_sync_db` . In order to read the valyes from the correct place, this PR extends the `StoreWrapper` to account for the `storage_sync_db`.
# Description

* `StoreWrapper` now takes `storage_sync` and `storage_sync_db` as a parameter
* `StoreWrapper` becomes the `storage_sync` DB when `storage_sync` is true and on default mode

# Manually testing the PR

Added unit test
```
cargo test --package jstz_node
```

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -45,8 +45,8 @@ pub struct AppState {
     pub queue: Arc<RwLock<OperationQueue>>,
     pub runtime_db: sequencer::db::Db,
     worker_heartbeat: Arc<AtomicU64>,
-    #[allow(dead_code)]
     storage_sync: bool,
+    storage_sync_db: sequencer::db::Db,
 }
 
 impl AppState {
@@ -111,11 +111,7 @@ pub async fn run(options: RunOptions) -> Result<()> {
     })));
 
     // will make db_path configurable later
-    let db_file = NamedTempFile::new()?;
-    let db_path = db_file.path().to_str().ok_or(anyhow::anyhow!(
-        "failed to convert temp db file path to str"
-    ))?;
-    let runtime_db = sequencer::db::Db::init(Some(db_path))?;
+    let (runtime_db, _runtime_db_file) = temp_db()?;
     let worker = match mode {
         #[cfg(not(test))]
         RunMode::Sequencer {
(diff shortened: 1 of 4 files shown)


ryutamago:
> I don't quite understand what you are trying to achieve here. `StoreWrapper` is readonly and its only purpose is to read data from different sources based on run mode. This has nothing to do with your consumer that only writes.

The consumer writes to the local db when `storage_sync` is enabled. 
I'm extending the `StoreWrapper` so it will account for both the `run_mode` and the `storage_sync` so that it will read the values form the local db when `storage_sync` is enabled.

ryutamago:
Ok. Just to double check the behavior: if we introduce this temporary db, say `storage_sync_db`, would the consumer: 
* update the `storage_sync_db` on the `riscv` sequencer mode & the default mode (do nothing to the `runtime_db`)
* do nothing on the `native` sequencer mode (as `runtime_db` is being synced as it is)

crates/jstz_node/src/lib.rs
```diff
@@ -45,8 +45,8 @@ pub struct AppState {
     pub queue: Arc<RwLock<OperationQueue>>,
     pub runtime_db: sequencer::db::Db,
     worker_heartbeat: Arc<AtomicU64>,
-    #[allow(dead_code)]
     storage_sync: bool,
+    storage_sync_db: sequencer::db::Db,
```
huancheng-trili, line 49: I understand what you are trying to do, but perhaps this should be `Option<Db>` because it should only exist when `storage_sync` is true? I know the same thing applies to queue and runtime db and I want to clean them up later.

I know in `StoreWrapper` there will be one more check for this but it's better overall. Always initialising `storage_sync_db` means that there is always one more connection pool idling, especially in tests.

ryutamago: I was going to clean it up later including that later after the event consumer is in place. I think we could skip this for now given that we don't know how long we need the `storage_sync_db` until the core team implements kv on risdv.

huancheng-trili (approved):

PR #1265, opened 2025-08-11
feat(wpt): riscv testing

alanmarkoTrilitech:
# Context
[JSTZ-880](https://linear.app/tezos/issue/JSTZ-880/running-wpt-tests-in-riscv-sandbox)
# Description
This PR creates pipeline for running WPT tests in riscv sandbox when ```wpt-in-riscv``` flag is set. It splits each tests into multiple chunks of at most 2000 chars and sends them as messages to a dedicated kernel which combines them together and produces TestHarnessReport. The kernel with respective inbox messages gets execucted within the riscv-sandbox and the report is read from its debug logs. The rest of the pipeline processes the report as before. 

The total pass rate in riscv sandbox is 3.12% vs 3.33% when running natively.

# Manual testing
```cd jstz_runtime && STATS_PATH=$(pwd)/out.txt cargo test --features wpt-in-riscv --test wpt -- --nocapture ```

### CODE REVIEW

File: crates/jstz_kernel/src/inbox.rs [hunk 1]
```diff
@@ -132,16 +132,19 @@ pub fn parse_inbox_message(
...
+                                jstz_rollup_address, address.hash()
+                            ),
                         );
                         None
                     } else {
                         match read_external_message(logger, contents) {
                             Some(msg) => Some(Message::External(msg)),
                             None => {
-                                logger.write_debug(
-                                    "Failed to parse the external message\n",
```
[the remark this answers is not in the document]
alanmarkoTrilitech: most likely isn't since everything else includes it, will fix it

crates/jstz_runtime/tests/wpt.rs
```diff
@@ -26,248 +27,112 @@ use std::{
...
-                    "failed to convert value '{s}' into TestsStatus",
-                ))
-            })
+#[cfg(feature = "wpt-in-riscv")]
+fn run_wpt_test_harness_in_riscv_sandbox(source: String) -> TestHarnessReport {
+    use jstz_proto::runtime::ParsedCode;
+    use jstz_utils::inbox_builder::InboxBuilder;
+    use tezos_smart_rollup::types::SmartRollupAddress;
+
+    // Build inbox message using the same logic as the former external binary
```
zcabter, line 52: What is the "former external binary" that you're referring to?

crates/jstz_runtime/tests/wpt.rs
```diff
@@ -26,248 +27,112 @@ use std::{
...
+    let mut inbox_builder = InboxBuilder::new(
+        SmartRollupAddress::from_b58check("sr1BxufbqiHt3dn6ahV6eZk9xBD6XV1fYowr")
+            .unwrap(),
+        None,
+    );
+
+    let chunk_size = 2000usize;
```
zcabter, line 59: Could probably do `3915` which is the max size the jstz node uses for regular payloads.

crates/jstz_kernel/src/riscv_wpt_test_kernel.rs
```diff
@@ -0,0 +1,96 @@
...
+                        _ => {}
+                    }
+                }
+                _ => {}
+            },
+            None => {
+                break;
+            }
+        }
+    }
```
zcabter, line 71: Single match paths should almost always be `if lets`

```suggestion
   while let Some(ParsedInboxMessage::JstzMessage(Message::External(signed_operation))) =
        read_message(rt, &ticketer)
    {
        let operation: Operation = signed_operation.into();
        if let Content::DeployFunction(deploy_function) = operation.content {
            if deploy_function.function_code.to_string() == "STOP" {
                break;
            }
            source += &deploy_function.function_code.to_string();
        }
    }
```

crates/jstz_runtime/tests/report_parser.rs
```diff
@@ -0,0 +1,152 @@
...
+                Err(ParseError::Ron(
+                    "RON parse failed and JSON-unescape also failed".into(),
+                ))
+            }
+        }
+    }
+}
+
+/// Convert to RON and attempt to parse
+fn try_ron_from_debug(debug_text: &str) -> Result<TestHarnessReport, String> {
```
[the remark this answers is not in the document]
zcabter: Ah ok its because you are printing `TestHarnessReport` at the end of the test as debug right? What about if `Serialize` was derived for `TestHarnessReport` instead?  Then you could serialize into a flat json line and parse it with `serde`, avoiding  most of the custom logic.

crates/jstz_runtime/tests/wpt.rs
alanmarkoTrilitech: I had a binary for doing the chunking before to avoid some dependency issues, now removed. Will update the comment to describe the chunking.

crates/jstz_kernel/src/riscv_wpt_test_kernel.rs
alanmarkoTrilitech: not exactly like this since you can also have an internal message, but I've updated it to the extend its possible

crates/jstz_runtime/tests/wpt.rs
alanmarkoTrilitech: changed to 3800 as the operation type itself also adds some bytes. I think there's no need to figure out exactly what's the maximal boundary here

crates/jstz_runtime/tests/test_harness_api.js
```diff
@@ -16,13 +16,6 @@ Object.defineProperty(globalThis, "location", {
 delete globalThis.setTimeout;
 delete globalThis.clearTimeout;
 
-Object.defineProperty(globalThis, "self", {
```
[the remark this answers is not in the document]
alanmarkoTrilitech: the tests error when its there. This is related to https://github.com/jstz-dev/jstz/pull/1247/files that defines "self" in 99_main.js. Not sure now precisely why redefining it also here makes it work, buthaving it defined at two different places, with different enumerable/configurable/writable settings makes it fail.

crates/jstz_kernel/Cargo.toml
```diff
@@ -18,6 +18,11 @@ name = "kernel-executable"
 path = "src/executable.rs"
 required-features = ["riscv_kernel"]
 
+[[bin]]
```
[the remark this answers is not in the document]
zcabter: There is no relationship between the wpt kernel and the jstz kernel. The wasm and riscv kernel have strong relationship, they represent the program that we deploy on L1 whereas the wpt kernel is just a wrapper around wpt pipeline, not a different flavour of an actual kernel

crates/jstz_runtime/tests/report_parser.rs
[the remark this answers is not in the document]
alanmarkoTrilitech: done

crates/jstz_kernel/Cargo.toml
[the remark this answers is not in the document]
alanmarkoTrilitech: done

crates/jstz_kernel/src/lib.rs
```diff
@@ -16,19 +16,21 @@ pub mod parsing;
 #[cfg(feature = "riscv_kernel")]
 pub mod riscv_kernel;
 
-#[cfg(not(feature = "riscv_kernel"))]
```
zcabter, line 19: This change is no longer needed?

crates/jstz_kernel/src/lib.rs
```diff
@@ -69,7 +71,7 @@ pub async fn handle_message(
 // kernel entry
 #[entrypoint::main]
 pub fn entry(rt: &mut impl Runtime) {
-    #[cfg(not(feature = "riscv_kernel"))]
+    #[cfg(not(any(feature = "riscv_kernel")))]
```
zcabter, line 74: This change is no longer needed?

crates/jstz_kernel/Cargo.toml
```diff
@@ -33,6 +35,10 @@ serde.workspace = true
 tezos-smart-rollup.workspace = true
 tezos_crypto_rs.workspace = true
 tezos_data_encoding.workspace = true
+tezos-smart-rollup-mock.workspace = true
+regex.workspace = true
+
```
zcabter, line 40:
```suggestion
```

PR #1271, opened 2025-08-13
refactor(proto): use durable storage to verify nonces

johnyob:
# Context

Part of the work on CTSL (linear project not filled out yet). 

# Description

Changes nonce verification to read and write directly from durable storage instead of using a transaction. 

The PR introduces `Account::storage_*` functions to access storage directly. 
These are dual to the existing transaction functions we already have. 
It might be nice in future to unify these functions under some form of 'Kv/Storage' trait. 

# Manually testing the PR

Refactor, run existing test-suite: 

```sh
make test-v2
```


zcabter (approved):
LGTM

PR #1277, opened 2025-08-14
feat(ci): job for running wpt tests in risc-v sandbox nightly

alanmarkoTrilitech:
# Context
[JSTZ-882](https://linear.app/tezos/issue/JSTZ-882/ci-for-wpt-tests-in-riscv-sandbox)

# Description
This PR adds job for running wpt tests in riscv sandbox nightly. The execution of tests is using riscv-sandbox and thus needs to be run in nix environment.

# Testing
[Test run](https://github.com/jstz-dev/jstz/actions/runs/17913034691/job/50928727330?pr=1277)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,36 +47,6 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-  wpt_riscv:
-    name: Web Platform Test (RISC-V)
-    runs-on: wpt
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          submodules: recursive
-      - name: Build RISCV WPT kernel
-        run: nix --accept-flake-config --log-format raw -L develop -j auto --command make riscv-wpt-test-kernel
-      - name: Setup WPT test server
-        run: |
-          cd crates/jstz_wpt/wpt && python3 wpt make-hosts-file >> /etc/hosts
-          python3 wpt serve &
-      - name: Download Deno test baseline
-        run: |
-          cd crates/jstz_runtime
-          curl -s --output tests/deno_report.json https://storage.googleapis.com/wptd-results/e78446e34a1921371658a5df08c71d83f50a2a2f/deno-2.1.10_4921411-linux-unknown-fccd901f99/report.json
-      - name: Execute WPT tests
-        run: |
-          nix --accept-flake-config --log-format raw -L develop -j auto --command sh -c '
-          cd crates/jstz_runtime
-          STATS_PATH=$(pwd)/out.txt cargo test --features wpt-in-riscv --test wpt -- --nocapture
-          '
-      - name: Collect and display test results
-        run: |
-          cd crates/jstz_runtime
-          while read line; do
-            echo "$line" >> $GITHUB_STEP_SUMMARY
-          done < $(pwd)/out.txt
-
   build-docs:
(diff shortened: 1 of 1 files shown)


huancheng-trili:
Please do not trigger test runs with `test-nightly.yml`.

huancheng-trili:
You can patch the CI workflow to run your job. Also, I don't think we need a new job just for WPT in RISCV. Updating the existing WPT job should be sufficient.

.github/workflows/test-nightly.yml
```diff
@@ -6,28 +6,29 @@ on:
...
 
 jobs:
-  wpt:
-    name: Web Platform Test
-    runs-on: ubuntu-latest
-    container:
-      image: rust:1.82-slim
+  wpt_riscv:
+    name: Web Platform Test (RISC-V)
+    runs-on: [macOS, ARM64]
```
zcabter, line 11: Why can't we use ubuntu machines?

.github/workflows/test-nightly.yml
```diff
@@ -6,28 +6,29 @@ on:
...
           python3 wpt serve &
           cd ../../jstz_runtime
           curl -s --output tests/deno_report.json https://storage.googleapis.com/wptd-results/e78446e34a1921371658a5df08c71d83f50a2a2f/deno-2.1.10_4921411-linux-unknown-fccd901f99/report.json
-          STATS_PATH=$(pwd)/out.txt cargo test --test wpt
+          export PATH="$PATH:$(nix --accept-flake-config --log-format raw -L develop -j auto --command which riscv-sandbox | xargs dirname)"
+          STATS_PATH=$(pwd)/out.txt cargo test --features wpt-in-riscv --test wpt -- --nocapture
           while read line; do
             echo "$line" >> $GITHUB_STEP_SUMMARY
           done < $(pwd)/out.txt
+          '
```
zcabter, line 31: It may not be obvious for future readers why this needs to be run within the Nix context. Having the rationale of it included in the descrption would be helpfil.

It's also a bit odd that the step is called run but there's actually a lot of stuff going on, I'm not immediately sure what "Run" entails (ie. sets up wpt test server, gets the test set from deno, executes the wpt tests, collects the result, prints the comparison table?). Is it possible to separate into multiple steps?

alanmarkoTrilitech: updated the description and split the job into multiple steps, while removing nix where it seemed its not needed. I will test it once we agree that it looks correct as it takes quite long to run.

zcabter: Looks good overall.

.github/workflows/test-nightly.yml
zcabter: We shutdown most machines to save $$. I think only 1 CI machine runs. And one more machine gets spun up a short time in the wee hours of the morning. We usually only use non-ubuntu machines to build the multi-arch docker images.

alanmarkoTrilitech: ok, will switch it back to the ubuntu one

alanmarkoTrilitech: it doesn't work either on ubuntu one or [x86_64, linux, nix], only on [macOS, ARM64]

zcabter: Why doesn't it work? Do you have a link to the error?

zcabter (approved):
LGTM 🚢

PR #1285, opened 2025-08-18
feat(docs): basic info about using privatenet

timothymcmackin:
Basic info about Privatenet. Should the URLS for the network go in the table?


huancheng-trili:
Do we want to expose privatenet now? If so, do we want to give it a better name (including a new domain name)?

PR #1286, opened 2025-08-19
 feat(jstz_node): add death signal to storage sync

ryutamago:
# Context

Part of: [JSTZ-812](https://linear.app/tezos/issue/JSTZ-812/implement-storageupdate-consumer-in-jstz-node)

Next PR: #1291 

We need to shutdown the jstz node in case storage sync dies, this PR is a preparation step for it.
# Description

* Adds a death signal so that the caller can know when storage sync dies

* Implement a future to `StorageSync` so that when `StorageSync` dies it resolves (check #1291  on how it's used)

# Manually testing the PR

unit tests

```
cargo test --package jstz_node --lib --features v2_runtime -- test::server_dies_when_storage_sync_dies --exact --show-output
```

### DIFF
--- a/crates/jstz_node/src/storage_sync.rs
+++ b/crates/jstz_node/src/storage_sync.rs
@@ -1,21 +1,43 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 use std::thread;
 use std::{path::PathBuf, time::Duration};
 
 use crate::sequencer::{self, db::Db};
-use anyhow::{Context, Result};
+use anyhow::{Context as _, Result};
 use futures_util::StreamExt;
 use jstz_core::kv::storage_update::{BatchStorageUpdate, StorageUpdate};
 use jstz_utils::event_stream::EventStream;
 use jstz_utils::retry::{exponential_backoff, retry_async};
 use log::{error, warn};
-use tokio::sync::oneshot::{self, Sender};
+use tokio::sync::oneshot::{self, Receiver, Sender};
 
+/// A handle to a long-running background worker that consumes an event stream and
+/// applies storage updates to the database.
+///
+/// When this handle is dropped, it sends a shutdown signal to the worker and waits for the thread to terminate.  
+/// Awaiting the handle completes when the worker exits. If the worker exits unexpectedly (e.g., due to a DB error or stream failure),
+/// it returns an error.
+/// Note: Dropping this handle does **not** initiate a shutdown; it only observes and waits for the worker's termination.
+///
 pub struct StorageSync {
-    kill_sig: Option<Sender<()>>,
-    handle: Option<thread::JoinHandle<()>>,
+    inner: Option<thread::JoinHandle<()>>,
+    kill_tx: Option<Sender<()>>,
+    death_rx: Receiver<Result<()>>,
 }
 
-/// Spawns a new storage sync thread.
+impl Future for StorageSync {
+    type Output = Result<()>;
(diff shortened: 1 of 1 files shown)


huancheng-trili:
Fix the PR title please.

crates/jstz_node/src/storage_sync.rs
```diff
@@ -1,21 +1,43 @@
...
+/// - **`.await` it → unexpected death.**  
+///   Awaiting resolves only if the worker dies on its own (e.g. DB error, stream failure).  
+///   It is not a way to shut the worker down.
+///
 pub struct StorageSync {
-    kill_sig: Option<Sender<()>>,
-    handle: Option<thread::JoinHandle<()>>,
+    inner: Option<thread::JoinHandle<()>>,
+    kill_tx: Option<Sender<()>>,
+    death_rx: Receiver<()>,
```
huancheng-trili, line 29: `death_rx` should carry information about why/how the worker terminated.

crates/jstz_node/src/storage_sync.rs
```diff
@@ -55,16 +80,17 @@ pub fn spawn(
                                 match next_item {
                                     Some(Ok(updates)) => {
                                         if let Err(e) = apply_batch_tx_with_retry(&db, updates).await {
-                                            error!("storage_sync: db error, aborting: {e}");
+                                            error!("db error, aborting: {e}");
                                             break;
                                         }
+                                        info!("applied batch of updates");
```
huancheng-trili, line 86: I don't think this is going to be helpful as one knows if updates are applied when they are reflected in the data store. Logs will be flooded with these not very helpful lines that do not provide much information.

crates/jstz_node/src/storage_sync.rs
```diff
@@ -1,21 +1,43 @@
...
-use log::{error, warn};
-use tokio::sync::oneshot::{self, Sender};
+use log::{error, info, warn};
+use tokio::sync::oneshot::{self, Receiver, Sender};
 
+/// A long-running background worker that consumes an event stream and
+/// applies storage updates to the database.
+///
+/// - **Drop it → graceful shutdown.**  
+///   Dropping the handle stops the worker and waits for the thread to exit.
```
huancheng-trili, line 20: I think this line needs to be rephrased as it doesn't seem semantically correct.

crates/jstz_node/src/storage_sync.rs
ryutamago: I agree that death_rx should ideally carry the error message, and I initially tried that approach as well. However, it’s not particularly useful in this context because the Future implementation for StorageSync cannot propagate the error to the caller—this is due to the API of axum’s graceful_shutdown, which requires the future’s Output type to be (). (See the referenced PR for details on how it’s used.)

The main purpose of death_rx is to provide a mechanism for the caller to detect whether StorageSync is still running, so that we can react accordingly and shut down the server if necessary. Additionally, StorageSync is intended to run indefinitely; if it does terminate, the error is already logged, so there’s little practical benefit to returning the error message via death_rx.

ryutamago: ok. I will go for the first option and make the death_rx propogate the error then

crates/jstz_node/src/storage_sync.rs
```diff
@@ -55,16 +80,17 @@ pub fn spawn(
                                 match next_item {
                                     Some(Ok(updates)) => {
                                         if let Err(e) = apply_batch_tx_with_retry(&db, updates).await {
-                                            error!("storage_sync: db error, aborting: {e}");
```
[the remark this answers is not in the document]
ryutamago: file where the caller resides.

PR #1298, opened 2025-08-26
refactor(jstz_node): implement shutdown for log service

ryutamago:
# Context

# Description

# Manually testing the PR


crates/jstz_node/src/lib.rs
```diff
@@ -199,8 +196,7 @@ pub async fn run(options: RunOptions) -> Result<()> {
     let listener = TcpListener::bind(format!("{addr}:{port}")).await?;
     axum::serve(listener, router).await?;
 
-    cancellation_token.cancel();
-    tail_file_handle.await.unwrap()?;
+    log_service_handle.shutdown().await?;
```
ryutamago, line 199: I don't think we need to propagate the error to the parent like we do for storage sync as long as we print out the error - wdyt about implementing `AsyncDrop` instead of calling the shutdown ? @huancheng-trili

huancheng-trili:
Why does this need to be done now?

ryutamago:
> Why does this need to be done now?

It doesn't have to be done now specifically but since I was looking at the cleanup logic of jstz node, I cleaned it up.

crates/jstz_node/src/lib.rs
huancheng-trili: This log service will be obsolete soon anyway so I don't think this needs to be further refined.

ryutamago: Ok I won't furthur refactor - since this is done can we still merge this branch?

huancheng-trili (approved):

PR #1299, opened 2025-08-26
chore(octez): upgrade octez version to v23.0

zcabter:
# Context

NL has archived the v22 octez binaries from their package repository and now support v23 instead. This means we must reflect the version bump in our release software and should also use the same version in our dev environment.

Closes https://linear.app/tezos/issue/JSTZ-897/bump-octez-to-v23 

# Description
* Bump version in docker
* Bump version in nix and `nix flake lock --update-input octezPackages` to update local store
* Re-generated the octez nix patches
* Removed Quebec, added Seoul protocol
* Rollup wasm debugger is no longer an octez executable output and we don't use it anyway.. so removed from `OCTEZ_EXECUTABLES`

# Manually testing the PR
`make` and CI passes

Tested manually with get-tez


crates/octez/resources/protocol_parameters/sandbox/PtSeouLouXkxhg39oWzjxDWaCydNfR3RxCUrNe4Q9Ro8BTehcbh
```diff
@@ -0,0 +1,141 @@
+{
+  "consensus_rights_delay": 2,
+  "blocks_preservation_cycles": 1,
+  "delegate_parameters_activation_delay": 2,
+  "tolerated_inactivity_period": 2,
+  "blocks_per_cycle": 900,
```
ryutamago, line 6: in the previous settings, this was 8. Does jstzd locally work well with this settings?

PR #1302, opened 2025-08-27
 feat(sequencer): enable storage sync

ryutamago:
# Context

Closes: [899](https://linear.app/tezos/issue/JSTZ-899/integrate-storage-sync-for-sequencer)

The sequencer should be able to sync storage updates even when running in RISC-V mode. Currently, storage syncing only works on default mode. This PR extends the functionality so that kernel storage update events can be consumed by the sequencer in RISC-V mode as well.

# Description

* Added support for storage sync in sequencer RISC-V mode.
* The sequencer now listens for kernel storage update events and applies them to its local state `storage_sync_db`.

Note: the StoreWrapper continues to read from runtime_db, not storage_sync_db. The eventual synchronization between the sequencer’s debug_file and runtime_db will be handled separately once RISC-V integration is complete.

Testing
* Existing unit tests have been updated and are passing.

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -34,6 +34,8 @@ pub mod config;
 pub mod sequencer;
 pub use config::RunMode;
 
+use crate::config::RuntimeEnv;
+
 #[derive(Clone)]
 pub struct AppState {
     pub rollup_client: OctezRollupClient,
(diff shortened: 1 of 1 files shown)


crates/jstz_node/src/lib.rs
```diff
@@ -248,9 +250,17 @@ pub fn openapi_json_raw() -> anyhow::Result<String> {
 }
 
 fn validate_run_options(options: &RunOptions) -> Result<()> {
```
huancheng-trili, line 252: What's the point of keeping this check now? `storage_sync` should be independent of run mode.

ryutamago: I'm keeping it just to avoid running the sequencer in native mode w the storage sync. Can remove the entire validation logic too

huancheng-trili (approved):

PR #1316, opened 2025-09-09
feat(sequencer): use sequential stream in inbox monitor

ryutamago:
# Context

Part of: [JSTZ-574](https://linear.app/tezos/issue/JSTZ-622/handle-retrial-when-stream-connection-is-lost)

# Description

* Integrated the sequential stream to the inbox monitor process. The `process_inbox_messages` function now parse the inbox messages, push them to the queue and save the block level as `checkpoint`. In case of io error on saving the checkpoint (which is less likely to happen as we are writng to a file) It keeps retrying until the checkpoint is saved. 
* Removed prints statements from the sequential block stream.
* The`retry_expo_retries_and_succeeds` test was being flaky so I fixed it by changing the initial delay

Note that a temp file is used for now for in the sequencer as the checkpoint store. The sequencer's behavior stays the same and will not be resumable. We can revisit this later when working on persisting the sequencer's state / reliability.

# Manually testing the PR

Added unit tests

### DIFF
--- a/crates/jstz_node/src/lib.rs
+++ b/crates/jstz_node/src/lib.rs
@@ -156,10 +156,14 @@ pub async fn run(
         RunMode::Default => None,
     };
 
+    // TODO: make checkpoint path configurable
+    // https://linear.app/tezos/issue/JSTZ-912/make-inbox-checkpoint-file-configurable
+    let _checkpoint = NamedTempFile::new()?;
     let _monitor: Option<Monitor> = match mode {
         #[cfg(not(test))]
         RunMode::Sequencer { .. } => {
-            Some(inbox::spawn_monitor(rollup_endpoint, queue.clone()).await?)
+            let path = _checkpoint.path().to_path_buf();
+            Some(inbox::spawn_monitor(rollup_endpoint, queue.clone(), path).await?)
         }
         #[cfg(test)]
         RunMode::Sequencer { .. } => None,
--- a/crates/jstz_node/src/sequencer/inbox/mod.rs
+++ b/crates/jstz_node/src/sequencer/inbox/mod.rs
@@ -1,11 +1,14 @@
-#![allow(unused_variables)]
-#![allow(unreachable_code)]
+use crate::sequencer::inbox::store::{CheckpointStore, FileCheckpointStore};
+use crate::sequencer::inbox::stream::{
+    Error, PendingBlock, SequentialBlockStream, StreamFactory,
+};
 use crate::sequencer::queue::{OperationQueue, WrappedOperation};
 use crate::sequencer::runtime::{JSTZ_ROLLUP_ADDRESS, TICKETER};
 use anyhow::Result;
 use api::BlockResponse;
 use async_dropper_simple::AsyncDrop;
 use async_trait::async_trait;
+use futures_util::{StreamExt, TryStreamExt};
 use jstz_core::host::WriteDebug;
 use jstz_kernel::inbox::parse_inbox_message_hex;
 use jstz_proto::operation::internal::InboxId;
@@ -13,13 +16,13 @@ use jstz_proto::BlockLevel;
 use log::{debug, error};
 use std::collections::VecDeque;
 use std::future::Future;
+use std::path::PathBuf;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use tezos_crypto_rs::hash::{ContractKt1Hash, SmartRollupHash};
 use tokio::{select, task::JoinHandle};
 use tokio_retry2::strategy::ExponentialFactorBackoff;
 use tokio_retry2::{Retry, RetryError};
-use tokio_stream::StreamExt;
 use tokio_util::sync::CancellationToken;
 
 pub mod api;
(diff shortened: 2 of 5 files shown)


crates/jstz_node/src/sequencer/inbox/mod.rs
```diff
@@ -105,19 +102,49 @@ pub async fn spawn_monitor<
...
+                // Save as checkpoint once the last op is pushed to the queue
+                (true, true) => block.commit().await.is_ok(),
+                // continue on with the next op
+                (true, false) => true,
+                // Push the op to the queue and save as checkpoint
+                (false, true) => {
+                    queued = push(&op);
+                    queued && block.commit().await.is_ok()
+                }
+                (false, false) => push(&op),
```
zcabter, line 145: Why is this line not `queued = push(&op);`?

crates/jstz_node/src/sequencer/inbox/mod.rs
```diff
@@ -105,19 +102,49 @@ pub async fn spawn_monitor<
...
+            };
+            if exit {
                 break;
             }
             tokio::time::sleep(Duration::from_millis(100)).await;
```
zcabter, line 150: What is this sleep for btw?

crates/jstz_node/src/sequencer/inbox/stream.rs
```diff
@@ -166,8 +166,7 @@ where
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let mut this = self.project();
         loop {
-            println!("[poll_next] starting loop");
-            let store = this.store.clone();
+            let _store = this.store.clone();
```
zcabter, line 169: If unused, can remove?

crates/jstz_node/src/lib.rs
```diff
@@ -148,10 +148,13 @@ pub async fn run(
         RunMode::Default => None,
     };
 
+    // TODO: make checkpoint path configurable
```
zcabter, line 151: Track with linear please

crates/jstz_node/src/lib.rs
```diff
@@ -148,10 +148,13 @@ pub async fn run(
...
+    let _checkpoint = NamedTempFile::new()?;
     let _monitor: Option<Monitor> = match mode {
         #[cfg(not(test))]
         RunMode::Sequencer { .. } => {
-            Some(inbox::spawn_monitor(rollup_endpoint, queue.clone()).await?)
+            let path = _checkpoint.path().to_path_buf();
```
zcabter, line 156:
```suggestion
            let path = checkpoint.path().to_path_buf();
```

crates/jstz_node/src/sequencer/inbox/mod.rs
ryutamago: push returns a boolean, if true the loop is exited and will move on to the next op if false, it will keep coming back to this branch to try push

crates/jstz_node/src/sequencer/inbox/mod.rs
ryutamago: just to give it a time to wait

crates/jstz_node/src/lib.rs
```diff
@@ -148,10 +148,13 @@ pub async fn run(
         RunMode::Default => None,
     };
 
+    // TODO: make checkpoint path configurable
+    let _checkpoint = NamedTempFile::new()?;
```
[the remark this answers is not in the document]
zcabter: Could checkpoint file creation be moved to the code block on line 157?

crates/jstz_node/src/sequencer/inbox/mod.rs
```diff
@@ -105,15 +102,31 @@ pub async fn spawn_monitor<
...
     queue: Arc<RwLock<OperationQueue>>,
     ticketer: &ContractKt1Hash,
     jstz: &SmartRollupHash,
 ) {
-    let mut ops = parse_inbox_messages(block_level, block_content, ticketer, jstz);
+    let mut ops = parse_inbox_messages(block.level(), block_content, ticketer, jstz);
+
     while let Some(op) = ops.pop_front() {
         loop {
             let success = queue.write().is_ok_and(|mut q| q.insert_ref(&op).is_ok());
```
zcabter, line 132:
```suggestion
            let success = queue.write().is_ok_and(|mut q| q.insert(&op).is_ok());
```
insert to prevent unnecessary clone?

ryutamago: it's in the loop so it ends up moving

crates/jstz_node/src/lib.rs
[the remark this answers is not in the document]
zcabter: Is there a reason why the checkpoint file is exposed? Is there going to be a secondary consumer that will need access to the file in future PRs?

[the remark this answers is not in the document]
zcabter: You can make it configurable but just allow the inbox monitor to own it?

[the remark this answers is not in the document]
zcabter: The inbox monitor takes a `T: Read + Write`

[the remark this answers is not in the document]
ryutamago: I see, so instead of passing a file, we could pass an abstraction over a file-like object. Given that `CheckpointStore` already uses a path-based file approach, refactoring to a generic file-like abstraction would add a lot of complexity— including requiring everything to implement `AsyncRead` and `AsyncWrite`—for little benefit at this stage, since all we need is to store the block level (and the inbox id later) in a file. If we ever need more flexibility or want to change the backend, we can always revisit this. What do you think?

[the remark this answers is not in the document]
zcabter: Ok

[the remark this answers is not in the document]
ryutamago: updated

PR #1320, opened 2025-09-12
feat(kernel): record inbox id of parsed inbox messages

huancheng-trili:
# Context

Part of JSTZ-804.
[JSTZ-804](https://linear.app/tezos/issue/JSTZ-804/configure-riscv-pvm-in-sequencer)

Since inbox ID is used to derive operation hashes for deposit operations, inbox ID must be preserved with parsed inbox messages such that the RISCV PVM is aware of this information while loading messages.

# Description

Introduced one wrapper struct `ParsedInboxMessageWrapper` that packs the existing `ParsedInboxMessage` and an inbox message's inbox ID. To avoid unnecessary distracting changes, the awkward name `ParsedInboxMessageWrapper` is chosen instead of using `ParsedInboxMessage` and renaming the enum. Renaming can be carried out later on if necessary.

# Manually testing the PR

* Unit testing: updated existing tests to cover index IDs


ryutamago (approved):
🛳️

PR #1321, opened 2025-09-12
feat(benchmark): benchmarking with flexible transaction types

alanmarkoTrilitech:
# Context
[JSTZ-913](https://linear.app/tezos/issue/JSTZ-913/flexible-tps-benchmark)
# Description
This PR make the TPS benchmark more flexible to allow execution of different transaction types. 
In particular it splits the generate command to generate fa2 and generate other, where fa2 is used for the original mock FA2 contract and generate other is used for generation of inbox messages of other type with specified init/transfer/check endpoints. 
Additionally this PR adds the actual FA2 code which was previously not included, thus it was difficult to know/modify what we are actually benchmarking.
Lastly, it introduces a new benchmarking smart function with different endpoints for different transaction types as specified in the JSTZ TPS Benchmark doc -https://docs.google.com/document/d/1kg3QuRXQO8mtlv33VO3TwaSmaiV5t6_Gl61EzINZ8Zg/edit?usp=sharing and updates the CI job for running the benchmark

EDIT: extended with the estimation of overhead (requested by Asutosh)

# Manual testing
```./crates/jstz_tps_bench/run_all.sh```

### CODE REVIEW

File: crates/jstz_tps_bench/fa2/tsconfig.json [hunk 1]
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "types": ["@tezos/jstz-types"]
```
zcabter, line 4: I'm not aware that we have a jstz-types package

crates/jstz_tps_bench/fa2/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@tezos/fa2",
+  "authors": "TriliTech Ecosystems and Tools Team <contact@trili.tech>",
+  "private": true,
+  "version": "0.0.0",
+  "main": "fa2.ts",
+  "dependencies": {
+    "@tezos/jstz": "^0.0.0"
```
zcabter, line 8: Package doesn't exists

crates/jstz_tps_bench/tps_test/src/index.ts
zcabter, line 1: Add a docstring explaining what this file is and its strcuture

crates/jstz_tps_bench/tps_test/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+): Promise<Response> {
+  const smartFunctionAddress = Kv.get("smartFunctionAddress");
+  let response = new Response();
+  for (let i = 0; i < 200; i++) {
+    response = await SmartFunction.call(
+      new Request(`jstz://${smartFunctionAddress}/`, {
+        method: "POST",
+        body: JSON.stringify({ message: "hello" }),
+      }),
+    );
```
zcabter, line 57: Could we use the `fetch` api instead?

crates/jstz_tps_bench/tps_test/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "@tezos/jstz": "^0.0.0"
+  },
+  "devDependencies": {
+    "esbuild": "^0.20.1"
+  },
+  "scripts": {
+    "check:types": "tsc --noEmit",
+    "build": "esbuild src/index.ts --bundle --format=esm --target=esnext --minify --outfile=dist/index.js",
+    "build:test": "esbuild test/index.ts --bundle --format=esm --target=esnext --minify --outfile=dist/test/index.js",
+    "build:test-actor": "esbuild test/actor.ts --bundle --format=esm --target=esnext --minify --outfile=dist/test/actor.js"
```
zcabter, line 17: can be removed

crates/jstz_tps_bench/run_other.sh
```diff
@@ -0,0 +1,45 @@
...
+    --release \
+    --target riscv64gc-unknown-linux-musl
+  ;;
+esac
+
+cargo build --bin bench --features v2_runtime
+
+# Compile contract
+cd $contract_folder_path
+npm run build
```
zcabter, line 34: Perhaps we can simply build and check in the minified code

alanmarkoTrilitech: I like it more that it builds the contract - it i then much more convenient if I want to run it locally, instead always having to both build it and then run the script after every change. It is also quite fast once its already build so doesn't really harm anything if it gets built again.

crates/jstz_tps_bench/fa2/package.json
alanmarkoTrilitech: regarding this and other 3 fa2 related comments - I would keep it it as is simply so that we know what is the source of the minified benchmark code. Before, there was just a pointer to an old commit that contained this. I would imply just add there a comment/or a readme that the FA2 is there for legacy reasons and we don't expect to improve it further - if someone wants to improve it, they should just use the new tps benchmark. Another reason why not to modify is, is that as I understood this is the same as riscv team uses, so I think we want to keep it the same without changes.

crates/jstz_tps_bench/run_other.sh
zcabter: If its pre-built, you don't have to build it even locally. 
Also, in the bench machine, it will almost never be pre-built.

.github/workflows/run-benchmark.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Build kernel
+        run: nix --accept-flake-config --log-format raw -L develop --command make riscv-pvm-kernel
+      - uses: actions/upload-artifact@v4
+        with:
+          name: riscv_kernel
+          path: target/riscv64gc-unknown-linux-musl/release/kernel-executable
+          if-no-files-found: error
+          retention-days: 1
+
+  benchmark:
```
zcabter, line 21: There seems to be a quite a bit of repeatable code between `run-benchmark.yml`, `run-benchmark-natively.yml` and `test-nightly.tml`. I was hoping that the common parts could be turned into parameterized actions

crates/jstz_tps_bench/tps_test/src/index.ts
zcabter: Where was the docstring added?

crates/jstz_tps_bench/fa2/package.json
alanmarkoTrilitech: ok, whatever, done

alanmarkoTrilitech:
> Could you also update `jstz_tps_bench/README.md`?

done

crates/jstz_tps_bench/src/generate_other.rs
```diff
@@ -0,0 +1,162 @@
...
+    transfers: usize,
+    transfer_endpoint: Option<&str>,
+) -> Result<()> {
+    let endpoint = transfer_endpoint.unwrap_or("transfer");
+    let transfer_endpoint_uri =
+        Uri::try_from(format!("jstz://{contract_address}/{endpoint}"))?;
+
+    for _transfer_id in 0..transfers {
+        builder.run_function(
+            &mut accounts[0],
```
zcabter: Again - If we only use the head account, why not just pass a single account into this function?

crates/jstz_tps_bench/src/generate_other.rs
```diff
@@ -0,0 +1,162 @@
...
+use jstz_proto::context::account::Address;
+use jstz_proto::HttpBody;
+use tezos_smart_rollup::types::SmartRollupAddress;
+use tezos_smart_rollup::utils::inbox::file::InboxFile;
+
+use jstz_utils::inbox_builder::{Account, InboxBuilder, Result};
+
+/// Generate the requested transactions from the given contract, writing to `./inbox.json`.
+///
+/// This includes contract deployment, initialisation, requested transactions, and checks at the end.
```
zcabter:
```suggestion
/// This includes contract deployment, initialisation, requested operations, and checks at the end.
```

crates/jstz_tps_bench/src/main.rs
```diff
@@ -47,13 +44,62 @@ enum Commands {
...
+    #[command(about = "Generate FA2 transactions")]
+    Fa2 {
+        #[arg(long, default_value = DEFAULT_ROLLUP_ADDRESS)]
+        address: String,
+        #[arg(long)]
+        transfers: usize,
+        #[arg(long, default_value = "inbox.json")]
+        inbox_file: Box<Path>,
+    },
+    #[command(about = "Generate other types of transactions")]
```
zcabter:
```suggestion
    #[command(about = "Generate other types of operations")]
```

PR #1322, opened 2025-09-15
feat(sequencer): run riscv pvm in sequencer

huancheng-trili:
# Context

Completes JSTZ-804.
[JSTZ-804](https://linear.app/tezos/issue/JSTZ-804/configure-riscv-pvm-in-sequencer)

# Description

Updated jstz node such that it runs the RISCV sequencer when configured. This includes a few things:
* Reconstruct runtime DB from sequencer debug log
* Fail jstz node when either rollup storage sync or sequencer storage sync errors out
* Spawn a worker that launches an embedded RISCV PVM that loads the given RISCV kernel

Note:
* The logger filter is updated since the previous one prints log lines from external libraries as well and the embedded RISCV PVM has a lot of them.
* The `poll_receipt` deadline in the integration test suite is increased considering that the RISCV setup might be slower.

# Manually testing the PR

* Integration test: one test is added. The test sits behind a feature flag because it cannot fit into CI due to resource consumption. To run the test,
  * Build the RISCV kernel and move that to `crates/jstz_node/tests/riscv_kernel`
  * ```
    cd $(git rev-parse --show-toplevel)/crates/jstz_node
    cargo test --test sequencer run_riscv_sequencer --features riscv_test --release
    ```
  The test needs to run in release mode for the sake of performance. The test will be added to the nightly tests later on.


ryutamago (dismissed):

crates/jstz_node/src/lib.rs
```diff
@@ -205,18 +227,25 @@ pub async fn run(
...
-    match storage_sync_handle {
-        Some(storage_sync_handle) => {
+    match storage_sync_handles.is_empty() {
+        false => {
             let (tx, rx) = tokio::sync::oneshot::channel();
             axum::serve(listener, router)
-                .with_graceful_shutdown(async {
-                    let _ = tx.send(storage_sync_handle.await);
+                .with_graceful_shutdown(async move {
+                    let _ = tx.send(tokio::select! {
```
ryutamago, line 235: Any reason to use `select!` over `join_next().await` here?

crates/jstz_node/src/sequencer/worker.rs
```diff
@@ -196,6 +238,96 @@ fn write_heartbeat(heartbeat: &Arc<AtomicU64>) {
...
+                                    // outdated heartbeats.
+                                    error!("worker failed to decode original inbox message: {e}");
+                                    break 'worker;
+                                },
+                                Ok(m) => (message.inbox_id, Ok(m))
+                            }
+                            },
+                            WrappedOperation::FromNode(signed_op) => {
+                                (
+                                    InboxId{l1_level: 0, l1_message_id: 0},
```
ryutamago, line 299: what does pvm do with this empty inbox id for operation from node?

ryutamago (approved):

PR #1332, opened 2025-09-24
feat(kernels): lightweight kernel

alanmarkoTrilitech:
# Context
[JSTZ-921](https://linear.app/tezos/issue/JSTZ-921/implement-lightweight-kernel)

# Description
This PR adds a lightweight kernel the just outputs the inbox messages to kernel debug log. It additionally moves all kernel crates that we currently have into a common folder.

# Manual testing
```cargo test -p jstz_lightweight_kernel```

### CODE REVIEW

File: crates/kernels/jstz_lightweight_kernel/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,73 @@
+use jstz_crypto::hash::Hash;
+use jstz_crypto::smart_function_hash::SmartFunctionHash;
+use jstz_kernel::inbox::{self, ParsedInboxMessage};
+use tezos_smart_rollup::{
+    entrypoint,
+    prelude::{debug_msg, Runtime},
+};
+
+const DEFAULT_TICKETER_ADDRESS: &str = "KT1F3MuqvT9Yz57TgCS3EkDcKNZe9HpiavUJ";
```
ryutamago, line 9: nit:
jstz node uses `pub const TICKETER: &str = "KT1F3MuqvT9Yz57TgCS3EkDcKNZe9HpiavUJ";` we could export /reuse this value.

alanmarkoTrilitech: this constant is defined at various other places too, so its a more involved task to unify them which is I think outside the scope of this PR. For now I would keep it here, and then we can later unify it if needed.

ryutamago (approved):

PR #1334, opened 2025-09-25
feat: generate and use runtime snapshot in kernel

zcabter:
# Context

Loading, parsing, and compiling JavaScript at runtime is expensive. V8’s snapshot mechanism allows the engine to start from a pre-initialized state, eliminating this overhead. Using snapshots yielded a 5× increase in transactions per second.

Closes: https://linear.app/tezos/issue/JSTZ-790/snapshot-jsruntime

# Description
1. Add generate_snapshot helper to JstzRuntime
2. Set global `OnceLock` `SNAPSHOT` variable
3. Update `JstzRuntimeOptions` to take a snapshot as input and initialize use snapshot if available in `load_and_run`

# Manually testing the PR
`make`

__Old TPS__
10 FA2 transfers took 8.557973791s @ 1.169 TPS

__New TPS__
10 FA2 transfers took 1.953596735s @ 5.119 TPS

### DIFF
--- a/crates/jstz_runtime/src/runtime.rs
+++ b/crates/jstz_runtime/src/runtime.rs
@@ -1,6 +1,10 @@
 use crate::error::Result;
 use crate::ext::jstz_fetch::FetchAPI;
 use crate::ext::jstz_fetch::NotSupportedFetch;
+use deno_core::error::CoreError;
+use deno_core::snapshot::create_snapshot;
+use deno_core::snapshot::CreateSnapshotOptions;
+use deno_core::snapshot::CreateSnapshotOutput;
 use deno_core::v8::new_single_threaded_default_platform;
 use deno_core::*;
 use derive_more::{Deref, DerefMut};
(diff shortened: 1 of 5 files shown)


crates/jstz_runtime/src/runtime.rs
```diff
@@ -106,28 +114,41 @@ impl JstzRuntime {
...
-        extensions.extend(init_base_extensions_ops_and_esm::<F>());
-        extensions.extend(options.extensions);
-        Self::new_inner(extensions, options.module_loader, options.protocol, None)
-    }
+        if options.snapshot.is_none() {
+            extensions.extend(init_base_extensions_ops_and_esm::<F>());
+        } else {
+            // Initializing from a snapshot only requires initializing
+            // the rust ops
+            extensions.extend(init_base_extensions_ops::<F>());
```
alanmarkoTrilitech, line 122: why can't those be also initialised in the snapshot?, probably good to also add a comment about it there

alanmarkoTrilitech (commented):
LGTM, though I think it would be better if the snapshot initialisation was separate.

I prefer that we merge this after merging the benchmark, so that we can get some initial numbers and then write an update showing how this change made the overhead 5x smaller

alanmarkoTrilitech (approved):
:shipit:

PR #1343, opened 2025-10-01
fix(proto): add test_utils flag

alanmarkoTrilitech:
# Context
Due to changes in #1332, a test fails due to unresolved import since the test_util feature does not propagate when only a single package is tested.

# Description
Adds test_utils feature as a dependency to jstz_kernel and jstz_proto. 
Additionally it removes gating on v2_runtime for some DebugLogSink, which isn't needed anymore and is anyway gated on test_utils feature. 

# Manual testing
```make```
```cargo nextest run -p jstz_proto --features v2_runtime```

### DIFF
--- a/crates/jstz_proto/Cargo.toml
+++ b/crates/jstz_proto/Cargo.toml
@@ -47,7 +47,7 @@ url.workspace = true
 [dev-dependencies]
 bincode.workspace = true
 jstz_mock = { path = "../jstz_mock" }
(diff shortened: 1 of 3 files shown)


ryutamago (commented):
the jstz rutnime was missing the features so i've added [here](https://github.com/jstz-dev/jstz/pull/1344) to unblock myself. Thanks for the quick fix 🙏

ryutamago (approved):

PR #1345, opened 2025-10-03
fix(cli): handle dev network in network commands

huancheng-trili:
# Context

Part of JSTZ-783.
[JSTZ-783](https://linear.app/tezos/issue/JSTZ-783/support-jstz-network-addupdatedelete)

# Description

Updated network commands so that they properly handle the reserved `dev` network.

* list: print one row for `dev` and ignore any "custom" dev network in the config file
* add, update, delete: fail when the network name is `dev`

# Manually testing the PR

* Integration test: updated relevant tests


ryutamago (approved):
Better UX! 💪

